PR #113, opened 2012-09-09
Added SimpleDBHandler

claylo:
I've added a SimpleDBHandler to log messages using a passed-in AmazonSDB object from the official Amazon SDK for PHP.

Given the size limitations (which are gracious, but still exist) of SimpleDB domains, I've taken the approach of mapping log channels-to-SimpleDB domains.

While SimpleDB has support for batch put operations, and the AmazonSDB class has support for sending items in parallel -- which, when combined, could conceivably allow for parallel batch puts -- I have kept this first commit simple, so that this handler's behavior mirrors that of most of the others.

If it winds up being popular and users request it, I'd be happy to contribute a further SimpleDBBatchHandler that dives into supporting those scenarios.

Meanwhile, I look forward to your feedback on this less complicated submission.


Seldaek:
Closing as it's very old, it can always be done as an third party handler package if anyone really wants.

PR #147, opened 2013-01-15
Added zeromq handler

mkraemer:
Simple handler which allows logging into a zeromq socket :)


Seldaek:
Closing as it's very old, it can always be done as an third party handler package if anyone really wants.

PR #384, opened 2014-06-30
Support for log file locking

### LINKED ISSUE #379: Parallel logging interference (I assume)
Hey! I've been noticing something lately:

```
[2014-06-21 01:57:18] app.DEBUG: [/next_item] requested item (19803) with duration (8.0s) will start playing in (50.0s), so we expect communication by 2014-06-21 01:58:21 [] []
[2014-06-21 01:57:22] app.DEBUG: [/next_item] requested item (18504) with duration (10.0s) will start playing in (12.0s), so we expect[2014-06-21 02:15:56] app.DEBUG: [/next_item] requested item (17933) with duration (24.0s) will start playing in (30.0s), so we expect communication by 2014-06-21 02:16:55 [] []
[2014-06-21 02:15:56] app.DEBUG: [/next_item] requested item (18507) with duration (10.0s) will start playing in (10.0s), so we expect communication by 2014-06-21 02:16:21 [] []
```

As you can see, the logs lines interfere, which of course messes with my automated log parsing as well.

This is all from exactly 1 webapp that's firing requests at the server, some of those requests may indeed be spaced very closely in time. I reckon the problem might be worse when you have a lot of concurrent web users. I have 2 PHP-FPM processes running.

What information could I provide to help better alleviate the issue?

bobrik:
Closes #379.

Not sure how to test failing `flock`. Any ideas?

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -26,14 +26,16 @@ class StreamHandler extends AbstractProcessingHandler
     protected $url;
     private $errorMessage;
     protected $filePermission;
+    protected $preferLocking;
 
     /**
      * @param string  $stream
      * @param integer $level           The minimum logging level at which this handler will be triggered
      * @param Boolean $bubble          Whether the messages that are handled can bubble up the stack or not
      * @param int     $filePermissions Optional file permissions (default (0644) are only for owner read/write)
+     * @param Boolean $preferLocking   Try to lock log file before doing any writes
      */
-    public function __construct($stream, $level = Logger::DEBUG, $bubble = true, $filePermission = null)
+    public function __construct($stream, $level = Logger::DEBUG, $bubble = true, $filePermission = null, $preferLocking = false)
     {
         parent::__construct($level, $bubble);
         if (is_resource($stream)) {
@@ -43,6 +45,7 @@ class StreamHandler extends AbstractProcessingHandler
         }
 
         $this->filePermission = $filePermission;
+        $this->preferLocking = $preferLocking;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mta59066:
Locking for who? Flock is an advisory lock and does not enforce anything. It is only useful if you look for it. So what this does is place an advisory lock on the file against other possible writers who might check for this lock. 

It does not do anything about #379 and the existence of it gives the false assumption that there is a locking mechanism between different processes utilizing monolog.

stof:
hmm indeed. Ignoring the locking error makes the locking useless to solve the issue. But please open an issue instead of commenting on a 4.5 year old merged PR. Most people won't ever look at it.

IMSoP:
@mta59066 (and anyone else that comes upon this comment) The lock is "advisory" in terms of code that never requests a lock, but it is still _blocking_ for all code that _does_ request a lock. If you have two instances of the same code, and both call `flock`, one process will be forced to wait until the other process releases the lock. That should be enough to stop two log messages interrupting each other in the same file.

You can demonstrate this for yourself with a simple test script like this:

```
$fh = fopen('/tmp/example.log', 'a');
echo getmypid(), " acquiring lock...\n";
flock($fh, LOCK_EX);
echo getmypid(), " lock acquired; sleeping for 10 seconds...\n";
sleep(10);
echo getmypid(), " releasing lock...\n";
flock($fh, LOCK_UN);
echo getmypid(), " done.\n";
```

Running that 3 times gives the following output:

```
3458 acquiring lock...
3458 lock acquired; sleeping for 10 seconds...
3468 acquiring lock...
3475 acquiring lock...
3458 releasing lock...
3458 done.
3475 lock acquired; sleeping for 10 seconds...
3475 releasing lock...
3475 done.
3468 lock acquired; sleeping for 10 seconds...
3468 releasing lock...
3468 done.
```

PR #391, opened 2014-07-17
new PDO handler, to be used with mysql

cramiro:
Add PDO Handler


fuzzy76:
Why was this closed if the branch was deleted?

cramiro:
Was not intended to commit here...was a personal use case

PR #402, opened 2014-07-31
Added microsecond support for JSON serialized DateTime objects

holtkamp:
When using the JsonFormatter, all records are encoded using json_encode(). For the timestamp this would mean that a DateTime object results in something like:

```
    "datetime": {
        "date": "2014-07-31 11:15:22",
        "timezone_type": 3,
        "timezone": "Europe/Amsterdam"
    }
```

This way we 'loose' the ability to format the timestamp with microsecond precision. A way to avoid this loss of detail is to extend the native DateTime class with one that implements [JsonSerializable](http://php.net/manual/en/class.jsonserializable.php) which appends the microseconds.

Note that due to https://bugs.php.net/bug.php?id=60302, this means we can no longer use DateTime::createFromFormat(), so I isolated it in a private function.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
         }
     ],
     "require": {
-        "php": ">=5.3.0",
+        "php": ">=5.4.0",
         "psr/log": "~1.0"
     },
     "require-dev": {
--- a/src/Monolog/DateTime.php
+++ b/src/Monolog/DateTime.php
@@ -1,0 +1,53 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog;
+
+/**
+ * Add microsecond detail to JSON encoded DateTime objects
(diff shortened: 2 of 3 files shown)


holtkamp:
Might/should this PR be reconsidered? I noticed the required PHP version has been increased to 7 :smile:

ZergMark:
@holtkamp Take a look at the updated [Logger.php](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Logger.php), it appears to handle this with lines 316-320. I can't tell from your code that you are doing more that that.

holtkamp:
@ZergMark , you are totally right, great! Closing issue :+1:

Seldaek:
We do handle microsecond datetimes, except we don't format them with microseconds in json outputs. Not sure if we should by default or not. I'm leaning towards disabling the microsecond precision by default as it's a bit faster to create datetimes then. The formatting of json dates should just be configurable as well then ideally so if you enable microsecond you get it everywhere. I'll open a new issue.

holtkamp:
@Seldaek , aah, I checked the JsonFormatter, but missed the part that the microsecond formatting gets dropped. Thanks for being sharper than me this morning :smile: Good to discuss this in a separate issue.

PR #482, opened 2014-12-23
Fix warnings in case of incorrect data

vkhramtsov:
If somebody pass incorrect input array to `format`, php will raise several notices, because no any checks performed before access to array indexes.
Please review.


vkhramtsov:
@Seldaek,  what do you think about this pr?

Seldaek:
I fixed it in another way so it mostly works except if essential data is missing then you get an exception.

PR #490, opened 2015-01-05
Remove the unused private methods

SofHad:


stloyd:
This is wrong. Those methods are called basing on [schema choose](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/CubeHandler.php#L119).

Seldaek:
Yup, IDEs don't know it all.

PR #495, opened 2015-01-14
Fix spurious E_WARNING being emitted from date_default_timezone_get().

jchesterpivotal:
Other packages relying on Monolog will sometimes interpret E_WARNINGs as exceptions, causing
them to fail. The usual advice is to edit php.ini to set the date.timezone setting. But this
approach requires turning every development workstation into a snowflake. Instead, we follow
the lead of other projects and introduce this snippet to modify the default PHP behaviour.

Normally, we would provide a test case to demonstrate that the E_WARNING is no longer
emitted from Monolog. However we could not find a way to test this behaviour. To start with,
multiple sites in the test code (both PHPUnit and the Monolog bootstrap.php) will either
use ini_set() on date.timezone, or date_default_timezone_set(), to set the timezone to UTC.
Once this is done, the E_WARNING is never emitted again. Then, if we try to use ini_set()
or ini_restore() in the test, PHP either rejects (ini_set) or ignores (ini_restore) our
attempt to recreate the default conditions which lead to this problem. There may be a way
to test this, but we are not aware of one.

The snippet included here, or similar snippets, appears widely in the PHP universe. An
identical snippet appears in Composer[1] and semantically identical code has been baked
into phpBB[2] and Mediawiki[3], amongst others that we found while searching.

In an ideal world, the PHP interpreter would stop emitting the E_WARNING. UTC is a sane
default value for timezone settings, and users who required a different setting would
be motivated to find the mechanism for changing it, leaving the common case unpunished.
However, it seems that this is unlikely to occur[4]. Hence, this patch is required to
work around the shortcomings of PHP.

[1] https://github.com/composer/composer/commit/e5ecad7eeb035e7579e5cc729b46412188ebb2ae
[2] https://tracker.phpbb.com/browse/PHPBB3-7729
[3] http://git.wikimedia.org/commit/mediawiki%2Fcore.git/8e41f42b67cec4a6256080aea8a89c9e5b0ceb05
[4] http://www.serverphorums.com/read.php?7,712151

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -138,6 +138,10 @@ class Logger implements LoggerInterface
         $this->name = $name;
         $this->handlers = $handlers;
         $this->processors = $processors;
+
+        if (function_exists('date_default_timezone_set') && function_exists('date_default_timezone_get')) {
+            date_default_timezone_set(@date_default_timezone_get());
+        }
     }
 
     /**


Seldaek:
I am not sure if it's really the responsibility of Monolog to do this.

Composer, phpBB and mediawiki are all applications, which are executed as is by users and therefore take the blame whenever something goes wrong, that's why it makes sense to fix it there IMO. But Monolog is a library, and if every single library using date functions had to do this we'd end up with a few dozen *set* calls for the default timezone in every project, which is a bit absurd. I would say whoever bundles it and does not want hell from their users should make sure a sane default timezone is set.

jchesterpivotal:
We (the Cloud Foundry buildpacks team) observed this problem in a post-process step of Symfony's composer installation. The timezone setup that Symfony performs in its code does not propagate. Composer settings don't propagate either, because it's running in a separate PHP instance.

I understand your argument, but from my point of view, Monolog is a point at which this observed issue can be nixed. That it gets solved in multiple places is a consequence of the PHP design decision -- there's no way general way around it if you assume that your code runs against the default php.ini.

Yes, it's absurd that this winds up being done multiple times between different libraries, frameworks and applications. But that's because the default behaviour is broken and each library, framework and application author cannot safely assume that someone else has patched it. I'm not saying this is *good*, so much as a frustrating fact of life.

We've modified our buildpack's php.ini to set the timezone and to ensure that it is used throughout the staging process. This PR was one half of a two-pronged approach to resolving this issue. 

Normally we'd look at opening a bug report or PR against PHP itself, but the linked thread makes it clear that such an effort would be unwelcome.

Thanks for your consideration.

Seldaek:
I am not extremely familiar with cloud foundry so can you explain what you mean by staging process? Is that fixing your issue or not? IMO php.ini should be fixed if you're managing the platform, or date_default_timezone_set should be used if you are managing an application. Libraries should really not have a play in that. Whichever piece of code is using monolog should fix it's timezone.

stof:
Btw, Symfony does not fix the default timezone at the framework-level either. It complains about your php.ini being invalid when you check whether your platform fulfills the requirements

jchesterpivotal:
@Seldaek 

Sorry for the slow reply -- travel sucks.

[Cloud Foundry](http://cloudfoundry.org/index.html) is an opensource PaaS that can be installed in a public or private cloud. As part of the system we support the buildpacks mechanism pioneered by Heroku. By "staging", we refer to the process whereby sourcecode is accepted into a Cloud Foundry installation, identified as being interpreted by a buildpack, and then prepared for execution.

> IMO php.ini should be fixed if you're managing the platform

The PHP buildpack ships with corrected `php.ini` files which are present when an application launches, and we're changing staging to behave the same way. 

What's not fixed is our development environment, which is where this behaviour was observed. We develop on OS X and `brew` installs stock PHP with the stock `php.ini` file. I'm averse to turning our workstations into snowflakes, but your remarks make me think that we could try opening a modification against the brew recipe for PHP.

Thanks again!

Seldaek:
No problem, I hope you get it resolved somehow, or that users eventually learn to configure their PHP :/

PR #496, opened 2015-01-18
Improve a view of extra fields in SlackHandler.

imunew:
SlackHandler supported extra fields in v1.12.0.
But the all of extra fields is set into an 'Extra' field. So, it's hard to see each extra field.


Seldaek:
Do you have a before/after screenshot of how this renders in slack maybe?

imunew:
before (v1.12.0)
monolog_slack_v1 12 0

after(this PR)
monolog_slack_pr496

Seldaek:
#507 fixed it in a better way I believe so closing this, you just need to set $includeExtra to true in the constructor.

PR #497, opened 2015-01-19
Added support for swift message customization in SwiftMailerHandler

iltar:
Currently it's not possible to dynamically alter the `Swift_Message` instance. This because the message is created and injected in the constructor. My goal is to create a dynamic subject and optionally add custom headers based on the `$context` provided in the `$records`. This will allow proper filtering to inboxes and make it easier to determine what the log is about if you decide to customize it. 

I couldn't change the `Swift_Message` instance because I had no access to it. The only existing way of doing so would be to overwrite the `send()` method. This however, seems very ugly when you work in Symfony for example. The problem here is that Symfony extends the Swift handler and [adds some custom logic to the `send()` method](https://github.com/symfony/MonologBridge/blob/master/Handler/SwiftMailerHandler.php#L64-L66). If I want to omit the parent `send()`, I will have to overwrite this `send()` which will make it fragile with updates/bug fixes.

This PR will allow people to extend the whole tree and only overwrite the `getMessage()` method. This will allow you to "pre-process" the data and add headers, a different subject, a CC on critical messages etc. 

List of things added:
 - `getMessage()` in `SwiftMailerHandler`.
 - `suggest` on `swiftmailer/swiftmailer`.
 - `require-dev` on `swiftmailer/swiftmailer` for unit-tests.
 - A unit-test for the `SwiftMailerHandler`.

### DIFF
--- a/tests/Monolog/Handler/SwiftMailerHandlerTest.php
+++ b/tests/Monolog/Handler/SwiftMailerHandlerTest.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Logger;
+use Monolog\TestCase;
+
+/**
+ * @author Iltar van der Berg <ivanderberg@hostnet.nl>
+ * @covers Monolog\Handler\SwiftMailerHandler
+ */
+class SwiftMailerHandlerTest extends TestCase
+{
+    private $mailer;
+
+    public function setUp()
+    {
+        $this->mailer = $this
+            ->getMockBuilder('Swift_Mailer')
+            ->disableOriginalConstructor()
+            ->getMock();
+    }
+
+    /**
+     * @dataProvider sendProvider
+     */
+    public function testSend($message)
+    {
+        $handler = new SwiftMailerHandler($this->mailer, $message);
+        $this->send($handler, is_callable($message) ? $message() : $message);
+    }
+
+    /**
+     * @expectedException \InvalidArgumentException
+     */
+    public function testConstructorSafeGuard()
+    {
+        new SwiftMailerHandler($this->mailer, null);
+    }
+
+
+    public function sendProvider()
+    {
+        $message = new \Swift_Message();
+        return array(
+            array(new \Swift_Message()),
+            array(function () use ($message) { return $message; }), // to make sure we have the same instance every call
+        );
+    }
+
+    private function send(SwiftMailerHandler $handler, \Swift_Message $message)
+    {
+        $newBody = 'henk';
+
+        $this->mailer
+            ->expects($this->once())
+            ->method('send')
+            ->with($this->callback(function ($value) use ($message, $newBody) {
+                return $value instanceof \Swift_Message && $value !== $message && $value->getBody() === $newBody;
+            }));
+
+        $oldBody = $message->getBody();
+
+        $c = new \ReflectionClass($handler);
+        $m = $c->getMethod('send');
+        $m->setAccessible(true);
+        $m->invoke($handler, $newBody, array());
+        $m->setAccessible(false);
+
+        $this->assertEquals($oldBody, $message->getBody());
+    }
+}
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
         "ruflin/elastica": "0.90.*",
         "doctrine/couchdb": "~1.0@dev",
         "aws/aws-sdk-php": "~2.4, >2.4.8",
-        "videlalvaro/php-amqplib": "~2.4"
+        "videlalvaro/php-amqplib": "~2.4",
+        "swiftmailer/swiftmailer": "5.*"
     },
     "suggest": {
         "graylog2/gelf-php": "Allow sending log messages to a GrayLog2 server",
@@ -34,7 +35,8 @@
         "ext-amqp": "Allow sending log messages to an AMQP server (1.0+ required)",
         "ext-mongo": "Allow sending log messages to a MongoDB server",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
-        "rollbar/rollbar": "Allow sending log messages to Rollbar"
+        "rollbar/rollbar": "Allow sending log messages to Rollbar",
+        "swiftmailer/swiftmailer": "Allow sending (buffered) log messages via Swiftmailer"
     },
     "autoload": {
         "psr-4": {"Monolog\\": "src/Monolog"}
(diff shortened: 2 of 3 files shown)


tests/Monolog/Handler/SwiftMailerHandlerTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $message = new \Swift_Message();
+        return array(
+            array(new \Swift_Message()),
+            array(function () use ($message) { return $message; }), // to make sure we have the same instance every call
+        );
+    }
+
+    private function send(SwiftMailerHandler $handler, \Swift_Message $message)
+    {
+        $new_body = 'henk';
```
stof: camelCase name please

iltar: @stof My bad, fixed both `$old_body` and `$new_body`.

johnkary:
Hi @iltar, thanks for starting to look at this. I was actually looking at `SwiftMailerHandler` today and wanted to make some similar changes to how Swift_Message is used, but for a different reason.

We currently use `SwiftMailerHandler` in our production Symfony 2.3 app via MonologBundle to send email to our dev team when critical errors occur. I started looking at the handler because I noticed via Blackfire.io profiling that a Swift_Message object is instantiated in the constructor and saved in the handler object.

My profiling showed creating the Swift_Message object was taking ~25ms when returning a blank Response from a Controller. (The slowness is partly due to a Vagrant VM + NFS mount + no APC in my profiling environment.) The bigger issue is, as far as I can tell, a Swift_Message object is instantiated even when an email is never sent by the handler! For responses that do not send email (99% of ours) I would rather not load Swift_Mailer classes I don't use.

I'd like to see the creation of Swift_Message become a lazy operation--so the message is only ever created if an email needs to be sent. My initial solution is seen here:

https://gist.github.com/johnkary/1a10a465721d540f3ab7

I took your use case and expanded my example to now pass the $content and $records to the callback responsible for creating Swift_Message. My test case also confirms this is working. I chose to test strictly through the class' public API instead of exposing its private methods:

https://github.com/johnkary/monolog/commits/LazySwiftMailerHandler

I'm pretty sure my implementation breaks BC for anyone that is extending SwiftMailerHandler, because after the Handler is instantiated `$this->message` is no longer fully initialized to a Swift_Message instance. But I think this is a big enough performance gain to warrant the BC break.

Let me know if I should add any more to my branch (composer require swiftmailer?) and submit mine as a PR.

iltar:
Hey @johnkary, glad to see I'm not the only one who doesn't like the design if this class. In your example provided, you still have a message injected. I'm pretty sure that this won't give a big benefit. The only thing I can imagine, is that the `call_user_func` is calling a method you have implemented which takes up a lot of time. Currently, it seems like the `Swift_Message` is used as a template, as you can note by the clone. Your solution will modify that template as you forgot to clone it. 

I'm actually having another idea which can make it BC, a config data object can be inserted which contains nothing more than simple getters/setters. This way, it can be truly lazy loaded without having to create and clone a `Swift_Message`.

johnkary:
@iltar Yes, the test case I supplied for your use case creates a Swift_Message, but passing a callback still supports lazy creation. I have added a test case to demonstrate this: 50f35a30  If the creation were not lazy the Exception would always be thrown.

Thanks for spotting the missing `clone` when a Swift_Message object is injected. I fixed the issue here: a95d87984  I don't think we need to clone the object returned by the callback because it will always be a freshly instantiated instance.

iltar:
@johnkary To be honest, I'm not really fond of that implementation. But as said, if this PR will get merged, I will add a nicer solution to the problem using a data object for the currently existing `SwiftMailerHandler`. I don't think it's a good idea to fragment this.

johnkary:
@iltar Thanks for your feedback and continued discussion. I still think my implementation works best and solves both use cases with less code and less work by the developer. Looking at my implementation:

1. Customizing how SwiftMailerHandler works can be done without inheritance. Composition works perfectly for this case. In fact this is exactly how MonologBundle customizes Swift_Message creation right now. See below.
2. SwiftMailerHandler becomes a stateless service.
3. Not only do we gain lazy instantiation but we gain customization of Swift_Message based on logged context and records.

As you commented on the commit, my implementation does have a drawback of invoking the callable on each call to ->send(). Let's keep in mind ->send() is **only** invoked when the log threshold is exceeded. Hopefully our applications don't exceed that threshold often else we're sending a lot of email :)

For my case I'm only sending email when a level ERROR message is logged, which is < 1% of all application responses. If my app had a 1% error rate I'd probably need to fix it so it didn't do that. I would rather take a very slight perf penalty invoking the callable each send than taking an autoload + object creation penalty **on 100% of my requests.** The first callable invocation will take the biggest perf penalty (hint: autoloading) while successive calls will only simply instantiate the object.

Symfony's MonologBundle right now can craft an object+method callable to create Swift_Message objects. Unfortunately this functionality isn't yet documented in the README or the [Symfony cookbook](http://symfony.com/doc/current/cookbook/logging/monolog_email.html) describing sending email. But check this out:

```yaml
monolog:
    handlers:
        # ... snipped some other config here ...
        swift:
            type:       swift_mailer
            level:      error
            # Below service builds the Swift_Message that will be sent
            email_prototype:
                id: my_error_email_prototype
                method: buildMessage
```

A callable is constructed using service id `my_error_email_prototype` and its `buildMessage()` method. Remember a callable be an anonymous function, closure or a class method like this:

```php
$prototype = new MonologErrorEmailPrototype();
$callable = array($prototype, 'buildMessage');
```

I'm sure you know this, but check out this example code for how each type of callable can be used with SwiftMailerHander: http://3v4l.org/Ftl6q

To support your use case of customizing Swift_Message creation SwiftMailerHandler really just needs this change:

```php
public function send($content, array $records) {
    // ...
    $message = call_user_func($this->message);
    // ...
}
```
to this:
```php
public function send($content, array $records) {
    // ...
    $message = call_user_func($this->message, $content, $records);
    // ...
}
```

I think the callback solution is more extensible than the negatives of requiring the coupling of inheritance, continuing to allow SwiftMailerHandler to be a stateful service and continuing to require Swift_Message instantiation on 100% of all calls.

iltar:
ping @Seldaek Is this okay to be merged or are there still issues?

Seldaek:
So.. looking at this, I am not sure what to say :)

I tend to agree with @johnkary because mails are typically not sent more than once within a request, but I don't see a PR.. I'd like to release a v1.13 ASAP and this is the only thing left I'd like to have in.

iltar:
@Seldaek If desired, I can add lazy loading of the message during `getMessage()`. I didn't want to add that functionality to this PR as it was not fully related.

johnkary:
@Seldaek I'll look at this again tonight and have PR(s) for review tomorrow.

iltar:
@johnkary is it possible you base that change on mine? It's exactly what I had in mind. Except that with your change, I still can't customize the handling properly due to the layer Symfony puts between it with the memory spool.

iltar:
Closed in facor of https://github.com/Seldaek/monolog/pull/519

PR #501, opened 2015-01-29
Nette Framework integration

fprochazka:
Hi, I've made an integration of Monolog into [Nette Framework](http://nette.org/en/) which has [reasonably big userbase](https://packagist.org/packages/nette/) and is the most popular PHP framework in Czech Republic.


stof:
@Seldaek what about moving this list to the wiki too rather than the readme ?

Seldaek:
Eh well that's not changing so often so it's ok there I think

fprochazka:
Thank you!

Issue #493, opened 2015-01-07
Hipchat has more detailed logs then Slack?


peric:
Is it possible to have more detailed logs on Slack?

On Hipchat, I get where request was sent and then completely detailed critical error/exception. 
On Slack, I get just where exception has happened, without any details about request.

Seldaek:
I guess this is due to the fact Slack doesn't include context info and includes extra only optionally. You could try to enable the $includeExtra param in the constructor, and then possibly we should fix the handler to include context also when includeExtra is on.

peric:
Thanks for the answer. Hopefully I will have time to check it and maybe even contribute.

Seldaek:
I think this has been resolved by #507

PR #507, opened 2015-02-09
#493 Include context in slack extra info.

jamiel:
Adds the context items to Slack attachments.

```
$handler->addError('Foo', array('foo' => 'bar'));
```
screen shot 2015-02-09 at 22 42 36

### DIFF
--- a/src/Monolog/Handler/SlackHandler.php
+++ b/src/Monolog/Handler/SlackHandler.php
@@ -157,26 +157,47 @@ class SlackHandler extends SocketHandler
             }
 
             if ($this->includeExtra) {
-                $extra = '';
-                foreach ($record['extra'] as $var => $val) {
-                    $extra .= $var.': '.$this->lineFormatter->stringify($val)." | ";
+
+                if (!empty($record['extra'])) {
+
+                    if ($this->useShortAttachment) {
+
+                        $attachment['fields'][] = array(
+                            'title' => "Extra",
+                            'value' => $this->stringify($record['extra']),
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/SlackHandler.php
```diff
@@ -169,6 +169,15 @@ private function buildContent($record)
...
                 );
+
+                // Add all context items as individual fields
+                foreach ($record['context'] as $var => $val) {
+                    $attachment['fields'][] = array(
+                        'title' => $var,
+                        'value' => $val,
+                        'short' => true
+                    );
+                }
```
Seldaek: I am ok with this but I think it should be in line with the Extra stuff, and both should actually follow the useShortAttachment flag. If it's on both context and extra should be like lines 161-171 above, and if it's false both should be like you have it for the context. And if extra|context are empty no attachment should be created IMO, even in useShortAttachment mode.

noetix:
+1

jamiel:
That's been adjusted so that both extra and context are consistent depending on useShortAttachment and are only added if the fields are present and not empty.

To be honest, it's still not perfect as you should be able to choose one or the other and short attachment shouldn't be used to decide whether to stringify the the object or not as Slack's intended use is just whether the fields can be stacked next to each other not in two columns.

Seldaek:
Thanks! That sounds like good progress already so I'll merge, but if you feel like this can be improved upon please send another PR. I'd just like to see before/after screenshots because I don't really have the time or ability to test with slack to see how it comes out.

PR #511, opened 2015-02-20
Add support for formatting DateTimeImmutable

thetrompf:


src/Monolog/Formatter/LogglyFormatter.php
```diff
@@ -37,7 +37,7 @@ public function __construct($batchMode = self::BATCH_MODE_NEWLINES, $appendNewli
      */
     public function format(array $record)
     {
-        if (isset($record["datetime"]) && ($record["datetime"] instanceof \DateTime)) {
+        if (isset($record["datetime"]) && ($record["datetime"] instanceof \DateTimeInterface)) {
```
xabbuh: This won't work with PHP 5.3 and 5.4. The `DateTimeInterface` was introduced in PHP 5.5.

thetrompf: Oh my bad, I hadn't checked up on that

Seldaek:
Indeed, let's keep this around for v2 which I believe will require 5.5 or above

stof:
@Seldaek just add instanceof checks for both DateTime and DateTimeInterface and it will keep working on 5.3 and 5.4 while supporting 5.5+ fully.

Seldaek:
Yeah that's alright for instanceof but my concern was more about the type hints. We can just remove the hints but arguably that'd make it a BC break...

ZergMark:
This should be fine now that 2.x is PHP7+

Seldaek:
Replaced by https://github.com/Seldaek/monolog/pull/793

PR #515, opened 2015-02-27
Added overwriting of user context to RavenHandler

smazurov:
Its not currently possible to set user context in RavenHandler (for getsentry.com), this PR adds the functionality and related test.


Seldaek:
Thanks

PR #520, opened 2015-03-02
Use data.logentries.com and port 443 instead of api

StephenHynes7:
It's preferable to be using data.logentries.com when sending Token Based logs.

Also it's nicer if we use port 443 since it is more then likely going to be open compared to port 20000 in environments.


Seldaek:
Merging but please note that they say on the page HTTP PUT will be deprecated and users should move to https://logentries.com/doc/input-token/ - might be worth it sending a PR for that :)

StephenHynes7:
@Seldaek Thanks. This integration is using Token TCP. While we say we plan to deprecate HTTP PUT it will more then likely be some time before this is put into place.

Seldaek:
@StephenHynes7 ah you do actually work for logentries, thanks then nothing to see here I assume you know what the plan is more than I do ;)

StephenHynes7:
@Seldaek Yup =P

PR #521, opened 2015-03-03
Allow passing additional parameters to PHP's mail().

elyobo:
PHP's mail() function accepts an optional additional_parameters argument which can be used to provide additional information, such as setting the envelope sender, for some MTAs.  This patch allows the NativeMailerHandler to pass these through.

### DIFF
--- a/src/Monolog/Handler/NativeMailerHandler.php
+++ b/src/Monolog/Handler/NativeMailerHandler.php
@@ -40,6 +40,12 @@ class NativeMailerHandler extends MailHandler
     protected $headers = array();
 
     /**
+     * Optional parameters for the message
+     * @var array
+     */
+    protected $parameters = array();
+
+    /**
      * The wordwrap length for the message
      * @var integer
      */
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/NativeMailerHandler.php
```diff
@@ -40,6 +40,12 @@ class NativeMailerHandler extends MailHandler
     protected $headers = array();
 
     /**
+     * Optional parameters for the message
+     * @var array
+     */
+    protected $parameters = array();
```
xabbuh: Imho this should be private.

elyobo: Can do, I just followed the implementation of $headers for consistency.

elyobo: @Seldaek Any preference?

Seldaek: It's ok IMO since header is already protected, might as well keep it all open for extension.

elyobo:
Great, thanks.

PR #522, opened 2015-03-03
Return self from NativeMailerHandler::addHeader().

elyobo:
NativeMailerHandler::setContentType() and NativeMailerHandler::setEncoding() both return the NativeMailerHandler object, allowing a fluent coding style.  This fix adds the same behaviour to NativeMailerHandler::addHeader() for so that all public non-inherited methods behave this way for consistency.


Seldaek:
Thanks

elyobo:
No probs, thanks for merging.

PR #526, opened 2015-03-10
Add Hipchat V2 support

esalter:
Adds support for Hipchat's V2 API by way of setting the version flag.  V1 paths
should remain unchanged, and v1 remains the default for BC.

I have no way of testing V1 at this time, can someone pull this down to verify that still works?

### DIFF
--- a/src/Monolog/Handler/HipChatHandler.php
+++ b/src/Monolog/Handler/HipChatHandler.php
@@ -21,12 +21,23 @@ use Monolog\Logger;
  * Room      - HipChat Room Id or name, where messages are sent
  * Name      - Name used to send the message (from)
  * notify    - Should the message trigger a notification in the clients
+ * version   - The API version to use (HipChatHandler::API_V1 | HipChatHandler::API_V2)
  *
  * @author Rafael Dohms <rafael@doh.ms>
  * @see    https://www.hipchat.com/docs/api
  */
 class HipChatHandler extends SocketHandler
 {
+    /**
+     * Use API version 1
+     */
+    const API_V1 = 'v1';
+
+    /**
+     * Use API version v2
+     */
+    const API_V2 = 'v2';
+
     /**
      * The maximum allowed length for the name used in the "from" field.
      */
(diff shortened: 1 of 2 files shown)


m4tthumphrey:
Have just tested and all seems good. :+1:

src/Monolog/Handler/HipChatHandler.php
```diff
@@ -93,6 +106,22 @@ public function __construct($token, $room, $name = 'Monolog', $notify = false, $
...
+        $this->setVersion($version);
+    }
+
+    /**
+     * Convenience method for setting the API version. Valid values are
+     * - HipChatHandler::API_V1
+     * - HipChatHandler::API_V2
+     * @param $version string the API version to use
+     */
+    public function setVersion($version) {
```
xabbuh: Is it really needed to be able to change the handler version after creation?

m4tthumphrey: I would assume it's so the user doesn't have to specify all of the default arguments just to use v2?

xabbuh: Wouldn't it then be better to have convenient factory methods for this?

m4tthumphrey: You mean something like

```php
$logger->pushHandler(HipChatHandler::createInstance('v1', ...))
$logger->pushHandler(HipChatHandler::createInstance('v2', ...))
```

I guess thats another way of doing it!

esalter: Yeah the idea was that you could create a new handler without having to specify all the default args.  Having a constructor or a factory that would take arguments might be better but I didn't want to break BC by doing the constructor.  Not sure if a factory is worth it.

m4tthumphrey: I don't think it is either. If the additional `setVersion()` method is an issue I'd just say drop it and then people will have to set all arguments...

esalter: I don't mind that much so I can take it out.

esalter:
I can't get notify to work either...tried using Postman using a variety of values.

esalter:
FWIW we are able to use this on our side without issue.  Also a coworker reports that the notify works for him, so it must be some kind of client setting, though I don't know what it is because everything looks to be turned on for me.

andrioli:
Hey guys, there are something blocking this PR? It will be included in the next release?

esalter:
I'm not aware of any reason this can't be merged.   BTW confirmed that notification works, it just doesn't notify yourself. :)

m4tthumphrey:
@esalter Ah yes, that would sense as v2 uses a real user account as the author!

andrioli:
V2 also allows use of "room tokens". probably with this tokens the
notifications work

andrioli:
Using the "room token" do not work. After take a look in the doc I realize that the **notify** property should be a String|Boolean.

I did these changes `'notify' => $this->notify ? "true" : "false",` and the notifications became to work :)

Seldaek:
Sorry to ask but I am not so familiar with hipchat, what's the use case for still having v1 support? Also is the PR ready and are you both V1 and V2 users happy with it? I'll take your word for it that it's been tested and go ahead and merge if so.

andrioli:
V1 is deprecated. But I don't know when it will be disabled. This PR only
added support to V2 API (V1 still works as the default behavior).

IMO this can be merged after the fix I suggest in the previous comment.

WDYT @esalter?

Seldaek:
Yeah what I mean is more.. why not *only* support v2? Is there anything v2 is missing? Or are some people stuck on v1 for some reason? Are the API tokens maybe not the same? Curious why we need to support **both**.

esalter:
I think we should support both for the time being - I don't think there's any reason to lock out old clients, all this does is add the option to use V2 (emphasis on option).  I'd consider dropping v1 support to be a breaking change.

esalter:
Oh - yes, the API tokens are totally different.  V1 isn't compatible with V2.

esalter:
@andrioli Did you make the changes to the notification part locally?  Want to push your change to this PR?

andrioli:
@esalter I made a patch in your fork (https://github.com/esalter/monolog/pull/1)

Note that in [V1](https://www.hipchat.com/docs/api/method/rooms/message) the **notify** parameter is 0 or 1. On the other hand, [V2](https://www.hipchat.com/docs/apiv2/method/send_room_notification) uses strings "false" or "true"

If you want to apply the patch directly, you can found the diff file [here](https://gist.github.com/andrioli/a872856a1669cb0e226e)

esalter:
Merged in.  @Seldaek let me know if this looks ok, I'll squash the commits down first.

andrioli:
@esalter we forget to fix the unit tests

esalter:
Ah whoops.  I'll fix it up when i get some time.

andrioli:
@Seldaek I think is all right now :)

Seldaek:
Thanks everyone

PR #527, opened 2015-03-13
Add AirbrakeHandler

z38:
This PR adds `AirbrakeHandler` to send logs to [Airbrake][1] or any self-hosted [Errbit][2] instance.

At the moment the constructor takes an `Airbrake\Client` object (very similar to `RollbarHandler`). Just leave me a message if I should do the instantiation and configuration of the `Client` object in the constructor.

Any feedback is welcome!

[1]: https://airbrake.io/
[2]: http://errbit.github.io/errbit/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
         "doctrine/couchdb": "~1.0@dev",
         "aws/aws-sdk-php": "~2.4, >2.4.8",
         "videlalvaro/php-amqplib": "~2.4",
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks for the PR, but as I am trying to reduce the amount of core handlers I have to support and especially don't want to add more optional dependencies, I think you should rather release this as a standalone handler so that you can require the systemd extension explicitly and then add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

PR #529, opened 2015-03-18
Add strpos based TestHandler methods

rtuin:
These methods allow for writing tests with partial log matches. For example:

```php
$this->assertTrue(
   $this->logHandler->hasErrorThatContains('The connection to the server has failed')
);

$this->assertTrue(
   $this->logHandler->hasErrorThatContains('status 500')
);
```


rtuin:
@Seldaek what do you think of this improvement?

Seldaek:
Sure why not :) Thanks

Issue #416, opened 2014-09-04
Add php-console handler


barbushin:
Hi

There is a Google Chrome extension - PHP Console https://chrome.google.com/webstore/detail/php-console/nfhmhhlpfleoednkpnnnkolmclajemef with ~60k active, server side library https://github.com/barbushin/php-console, and having integration modules with Symfony, Yii, Silex, Laravel and Drupal frameworks.

Is it possible to develop monolog handler for php-console? I can do my best to help you with it.

Thank you!

barbushin:
@Seldaek Hi, Jordi. I submitted pull request https://github.com/Seldaek/monolog/pull/533 with `PHPConsoleHandler` and `PHPConsoleHandlerTest`. I'm very worried that after 3 weeks there is no reaction from you.

I now you're very busy, but please just tell me if it's everything okay with https://github.com/Seldaek/monolog/pull/533 or there is something I can improve? 

Thank you

stof:
this should be closed as #533 got merged.

PR #532, opened 2015-03-30
Add PHPConsoleHandler to debug in Google Chrome

barbushin:
There is a [PHPConsoleHandler](https://github.com/barbushin/monolog/blob/php-console-handler/src/Monolog/Handler/PHPConsoleHandler.php) that integrates Monolog with very popular Google Chrome extension [PHP Console](https://chrome.google.com/webstore/detail/php-console/nfhmhhlpfleoednkpnnnkolmclajemef) and its server library [php-console](https://github.com/barbushin/php-console). 

There is also [PHPConsoleHandlerTest](https://github.com/barbushin/monolog/blob/php-console-handler/tests/Monolog/Handler/PHPConsoleHandlerTest.php) with 25 tests  that are passed on PHP 5.3, 5.4 and 5.5.

I keep code style compatible with PSR1/2 and added documentation notes.

Referenced to #416

### DIFF
--- a/README.mdown
+++ b/README.mdown
@@ -154,6 +154,8 @@ Handlers
   inline `console` messages within Chrome.
 - _BrowserConsoleHandler_: Handler to send logs to browser's Javascript `console` with
   no browser extension required. Most browsers supporting `console` API are supported.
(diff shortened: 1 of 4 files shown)


barbushin:
There was a mistake in composer.json with minimal PHP Console version, so tests was not passed. Will create new correct pull request.

PR #535, opened 2015-04-05
Making SlackHandler more extendable

dec5e:
Now it is impossible to add additional data fields to Slack request without copying large pieces of code or without dirty hacks. Fixing this.


Seldaek:
Thanks

PR #539, opened 2015-04-09
Allow use of formatted message in PushoverHandler

djsipe:
This change allows you to use the "formatted" message from your formatter.  This is helpful if you want to create a formatter that would include *some* context information beyond the message itself.

Since the default `LineFormatter` would be too verbose, I added a `useFormattedMessage()` method that accepts a boolean and would turn the use of the formatter on or off. This way, if you want to use a custom formatter, you can enable the functionality explicitly...but it will still work the same out of the box otherwise with just the message getting sent to Pushover.


Seldaek:
Thanks

PR #541, opened 2015-04-14
made raven user context less destructive

smazurov:
as discussed in c1fd9cddf2f2d4f49dc56cb647681ee086c6fca3, the `$this->ravenClient->user_context(null)` was too destructive in a hybrid use case. This PR fixes it. 

Changes made:

* changed logic from "reset first" into a "reset if user context exists"
* added assertions that verify newly discovered use cases won't break
* used single quotes for strings

### DIFF
--- a/src/Monolog/Handler/RavenHandler.php
+++ b/src/Monolog/Handler/RavenHandler.php
@@ -127,8 +127,7 @@ class RavenHandler extends AbstractProcessingHandler
      */
     protected function write(array $record)
     {
-        // ensures user context is empty
-        $this->ravenClient->user_context(null);
+        $previousUserContext = false;
         $options = array();
         $options['level'] = $this->logLevels[$record['level']];
         $options['tags'] = array();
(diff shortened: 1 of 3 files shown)


src/Monolog/Handler/RavenHandler.php
```diff
@@ -160,11 +160,14 @@ protected function write(array $record)
         if (isset($record['context']['exception']) && $record['context']['exception'] instanceof \Exception) {
             $options['extra']['message'] = $record['formatted'];
             $this->ravenClient->captureException($record['context']['exception'], $options);
+        } else {
+            $this->ravenClient->captureMessage($record['formatted'], array(), $options);
+        }
 
-            return;
+        if ($previousUserContext !== null) {
+            $this->ravenClient->user_context($previousUserContext);
```
tremby: If the previous user context was null, shouldn't it be put back to null?

smazurov: thanks, added a test for that and fixed the issue.

Seldaek:
Thanks!

PR #546, opened 2015-04-22
Update NormalizerFormatter.php

draev:
Some objects implement magic __toString method
Allow to print this kind of objects in the logger


draev:
@draev is fixing tests..

Seldaek:
Looks good but yes fixing tests would be nice

Seldaek:
Well actually no it's not good, it should check if the object does not implement [JsonSerializable](http://php.net/manual/en/class.jsonserializable.php) before and if it does skip the __toString check because json_encode will do the right thing.

draev:
yep, good observation ;)

Seldaek:
Thanks!

PR #548, opened 2015-04-25
Removed a small code duplication, and added Logger::toMonologLevel() tests.

breda:
This is a rather small change, but keeping things DRY is always a good thing.
I removed the duplicated logic and replaced it with a `Monolog\Logger::toMonologLevel()` call.
Plus added the tests that cover that method.


Seldaek:
Thanks

PR #550, opened 2015-04-26
Make PHP7 and HHVM builds optional

JeroenDeDauw:
They failed in the last bunch of TravisCI builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,11 @@ php:
   - 7
   - hhvm
 
+matrix:
+  allow_failures:
+    - php: 7
+    - php: hhvm
+
 before_script:
   - if [ "`phpenv version-name`" != "hhvm" ]; then echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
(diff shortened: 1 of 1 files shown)


stof:
The HHVM failure seems a big in PHPUnit_Mock_Object. It should be reported to PHPUnit.

For PHP 7, there is 2 things to change:

- we should not try to enable the AMQP and Mongo extensions, as done for HHVM, given they are not yet compatible with PHP 7
- we should report an issue to Elastica about the PHP 7 compatibility

stof:
For Elastica, I reported it at https://github.com/ruflin/Elastica/issues/825

JeroenDeDauw:
By no means am I suggesting that these issues should not be fixed. However, since doing so takes time, it does seem good to mark these builds as optional until they can be made required again.

stof:
@JeroenDeDauw I totally agree about making these optional. However, if issues are not reported upstream, they might never get fixed.

bd808:
@stof The HHVM failure is a known issue for both [PHPUnit](https://github.com/sebastianbergmann/phpunit-mock-objects/issues/223) and [HHVM](https://github.com/facebook/hhvm/issues/5170). The underlying HHVM bug has been marked as fixed but apparently isn't in the builds used by Travis yet.

stof:
@bd808 this is because the fix in HHVM is not yet released

Seldaek:
Build is now fixed

PR #554, opened 2015-04-29
Fluentd handler, Composer self-update for travis

yegortokmakov:
Handler for fluentd.
Composer self-update for travisCI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ php:
 before_script:
   - if [ "`phpenv version-name`" != "hhvm" ]; then echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
   - if [ "`phpenv version-name`" != "hhvm" ]; then echo "extension = amqp.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
+  - composer self-update
   - composer install --prefer-source
 
 script: ./vendor/bin/phpunit
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,7 @@
         "graylog2/gelf-php": "~1.0",
         "raven/raven": "~0.5",
         "ruflin/elastica": "0.90.*",
+        "fluent/logger": "0.3.*",
         "doctrine/couchdb": "~1.0@dev",
         "aws/aws-sdk-php": "~2.4, >2.4.8",
         "videlalvaro/php-amqplib": "~2.4",
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
         "ext-mongo": "Allow sending log messages to a MongoDB server",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
         "rollbar/rollbar": "Allow sending log messages to Rollbar",
+        "fluent/logger": "Allow sending log messages to Fluentd",
         "php-console/php-console": "Allow sending log messages to Google Chrome"
     },
     "autoload": {
--- a/src/Monolog/Handler/FluentHandler.php
+++ b/src/Monolog/Handler/FluentHandler.php
@@ -27,30 +27,22 @@ class FluentHandler extends AbstractProcessingHandler
     /**
      * @var FluentLogger
      */
-    protected $logger;
+    private $logger;
 
     /**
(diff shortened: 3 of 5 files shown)


yegortokmakov:
[RuntimeException] Could not load package phpspec/prophecy in http://packagist.org: [UnexpectedValueException] Could not parse version constraint ^1.0.2: Invalid version string "^1.0.2"

Fails because of phpunit dependency.

src/Monolog/Handler/FluentHandler.php
```diff
@@ -0,0 +1,69 @@
...
+     * Initialize Handler
+     *
+     * @param FluentLogger $logger
+     * @param bool|string $host
+     * @param int $port
+     * @param int $level
+     * @param bool $bubble
+     */
+    public function __construct(
+        $logger = null,
```
xabbuh: I would add a type hint on `FluentLogger` here.

src/Monolog/Handler/FluentHandler.php
```diff
@@ -0,0 +1,69 @@
...
+        $port   = FluentLogger::DEFAULT_LISTEN_PORT,
+        $level  = Logger::DEBUG,
+        $bubble = true
+    )
+    {
+        parent::__construct($level, $bubble);
+
+        if (is_null($logger)) {
+            $logger = new FluentLogger($host, $port);
+        }
```
xabbuh: Not sure if that really makes sense. It over-complicates the constructor and requires the user to pass values that will never be used just because they want to, for example, change the logging level.

composer.json
```diff
@@ -37,6 +37,7 @@
         "ext-mongo": "Allow sending log messages to a MongoDB server",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
         "rollbar/rollbar": "Allow sending log messages to Rollbar",
+        "fluent/logger": "Allow sending log messages to Fluentd",
```
xabbuh: You should add the package to `require-dev` too.

src/Monolog/Handler/FluentHandler.php
```diff
@@ -0,0 +1,69 @@
...
+
+        $this->logger = $logger;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function write(array $record)
```
xabbuh: `write()` is `protected`

src/Monolog/Handler/FluentHandler.php
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * @author Dai Akatsuka <d.akatsuka@gmail.com>
+ * @author Yegor Tokmakov <yegor@tokmakov.biz>
+ */
+class FluentHandler extends AbstractProcessingHandler
+{
+    /**
+     * @var FluentLogger
+     */
+    protected $logger;
```
xabbuh: should be `private`

.travis.yml
```diff
@@ -12,6 +12,7 @@ php:
 before_script:
   - if [ "`phpenv version-name`" != "hhvm" ]; then echo "extension = mongo.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
   - if [ "`phpenv version-name`" != "hhvm" ]; then echo "extension = amqp.so" >> ~/.phpenv/versions/$(phpenv version-name)/etc/php.ini; fi
+  - composer self-update
```
Taluu: not important, see http://docs.travis-ci.com/user/languages/php/#Installing-Composer-packages

yegortokmakov: sometimes 30-60 days is not enough :) today test was failing for this PR because of old composer version

Taluu: maybe, but it is not your concern (or the lib actually), it is Travis'

yegortokmakov:
@xabbuh @Taluu I updated code, but left composer update for Travis

yegortokmakov:
any chance someone can review PR?

Seldaek:
Thanks for the PR, but as I am trying to reduce the amount of core handlers I have to support and especially don't want to add more optional dependencies, I think you should rather release this as a standalone handler so that you can require the systemd extension explicitly and then add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

PR #556, opened 2015-05-02
Saving some memory for echo.

dseguy:


Seldaek:
I doubt this matters a whole lot but why not I guess..

PR #557, opened 2015-05-04
Update LogglyHandler.php

barchard:
Add the ability to set multiple tags via a single call. Tags are now an array and imploded instead of appended. This prevents duplicate tags from being added.

### DIFF
--- a/src/Monolog/Handler/LogglyHandler.php
+++ b/src/Monolog/Handler/LogglyHandler.php
@@ -19,6 +19,7 @@ use Monolog\Formatter\LogglyFormatter;
  *
  * @author Przemek Sobstel <przemek@sobstel.org>
  * @author Adam Pancutt <adam@pancutt.com>
+ * @author Gregory Barchard <gregory@barchard.net>
  */
 class LogglyHandler extends AbstractProcessingHandler
 {
@@ -28,7 +29,7 @@ class LogglyHandler extends AbstractProcessingHandler
 
     protected $token;
 
-    protected $tag;
+    protected $tags = array();
 
     public function __construct($token, $level = Logger::DEBUG, $bubble = true)
     {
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/LogglyHandler.php
```diff
@@ -43,12 +44,16 @@ public function __construct($token, $level = Logger::DEBUG, $bubble = true)
...
+        $tag = !empty($tag) ? $tag : array();
+        $this->tags = is_array($tag) ? $tag : array($tag);
     }
 
     public function addTag($tag)
     {
-        $this->tag = (strlen($this->tag) > 0) ? $this->tag .','. $tag : $tag;
+        if (!empty($tag)) {
+            $tag = is_array($tag) ? $tag : array($tag);
+            $this->tags = array_merge($this->tags, $tag);
```
Seldaek: This does not really prevent duplicates from being added, unless you do array_unique() around the array_merge

src/Monolog/Handler/LogglyHandler.php
```diff
@@ -28,7 +29,7 @@ class LogglyHandler extends AbstractProcessingHandler
 
     protected $token;
 
-    protected $tag;
+    protected $tags = array();
```
stof: renaming a protected property is a BC break for code extending the class

src/Monolog/Handler/LogglyHandler.php
barchard: @Seldaek you're right. i forgot this is a list not a dictionary. Ill update it to add array_unique

src/Monolog/Handler/LogglyHandler.php
barchard: @stof good point. Ill change this.

Seldaek:
Thanks

PR #559, opened 2015-05-05
Multi array

snoob:
When we have more than other one array (or non scalar type) the log message after replacement looks like "some string ... [array] [array]". 

I think we should add the key to identity the array or maybe we can create a special case for array using json_decode or implode?


Seldaek:
The point of the PSR4 formatter is to be a very simple and succint approach to placeholders. The main reason placeholders were added in PSR4 was to do translation so I don't really feel like arrays make any sense there and that's why we normalize them to a simple [array]. I would strongly suggest you use formatters like the LineFormatter or NormalizerFormatter that can serialize array data to json or similar if that's what you need. Or if you want a fancier version of the PSR4 processor you are welcome to use a custom one, but I'd rather keep the one we have simple.

PR #564, opened 2015-05-13
Add SystemdJournalHandler

eagleoneraptor:
I just wanted to propose a handler to log to `systemd-journal` service, this is going to give us all the benefits of this service (better search and indexing, better integration with the OS, richer metadata for log records).

To use it, you're gonna need to have [php-systemd extension](https://github.com/systemd/php-systemd) installed in your system.

I have one concern for this PR though, I had to make `checkSystemdExtension` and `sendToJournal` public methods instead of private or protected to make it easy to mock in the tests (we don't want for sure to send real log records during test execution). I could wrap those methods functionality in a separated class but it looked like too much just for test cases stuff. You tell me :)

Thanks!

### DIFF
--- a/src/Monolog/Handler/SystemdJournalHandler.php
+++ b/src/Monolog/Handler/SystemdJournalHandler.php
@@ -1,0 +1,104 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Logger;
+
+/**
+ * Logs to systemd-journald service.
+ *
+ * This handler requires the sd_journal_send function from php-systemd
+ * extension (https://github.com/systemd/php-systemd).
+ *
+ * @author Damián Nohales <damiannohales@gmail.com>
+ */
+class SystemdJournalHandler extends AbstractPosixBasedHandler
+{
+    protected $extraFields;
+
+    /**
+     * @param array   $extraFields Extra fields to send to the journal, Monolog will prepend the
+     *                             MONOLOG prefix and will normalize each field name according to
+     *                             the restrictions pointed in sd_journal_print(3) manpage.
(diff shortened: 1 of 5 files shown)


src/Monolog/Handler/SystemdJournalHandler.php
```diff
@@ -0,0 +1,104 @@
...
+
+        parent::__construct($level, $bubble);
+
+        $this->extraFields = $extraFields;
+    }
+
+    /**
+     * Check if the php-systemd extension is properly enabled.
+     */
+    public function checkSystemdExtension()
```
staabm: why do you need this method? why not just inline this code in __construct()?

src/Monolog/Handler/SystemdJournalHandler.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $this->sendToJournal($fields);
+    }
+
+    /**
+     * Send log entry to the journal.
+     *
+     * @param array $fields Log entry fields
+     */
+    public function sendToJournal(array $fields)
```
staabm: testing this method doesn't really make sense, does it?

staabm: you should instead add a test which searched in systemd if you write succeeded

eagleoneraptor: Ok, so, the methods `checkSystemdExtension` and `sendToJournal` are going to be mocked (that's why I wrapped such simple code in separated methods), I thought it was a better idea to skip the journal mechanism (send the record to journal) while be able to test policy like fields generation and that stuff.

So, you say that we should skip `SystemdJournalHandlerTest` when `php-systemd` isn't available and when is available we need to query the journal for the created logs instead of mocking, right?

staabm: I think so. atm when you test this method  you are actually testing whether php is able to invoke the systemd function given your parameters (which is not business of this class).

In case we want this class test-covered when the php-ext is not available you could also create a subclass only for testing purposes (residing in the test-package) so you dont need this ugly public methods.

Seldaek:
I think you should rather release this as a standalone handler so that you can require the systemd extension explicitly and then add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

PR #565, opened 2015-05-19
Fix: Wrong raven/raven version

mabrahamde:
RavenHandler uses Raven_Client::user_context(). This method is available since 0.8.0.

see https://github.com/getsentry/raven-php/commit/7dd1fb3a353eb2b456336a996e5265fc5756a085

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require-dev": {
         "phpunit/phpunit": "~4.0",
         "graylog2/gelf-php": "~1.0",
-        "raven/raven": "~0.5",
+        "raven/raven": "~0.8",
         "ruflin/elastica": "0.90.*",
         "doctrine/couchdb": "~1.0@dev",
         "aws/aws-sdk-php": "~2.4, >2.4.8",


mabrahamde:
Anything wrong with this one? It just fixes a small bug inside dependency management. Build status is same as mainline.

smazurov:
`~0.5` translates to `>=0.5.0 < 1.0.0` in composer. [See this](https://getcomposer.org/doc/01-basic-usage.md#package-versions)

Seldaek:
Indeed it seems just fine as it is.

mabrahamde:
I does not make sense to require raven/raven versions < 0.8.0. Monolog is using functionality which is missing there. 

If there is a raven/raven 0.5.x package installed already...and you installing monolog via composer afterwards

- the unsupported 0.5.x version of raven/raven will not get updated
- no warning/conflict will be displayed during installation
- logging to sentry will fail.

Monolog requires raven/raven version >= 0.8.0 < 1.0.0 which translates to ~0.8.

Seldaek:
Right sorry for the misunderstanding, I checked and indeed it segfaults with 0.5 - 0.7 versions.

mabrahamde:
:+1:

PR #566, opened 2015-05-21
Add Logger::setTimezone()

bd808:
Add a setter method to allow changing the timezone used when creating
new log records. This provides a means to correct a behavioral change
introduced by 6cbdc04 where use of the PHP runtime default timezone
provided by `date_default_timezone_get()` was introduced. In most
environments the new default behavior is preferable, but if an
application allows the default timezone to vary between requests it
causes confusing log output.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -614,4 +614,14 @@ class Logger implements LoggerInterface
     {
         return $this->addRecord(static::EMERGENCY, $message, $context);
     }
+
+    /**
+     * Set the timezone to be used for the timestamp of log records.
+     *
+     * @param string $tz Timezone name
+     */
+    public static function setTimezone($tz)
+    {
+        self::$timezone = new \DateTimeZone($tz);
+    }
 }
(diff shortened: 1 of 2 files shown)


bd808:
See https://phabricator.wikimedia.org/T99581 for a concrete use case.

src/Monolog/Logger.php
```diff
@@ -614,4 +614,14 @@ public function emergency($message, array $context = array())
     {
         return $this->addRecord(static::EMERGENCY, $message, $context);
     }
+
+    /**
+     * Set the timezone to be used for the timestamp of log records.
+     *
+     * @param string $tz Timezone name
+     */
+    public static function setTimezone($tz)
```
staabm: this should take the object instead of a string, shouldn't it?

bd808: I suppose it could. Any particular reason that seems like a better API to you?

bd808:
The CI failures seem unrelated:
* PHP7 is failing due to "Bool" being considered a reserved word and an invalid class name in Elastica
* HHVM is failing due to a [known HHVM+PHPUnit bug](https://github.com/facebook/hhvm/issues/5170) for mocks

src/Monolog/Logger.php
Seldaek: Since ultimately we end up with a DateTimeZone object I'd say yes accepting that makes it easier to handle validation and so on upstream if needed, and people might already be working with a DTZ object in their app anyway. I'll do the change after merge though.

Seldaek:
Thanks!

PR #570, opened 2015-06-02
CodeIgniterMailerHandler.php

KashifChandio:
Mail Handler for CodeIgniter.

### DIFF
--- a/src/Monolog/Handler/CodeIgniterMailerHandler.php
+++ b/src/Monolog/Handler/CodeIgniterMailerHandler.php
@@ -1,0 +1,200 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Logger;
+
+/**
+ * NativeMailerHandler uses the mail() function to send the emails
+ *
+ * @author Christophe Coevoet <stof@notk.org>
+ * @author Mark Garrett <mark@moderndeveloperllc.com>
+ */
+class CodeIgniterMailerHandler extends MailHandler
+{
+    /**
+     * The email addresses to which the message will be sent
+     * @var array
+     */
+    protected $_to;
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/CodeIgniterMailerHandler.php
```diff
@@ -0,0 +1,200 @@
...
+ * @author Mark Garrett <mark@moderndeveloperllc.com>
+ */
+class CodeIgniterMailerHandler extends MailHandler
+{
+    /**
+     * The email addresses to which the message will be sent
+     * @var array
+     */
+    protected $_to;
+    protected $_from;
```
stof: no leading underscore please

src/Monolog/Handler/CodeIgniterMailerHandler.php
```diff
@@ -0,0 +1,200 @@
...
+    protected $encoding = 'utf-8';
+
+    /**
+     * @param string|array $to             The receiver of the mail
+     * @param string       $subject        The subject of the mail
+     * @param string       $from           The sender of the mail
+     * @param integer      $level          The minimum logging level at which this handler will be triggered
+     * @param boolean      $bubble         Whether the messages that are handled can bubble up the stack or not
+     * @param int          $maxColumnWidth The maximum column width that the message lines will have
+     */
```
stof: the phpdoc should not be separated from the method

src/Monolog/Handler/CodeIgniterMailerHandler.php
```diff
@@ -0,0 +1,200 @@
...
+    public function addHeader($headers)
+    {
+        foreach ((array) $headers as $header) {
+            if (strpos($header, "\n") !== false || strpos($header, "\r") !== false) {
+                throw new \InvalidArgumentException('Headers can not contain newline characters for security reasons');
+            }
+            $this->headers[] = $header;
+        }
+
+        return $this;
```
stof: this should not be fluent IMO. It is not consistent with other parts of Monolog

Seldaek:
Thanks for the PR but IMO this belongs in CI directly more than in monolog. Or if they don't want it then having it in a standalone package integrating the two together is an option, but having a CI-specific package in here makes little sense.

PR #576, opened 2015-06-08
Fixed an error in elastica requirements (missing comma)

### LINKED ISSUE #575: Error in composer.json
I just updated my developement environment and got these two error messages which show that the reqirements in the composer.json of seldaek/monolog have an error.

    [RuntimeException]                                                           
      Could not load package monolog/monolog in http://packagist.org: [Unexpected  
      ValueException] Could not parse version constraint >=0.90 <3.0: Invalid version string "0.90 <3.0"                                                      
                                                                               
                                                                
    [UnexpectedValueException]                                                   
      Could not parse version constraint >=0.90 <3.0: Invalid version string "0.9  0 <3.0" 


It seems that this line (https://github.com/Seldaek/monolog/blob/master/composer.json#L23) is the error:

    "ruflin/elastica": ">=0.90 <3.0",
a simple comma between the two version constrains fixes the error.

    "ruflin/elastica": ">=0.90, <3.0",

I have attached a pull request which fixes the problem.

Reflic:
Resolves #575

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "phpunit/phpunit": "~4.5",
         "graylog2/gelf-php": "~1.0",
         "raven/raven": "~0.8",
-        "ruflin/elastica": ">=0.90 <3.0",
+        "ruflin/elastica": ">=0.90, <3.0",
         "doctrine/couchdb": "~1.0@dev",
         "aws/aws-sdk-php": "^2.4.9",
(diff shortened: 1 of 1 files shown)


stof:
this is not necessary. See my comment on the issue

stof:
btw, you will still get an error with our old composer version because another requirement uses the caret operator which was implemented at the same time

PR #578, opened 2015-06-10
Feature#logstash udp handler

supre:
A basic UDP handler for logstash


Seldaek:
As far as I can tell this is a generic UDP handler, I think you can simply do that using a SocketHandler with `udp://127.0.0.1:1234` or something as connection string and then setting the formatter to the logstash formatter. Or is there something not working with that?

PR #583, opened 2015-06-20
TestHandler improvements

austinhyde:
A couple of improvements to `Monolog\Handler\TestHandler`:

* DRY up record searching into a predicate-based method: `hasRecordThatPasses($predicate, $level)`
* DRY up boilerplate methods with magic `__call()`: e.g. `hasInfoThatContains($message)` => `hasRecordThatContains($message, Logger::INFO)`
  * Add `@method` annotations for the pseudo-methods
* Add a Regular Expression matcher for records: `hasRecordThatMatches($regex, $level)`
* Add tests for the above

### DIFF
--- a/src/Monolog/Handler/TestHandler.php
+++ b/src/Monolog/Handler/TestHandler.php
@@ -19,6 +19,42 @@ use Monolog\Logger;
  * It records all records and gives you access to them for verification.
  *
  * @author Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * @method boolean hasEmergency($record)
+ * @method boolean hasAlert($record)
+ * @method boolean hasCritical($record)
+ * @method boolean hasError($record)
+ * @method boolean hasWarning($record)
+ * @method boolean hasNotice($record)
+ * @method boolean hasInfo($record)
+ * @method boolean hasDebug($record)
+ * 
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/TestHandler.php
```diff
@@ -30,153 +75,48 @@ public function getRecords()
...
-        return isset($this->recordsByLevel[Logger::NOTICE]);
-    }
-
-    public function hasInfoRecords()
-    {
-        return isset($this->recordsByLevel[Logger::INFO]);
-    }
-
-    public function hasDebugRecords()
+    protected function hasRecordRecords($level)
```
Seldaek: Maybe hasRecords would suffice? :)

austinhyde: It's a side effect of how it resolves the method names - `__call` simply replaces the severity with 'Record' and looks for that method. So, `hasInfoRecords` resolves to `hasRecordRecords`. I thought it was a minor tradeoff since the equivalent wasn't present before this refactor anyways.

If you'd prefer, I can change that, or add `hasRecords` that just forwards to `hasRecordRecords`

Seldaek: Nah it's not really a huge deal don't worry about it.

Issue #581, opened 2015-06-18
create logdir if not exists


dominikzogg:
Hello @Seldaek

Would be cool if the StreamHandler or another handler would be able to create the log directory (if its a path, not a stream), if it not exists.

see: https://github.com/silexphp/Silex/pull/1183

If you like the idea i would prepare a pullrequest.

Regards @dominikzogg

Seldaek:
See #555 - but if you can provide a safe implementation I am not really against it..

dominikzogg:
@Seldaek well a resource definition is a string, which has :// within, or a resource itself, isn't it?

Seldaek:
'foo.log' is valid, 'file://foo.log' is valid but has ://, ftp://foo.log is valid but should perhaps not get mkdir called on it (well I guess with ftp it works but not necessarily with other streams). I think doing a is_dir+mkdir check for file:// or no-protocol paths is probably safe.

dominikzogg:
@Seldaek You mean only paths with file:// should get considered? Would be a compromise i could live with, should be in documentation as well. Should i prepare a pull request which works like this?

Seldaek:
No I mean paths with file:// and those that don't have a protocol defined, like "foo.log" has no `XXX://` and as such you know php defaults to the file:// stream. And yes if you like to send a PR please go ahead :)

PR #584, opened 2015-06-21
stream handler creates a directory if its possible

dominikzogg:
referers to #581

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -35,6 +35,7 @@ class StreamHandler extends AbstractProcessingHandler
      * @param int|null        $filePermission Optional file permissions (default (0644) are only for owner read/write)
      * @param Boolean         $useLocking     Try to lock log file before doing any writes
      *
+     * @throws \Exception                If a missing directory is not buildable
      * @throws \InvalidArgumentException If stream is not a resource or string
      */
     public function __construct($stream, $level = Logger::DEBUG, $bubble = true, $filePermission = null, $useLocking = false)
@@ -43,6 +44,16 @@ class StreamHandler extends AbstractProcessingHandler
         if (is_resource($stream)) {
             $this->stream = $stream;
         } elseif (is_string($stream)) {
+            $dir = $this->getDirFromStream($stream);
+            if (null !== $dir && !is_dir($dir)) {
+                $this->errorMessage = null;
+                set_error_handler(array($this, 'customErrorHandler'));
+                $status = mkdir($dir, 0777, true);
+                restore_error_handler();
+                if (false === $status) {
+                    throw new \UnexpectedValueException(sprintf('There is no existing directory at "%s" and its not buildable: '.$this->errorMessage, $dir));
+                }
+            }
             $this->url = $stream;
         } else {
             throw new \InvalidArgumentException('A stream must either be a resource or a string.');
@@ -99,6 +110,25 @@ class StreamHandler extends AbstractProcessingHandler
 
     private function customErrorHandler($code, $msg)
     {
-        $this->errorMessage = preg_replace('{^fopen\(.*?\): }', '', $msg);
+        $this->errorMessage = preg_replace('{^(fopen|mkdir)\(.*?\): }', '', $msg);
+    }
+
+    /**
+     * @param string $stream
+     *
+     * @return null|string
+     */
+    private function getDirFromStream($stream)
+    {
+        $pos = strpos($stream, '://');
+        if ($pos === false) {
+            return dirname($stream);
+        }
+
+        if ('file://' === substr($stream, 0, 7)) {
+            return dirname(substr($stream, 7));
+        }
+
+        return;
     }
 }
--- a/tests/Monolog/Handler/StreamHandlerTest.php
+++ b/tests/Monolog/Handler/StreamHandlerTest.php
@@ -112,7 +112,51 @@ class StreamHandlerTest extends TestCase
      */
     public function testWriteNonExistingResource()
     {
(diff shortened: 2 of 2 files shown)


dominikzogg:
@Seldaek What do you think about this solution?

src/Monolog/Handler/StreamHandler.php
```diff
@@ -97,8 +108,32 @@ protected function write(array $record)
...
 
-    private function customErrorHandler($code, $msg)
+    private function customMkDirErrorHandler($code, $msg)
+    {
+        $this->errorMessage = $msg;
+    }
+
+    private function customFopenErrorHandler($code, $msg)
     {
         $this->errorMessage = preg_replace('{^fopen\(.*?\): }', '', $msg);
```
Seldaek: I think you could use the same customErrorHandler for both, just change fopen here to be `(fopen|mkdir)`, the regex only cleans up php's error message which is like: `foo(arg): ACTUAL MESSAGE`.

src/Monolog/Handler/StreamHandler.php
```diff
@@ -43,6 +44,16 @@ public function __construct($stream, $level = Logger::DEBUG, $bubble = true, $fi
...
             $this->stream = $stream;
         } elseif (is_string($stream)) {
+            $dir = $this->getDirFromStream($stream);
+            if (null !== $dir && !is_dir($dir)) {
+                $this->errorMessage = null;
+                set_error_handler(array($this, 'customMkDirErrorHandler'));
+                $status = mkdir($dir, 0755, true);
+                restore_error_handler();
+                if (false === $status) {
+                    throw new \UnexpectedValueException(sprintf('There is no existing directory at "%s" and its not buildable.'.$this->errorMessage, $dir));
```
Seldaek: Needs a space in the string before $this->errorMessage

src/Monolog/Handler/StreamHandler.php
```diff
@@ -43,6 +44,16 @@ public function __construct($stream, $level = Logger::DEBUG, $bubble = true, $fi
         if (is_resource($stream)) {
             $this->stream = $stream;
         } elseif (is_string($stream)) {
+            $dir = $this->getDirFromStream($stream);
+            if (null !== $dir && !is_dir($dir)) {
+                $this->errorMessage = null;
+                set_error_handler(array($this, 'customMkDirErrorHandler'));
+                $status = mkdir($dir, 0755, true);
```
Seldaek: Mode should be 0777, php applies the umask on that automatically.

dominikzogg:
@Seldaek fixes are made

dominikzogg:
@Seldaek any other changes needed?

src/Monolog/Handler/StreamHandler.php
```diff
@@ -99,6 +110,25 @@ protected function write(array $record)
...
+     */
+    private function getDirFromStream($stream)
+    {
+        $pos = strpos($stream, '://');
+        if ($pos === false) {
+            return dirname($stream);
+        }
+
+        if ('file:///' === substr($stream, 0, 8)) {
+            return DIRECTORY_SEPARATOR.dirname(substr($stream, 8));
```
Seldaek: That DIRECTORY_SEPARATOR seems wrong to me on windows at least. IMO you should just check for `file://` and not with triple slash and then `return substr($stream, 7)` should do, because as it stands you break support for relative paths as far as I can see.

dominikzogg: @Seldaek is there really a filemapper file://, cause in my tests i only get file:/// running?

Seldaek: I guess there is no relative path support, but on windows both  of these are valid, and your code would break the first:

```
$ php -r 'var_dump(is_dir("file://c:/users/seld"));'
bool(true)
$ php -r 'var_dump(is_dir("file:///c:/users/seld"));'
bool(true)
```

dominikzogg:
@Seldaek My fault, is that i id not get, that the third / was the start of a unix like path, should be fixed

Seldaek:
Great thanks, and no worries it's good when someone learns something :)

dominikzogg:
@Seldaek thank you for merging

dominikzogg:
@Seldaek would be cool if you could add a new tag

Seldaek:
1.15 is out

dominikzogg:
@Seldaek thank you

TerjeBr:
This PR makes it impossible for us to upgrade monolog to the new version, because it breaks our build.

The command `composer.phar install` will at the end try to clean the cache by running a symfony2 command, and the symfony2 kernel builds the DI container with all the monolog handlers. On our test box that cleans the dev env, and that was harmless before. (At all other times our test box only uses the test env.) But because some directories that exists in the dev env, does not exist on our test box, this command now fails on the test box, even if it is not using the dev env for anything. So all our tests fails at the stage of running `composer.phar install`.

Would it be possible to change the code so that if the log directory cannot be crated, it does not throw an exception?

dominikzogg:
@Seldaek should i look into this?

Seldaek:
@TerjeBr One question: if the dir can't be created, it should fail when writing to the file that it also can't create, or am I missing something?

TerjeBr:
@Seldaek if the application tries to write something to the log, yes then of course it should fail with an exception. It is just the case when the directory does not exist and nothing is written to the log I was asking you to accommodate.

Seldaek:
@TerjeBr yes but as far as I can tell in the StreamHandler it only tries to create the directory on the first write() call so that should work as expected for you. Ah but I see we fixed that since the last release, if you require `^1.15@dev` it  should work, just until the next release which I'll try to do shortly.

PR #585, opened 2015-06-22
Add support for Javascript output in BrowserConsoleHandler

daniel-adlantic:
Up to now, BrowserConsoleHandler only shows output when output header is text/html. However many of our applications also use application/javascript (and obsolete text/javascript). With this PR, generated javascript will also be returned, but without the script tags.


Seldaek:
Not such a common use case to generate JS with php requests I guess but patch looks good, thanks

PR #586, opened 2015-06-22
Update composer.json

gonzalodenari:
There was missing the comma to separate versions

[UnexpectedValueException]                                                          
  Could not parse version constraint >=0.90 <3.0: Invalid version string "0.90 <3.0"


Seldaek:
Just needs an update of composer to support the new constraints

PR #591, opened 2015-06-24
Fix typo in composer.json

trsteel88:


trsteel88:
Composer was failing with the following error.

```
 [UnexpectedValueException]                                                          
 Could not parse version constraint >=0.90 <3.0: Invalid version string "0.90 <3:.0"
The PHP Package Repository
```

However, updating composer resolved the problem.

PR #601, opened 2015-07-09
Fix link to seperate readme

Xosofox:
After readme being split into parts, link needs to go to dedicated chapter


Seldaek:
Thanks

PR #602, opened 2015-07-12
Adds ability to cap log size in redis

ninjapenguin:
100% backward compatible, defaults to not capping collection size.

Additional constructor param of $capSize added which will ensure logs
list is treated as a FILO queue

Test coverage on new functionality


Seldaek:
How about using `LTRIM $list -$capSize -1` so we can keep using RPUSH and don't revert the order of the list depending on capSize being present?

ninjapenguin:
Great idea! I've made the change  :)

Seldaek:
Great, looking good now, thanks

PR #605, opened 2015-07-14
Fix: Type should be bool

localheinz:
This PR

* [x] fixes a docblock where `type` should be `boolean` instead
* [x] renames a parameter from `$boolean` to `$parameter`


Seldaek:
Thanks

Issue #606, opened 2015-07-14
Monolog throws an error on read-only filesystems


will-in-wi:
We have a webserver where the filesystem is mounted readonly. The log dir is a symlink to a read-write part of the filesystem. Monolog, as of 1.15.0, tries to create the dir on every log action and raises an exception.

Line 51 of StreamHandler.php is the issue: https://github.com/Seldaek/monolog/blob/6dd76b944d42d7b27bb3d09cd0f6fa88a80817f5/src/Monolog/Handler/StreamHandler.php#L51

If I comment that out, it works. Could a check as to whether the folder already exists be added?

Thanks!

will-in-wi:
This ticket is wrong. On looking into it further, it turns out that https://github.com/Flynsarmy/Slim-Monolog/blob/master/Log/MonologWriter.php#L108 creates an unused monolog instance with an invalid log dir. Since the mkdir seems to be happen on instantiation, it fails.

I think this is probably a monolog bug. I don't expect to have a directory be created when I instantiate a class. I would expect it to occur on first log.

will-in-wi:
I took a shot at fixing this in https://github.com/Seldaek/monolog/pull/607. Tests all pass locally.

bladi87:
I fixed this in [Flynsarmy/Slim-Monolog](https://github.com/Flynsarmy/Slim-Monolog). 
Please do a composer update.
In version 1.0.1 you can call the constructor with an extra parameter, set it false, so it won't create an unused monolog instance.

will-in-wi:
I'd posit that the strategy here (whether or not my specific patch), is better than adding a merge param to Slim-Monolog. The situation where a developer might want the default `$name` parameter for `MonologWriter` but passes in a `StreamHandler` isn't handled. Also, if you did want to work around this in Slim-Monolog, you could just always merge the `$name` param since there are no side-effects to creating a string, and only set a default `StreamHandler` if one isn't passed. The `$merge` param strategy asks the developer to determine something that should probably be an internal detail.

The fundamental issue here is that instantiating a class shouldn't modify the system environment. This is an unexpected side-effect as demonstrated by the former assumption in Slim-Monolog that it could instantiate a throwaway class without negative repercussions. I think that this assumption is sound. Writing to the log is expected to change the system environment, and creating the directory at that point is expected.

Does this make sense?

Thanks for your work on this!

bladi87:
[...]

PR #607, opened 2015-07-14
Create directory on initial log write rather than class instantiation.

will-in-wi:
Fix for https://github.com/Seldaek/monolog/issues/606.

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -27,6 +27,7 @@ class StreamHandler extends AbstractProcessingHandler
     private $errorMessage;
     protected $filePermission;
     protected $useLocking;
+    private $dirCreated;
 
     /**
      * @param resource|string $stream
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/StreamHandler.php
```diff
@@ -131,4 +123,22 @@ private function getDirFromStream($stream)
...
+
+    private function createDir()
+    {
+        // Do not try to create dir if it has already been tried.
+        if ($this->dirCreated) return;
+
+        $dir = $this->getDirFromStream($this->url);
+        if (null !== $dir && !is_dir($dir)) {
+            $this->errorMessage = null;
+            set_error_handler(array($this, 'customErrorHandler'));
```
mimmi20: Is it nessesary to set the error handler twice? (In line 79 and in line 135)

Seldaek:
Looks good, thanks!

PR #609, opened 2015-07-15
Prevent Monolog from ever being introspected

xploSEoF:
Ensure the Monolog namespace is always ignored, and doesn't have to be defined in use.

### DIFF
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -30,10 +30,10 @@ class IntrospectionProcessor
 
     private $skipClassesPartials;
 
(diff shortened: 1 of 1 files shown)


src/Monolog/Processor/IntrospectionProcessor.php
```diff
@@ -30,10 +30,10 @@ class IntrospectionProcessor
 
     private $skipClassesPartials;
 
-    public function __construct($level = Logger::DEBUG, array $skipClassesPartials = array('Monolog\\'))
+    public function __construct($level = Logger::DEBUG, array $skipClassesPartials = [])
```
staabm: composer supports php 5.3, so you have to use `array()`

Seldaek:
Thanks

PR #611, opened 2015-07-17
Add more unit tests for Registry

KevinMGranger:
Coverage should now be 100%.

### DIFF
--- a/tests/Monolog/RegistryTest.php
+++ b/tests/Monolog/RegistryTest.php
@@ -59,4 +59,96 @@ class RegistryTest extends \PHPUnit_Framework_TestCase
             ),
         );
     }
+
+    /**
+     * @covers Monolog\Registry::clear
+     */
+    public function testClearClears()
+    {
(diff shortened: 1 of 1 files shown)


tests/Monolog/RegistryTest.php
```diff
@@ -59,4 +59,96 @@ public function hasLoggerProvider()
...
+     * @dataProvider removedLoggerProvider
+     * @covers Monolog\Registry::addLogger
+     * @covers Monolog\Registry::removeLogger
+     */
+    public function testRemovesLogger($loggerToAdd, $remove)
+    {
+        Registry::addLogger($loggerToAdd);
+        Registry::removeLogger($remove);
+
+        $this->setExpectedException('\InvalidArgumentException');
```
KevinMGranger: I set the exception in here instead of as an annotation because it's a test implementation detail, not part of the interface we're testing.

Seldaek:
Thanks

PR #615, opened 2015-07-21
Symlink today's file in rotation handler

EmmanuelVella:
It would be very useful to edit/tail today's log file without having to handle the date in the filename. This PR adds a simple symlink to the current log file. Ex: 

```
prod.log -> /var/www/acme/app/logs/prod-2015-07-21.log
prod-2015-07-21.log
prod-2015-07-20.log
prod-2015-07-19.log
...
```

### DIFF
--- a/src/Monolog/Handler/RotatingFileHandler.php
+++ b/src/Monolog/Handler/RotatingFileHandler.php
@@ -97,6 +97,13 @@ class RotatingFileHandler extends StreamHandler
         $this->url = $this->getTimedFilename();
         $this->nextRotation = new \DateTime('tomorrow');
 
+        // symlink today's file
+        if (file_exists($this->filename)) {
+            unlink($this->filename);
+        }
+
+        symlink($this->url, $this->filename);
+
         // skip GC of old logs if files are unlimited
         if (0 === $this->maxFiles) {
             return;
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/RotatingFileHandler.php
```diff
@@ -80,6 +85,10 @@ protected function write(array $record)
             $this->mustRotate = !file_exists($this->url);
         }
 
+        if (null === $this->mustSymlink) {
+            $this->mustSymlink = !is_link($this->filename);
+        }
```
Seldaek: I don't really get the point here, if the current day's file is not a link it gets symlinked, which means the feature isn't really optional. That might be fine, but what happens once it got rotated, the symlink will point to an old one?

EmmanuelVella: You are completely right, this was a mistake. I have updated/simplified my PR ; now the symlink is updated on each rotation.

Seldaek:
It's updated on each rotation but it's not created anymore when you don't rotate, that sounds like a mistake, especially since rotation doesn't typically happen unless a script runs through the midnight mark. Otherwise every request just opens the log of the current date and writes into it.

That said I am not sure if this features makes sense at all though, as it means it might cause failures on windows and adds overall complexity for not so much gain.

I really think using logrotate is a better alternative than this handler for any kind of serious environment, so I would rather keep it simple.

PR #617, opened 2015-07-27
Added Logger::setHandlers()

sminnee:
Logger::setHandlers() is intended to help dependency injection systems that deal more elegantly with property setters than constructor arguments. Alongside getHandlers(), pushHandler(), popHandler(), it is
a logical addition to the API. 

More controversially. It also attempts to address some possible errors in the format of the data passed:

 - If a map is passed, the keys are removed, as these aren’t expected
   by Monolog
 - If falsey values are included, these are stripped

It relies on Logger::pushHandler() internally, so that if any special behaviour is added in the future, this only needs to be added in one place.

Does anyone else benefit from this facility?

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -176,6 +176,25 @@ class Logger implements LoggerInterface
     }
 
     /**
+     * Set handlers, removing all existing ones.
+     * Falsey values will be ignored, and if a map is passed, keys will be ignored.
+     *
+     * @param array $handlers All elements must be of type HandlerInterface
+     * @return $this
+     */
+    public function setHandlers(array $handlers)
+    {
+        $this->handlers = array();
+        foreach ($handlers as $handler) {
+            if ($handler) {
+                $this->pushHandler($handler);
+            }
+        }
+
+        return $this;
+    }
+
+    /**
      * @return HandlerInterface[]
      */
     public function getHandlers()
--- a/tests/Monolog/LoggerTest.php
+++ b/tests/Monolog/LoggerTest.php
@@ -140,6 +140,30 @@ class LoggerTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @covers Monolog\Logger::setHandlers
+     */
+    public function testSetHandlers()
+    {
+        $logger = new Logger(__METHOD__);
+        $handler1 = new TestHandler;
+        $handler2 = new TestHandler;
+
+        $logger->pushHandler($handler1);
+        $logger->setHandlers(array($handler2));
+
+        // handler1 has been removed
+        $this->assertEquals(array($handler2), $logger->getHandlers());
+
+        $logger->setHandlers(array(
+            "AMapKey" => $handler1,
+            "Falsey" => null,
+        ));
+
+        // Keys have been scrubbed
+        $this->assertEquals(array($handler1), $logger->getHandlers());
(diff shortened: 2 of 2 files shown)


tractorcow:
Nice work @sminnee :)

src/Monolog/Logger.php
```diff
@@ -176,6 +176,25 @@ public function popHandler()
...
+     * Falsey values will be ignored, and if a map is passed, keys will be ignored.
+     *
+     * @param array $handlers All elements must be of type HandlerInterface
+     * @return $this
+     */
+    public function setHandlers(array $handlers)
+    {
+        $this->handlers = array();
+        foreach($handlers as $handler) {
+            if($handler) {
```
camspiers: <nit> `if (`

src/Monolog/Logger.php
```diff
@@ -176,6 +176,25 @@ public function popHandler()
...
+     * Set handlers, removing all existing ones.
+     * Falsey values will be ignored, and if a map is passed, keys will be ignored.
+     *
+     * @param array $handlers All elements must be of type HandlerInterface
+     * @return $this
+     */
+    public function setHandlers(array $handlers)
+    {
+        $this->handlers = array();
+        foreach($handlers as $handler) {
```
camspiers: `foreach (`

sminnee: Thanks, @camspiers. I'll amend the squash 2 fixes into the same commit.

Do you have a linter that picks up on this? I'm using phpcs and it didn't warn me of these.

camspiers: Like much of PHP these days, I think Monolog uses something close to PSR2. http://www.php-fig.org/psr/psr-2/

If you set up phpcs with PSR2 it should give you the warnings.

src/Monolog/Logger.php
kamilsk: and `foreach (`

src/Monolog/Logger.php
```diff
@@ -176,6 +176,25 @@ public function popHandler()
     }
 
     /**
+     * Set handlers, removing all existing ones.
+     * Falsey values will be ignored, and if a map is passed, keys will be ignored.
```
stof: why ignoring falsey values ? It does not make snese IMO, especially given what you document below saying that all elements must be of type HandlerInterface. If you respect this, you cannot pass a falsey value

src/Monolog/Logger.php
```diff
@@ -176,6 +176,25 @@ public function popHandler()
...
+     *
+     * @param array $handlers All elements must be of type HandlerInterface
+     * @return $this
+     */
+    public function setHandlers(array $handlers)
+    {
+        $this->handlers = array();
+        foreach ($handlers as $handler) {
+            if ($handler) {
+                $this->pushHandler($handler);
```
Seldaek: I think the foreach should do an array_reverse on $handlers because otherwise you are essentially flipping the order if I call `setHandlers([a, b])` then `b` will be called first then `a`, which is not really expected IMO.

Seldaek:
I did the tweaks myself so that I can move on and release :) see https://github.com/Seldaek/monolog/commit/419ca7ac2562fe43941692ec442ce6d0398dfe10

sminnee:
Thanks, Jordi, sorry about dropping this one! ;-)

PR #618, opened 2015-07-29
Update Util.php

MangoMM:
Fixes bug with error handler. Once `curl_close($ch)` is called, then `curl_errno($ch)` and `curl_error($ch)` makes no sense and will cause an error in the exception handler.

### DIFF
--- a/src/Monolog/Handler/Curl/Util.php
+++ b/src/Monolog/Handler/Curl/Util.php
@@ -33,12 +33,18 @@ class Util
     {
         while ($retries--) {
             if (curl_exec($ch) === false) {
-                if (false === in_array(curl_errno($ch), self::$retriableErrorCodes, true) || !$retries) {
+                
+                $curlErrno = curl_errno($ch);
+                
+                if (false === in_array($curlErrno, self::$retriableErrorCodes, true) || !$retries) {
+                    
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/Curl/Util.php
```diff
@@ -33,12 +33,18 @@ public static function execute($ch, $retries = 5, $closeAfterDone = true)
...
-                if (false === in_array(curl_errno($ch), self::$retriableErrorCodes, true) || !$retries) {
+                
+                $curlErrno = curl_errno($ch);
+                
+                if (false === in_array($curlErrno, self::$retriableErrorCodes, true) || !$retries) {
+                    
+                    $curlError = curl_error($ch);
+                    
                     if ($closeAfterDone) {
                         curl_close($ch);
```
staabm: wouldn't it make more sense to make sure that after you close `$ch` no more `curl_error` / `curl_errno`  is called?

MangoMM:
@staabm - That's what the pull request does. It get's the values from `curl_errno` and `curl_error` and stores in variables for use after `curl_close` is run.

staabm:
you are right. I mixed things up.

Seldaek:
Thanks

PR #621, opened 2015-08-04
Ensure RollbarHandler can handle fatal errors

brendo:
I ran into an interesting issue while using the _RollbarHandler_ recently and this PR resolves the problem. When a fatal error was being thrown in our application, we were not getting any logs pushed to Rollbar while running in batched mode (which is the default).

This was because the `close()` method was never called to flush the log queue as _AbstractHandler::__destruct()_ is not called during fatal errors. Research seems to suggest that this is expected, as fatal errors do not trigger's PHP's usual shutdown sequence. _RollbarHandler_ currently relies on the destructor to send queued items at the end of a request when it is run in batched mode (where batched mode prevents the Handler for hitting the Rollbar API every time a record appears). 

This PR will inspect the record as it's being written and if the code (`$record['context']['code']`) matches one of PHP's fatal error constants, it will do a force flush. After implementing this, we were able to get all logs for a request, even if it did end in tears :)

### DIFF
--- a/src/Monolog/Handler/RollbarHandler.php
+++ b/src/Monolog/Handler/RollbarHandler.php
@@ -23,6 +23,14 @@ use Monolog\Logger;
 class RollbarHandler extends AbstractProcessingHandler
 {
     /**
+     * Array of fatal types that need the queue to be flushed immediately
+     * as the destructor will not be called.
+     *
+     * @var array
+     */
+    protected $fatalTypes = [E_ERROR, E_PARSE, E_CORE_ERROR, E_COMPILE_ERROR, E_USER_ERROR];
+
+    /**
      * Rollbar notifier
      *
      * @var RollbarNotifier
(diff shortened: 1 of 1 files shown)


Seldaek:
I am not sure if this will play very well with PHP7 as it changes fatals into exceptions so they will likely be logged differently.

Anyway I pushed an alternative fix (2525053e8ef01d7fe703e58aba5300e7881f9fc6) which I believe would do the job as well, at least if you are using monolog's ErrorHandler class to log fatals. If using another implementation that would not do the trick obviously, but your fix isn't very generic it only applies to RollbarHandler and that's not great.

brendo:
> I am not sure if this will play very well with PHP7 as it changes fatals into exceptions so they will likely be logged differently.

Great point, I didn't consider this.

> Anyway I pushed an alternative fix (2525053) which I believe would do the job as well, at least if you are using monolog's ErrorHandler class to log fatals. If using another implementation that would not do the trick obviously, but your fix isn't very generic it only applies to RollbarHandler and that's not great.

That's a much more elegant fix! I've pulled 1.16 and can confirm it works a treat. Thanks!

Seldaek:
Great, thanks for checking!

PR #623, opened 2015-08-08
Added IFTTT Handler

nehalvpatel:
I've added a fairly simple handler that allows you to trigger Maker actions on IFTTT.

The great thing about IFTTT is, you can essentially do anything you want with the logs, for free:
- You want to get a text alert when there's a database error? Yep.
- You want to create a GitHub issue when there's a critical error? Yep.
- You want your house to go lockdown mode and start blinking red lights when there's an emergency logged? Sure. 

On a more serious note though, how should the tests be handled for this? I looked through the other handlers that use cURL, such as Mandrill and Loggly, and they don't seem to have tests, so I didn't add one to this. **However**, if you do want one, should I just mock a cURL object? Or would you prefer I do something else?

### DIFF
--- a/src/Monolog/Handler/IFTTTHandler.php
+++ b/src/Monolog/Handler/IFTTTHandler.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Logger;
+
+/**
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, it's fine like that I think. I just did some adjustments to the coding style in https://github.com/Seldaek/monolog/commit/cd3f9f1e5071e393a5f592cd54838c6f53212c93 in case you're interested.

nehalvpatel:
Thanks for accepting the pull! :)

About the coding style though, how come you use spaces instead of tabs? Is that just a preference or is there some benefit to using spaces?

Seldaek:
Not really a big benefit except for being "standard". Most projects use 
that as well as the PSR-1/PSR-2 (http://www.php-fig.org/psr/psr-2/) 
guidelines these days which makes it easier for people to interact 
across projects if they don't have to change their editor settings all 
the time.

PR #608, opened 2015-07-15
Ignore "call_user_func" from debug backtrace

xploSEoF:
Helps make this compliant where there is no class defined

PR #625, opened 2015-08-11
FIX IntrospectionProcessor: E_NOTICE

frederikbosch:
Recent merge of #608 misses check whether the trace exists at all at the specific index, leading to undefined offset.

```
E_NOTICE: Undefined offset: 3
```

And because the while statement became unreadable (and too long), I moved it into a separate method.

### DIFF
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -91,7 +91,7 @@ class IntrospectionProcessor
         return $record;
     }
 
-    private function isTraceClassOrSkippedFunction ($trace, $index)
+    private function isTraceClassOrSkippedFunction (array $trace, $index)
     {
         if (isset($trace[$index]) === false) {
             return false;
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -61,7 +61,7 @@ class IntrospectionProcessor
 
         $i = 0;
 
-        while ($this->isTraceClassOrNotSkippedFunction($trace, $i)) {
+        while ($this->isTraceClassOrSkippedFunction($trace, $i)) {
             if (isset($trace[$i]['class'])) {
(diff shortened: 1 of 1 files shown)


src/Monolog/Processor/IntrospectionProcessor.php
```diff
@@ -90,4 +90,13 @@ public function __invoke(array $record)
 
         return $record;
     }
+
+    private function isTraceClassOrSkippedFunction ($trace, $index)
```
stof: you should typehint the first element as array

src/Monolog/Processor/IntrospectionProcessor.php
```diff
@@ -90,4 +90,13 @@ public function __invoke(array $record)
...
+    {
+        if (isset($trace[$index]) === false) {
```
stof: I would use ``!isset(...)`` here

src/Monolog/Processor/IntrospectionProcessor.php
frederikbosch: Damn right, updated.

frederikbosch:
Just a random notice: there is a focus on maintaining PSR-2, which is great. However, the error in #608 (which is quite obvious), went through without problems. I mean a style police is very fine. There is nothing against it, but it could lead to the focus on the wrong things. Again, just a random notice.

Seldaek:
I am not sure what you mean about the style police and focus on wrong things.. ? But aside from that, thanks for the fix :)

frederikbosch:
@Seldaek Never mind that comment. Thanks for merging.

Seldaek:
Well I am happy to discuss this but I just did not understand the point you were making.

snapshotpl:
Hi there! Can you release fixed stable version with it? My application cannot works property with this issue

frederikbosch:
@Seldaek I would also be happy with a new tag.

Seldaek:
1.17 is out

frederikbosch:
Thanks @Seldaek!

PR #630, opened 2015-08-15
RavenHandler handles custom checksums

prgTW:
It gives the ability to override how events are grouped together in Sentry


Seldaek:
Thanks

prgTW:
You're welcome :)

PR #631, opened 2015-08-16
RavenHandler supports extra parameters

prgTW:
refactored RavenHandler a bit to support many additional parameters, not only `checksum` but also (added in 0.11) `release` parameter
ref: https://github.com/getsentry/raven-php/blob/master/CHANGES

`release` allows to track the version of an application that the logs came from

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
     "require-dev": {
         "phpunit/phpunit": "~4.5",
         "graylog2/gelf-php": "~1.0",
-        "raven/raven": "~0.8",
+        "raven/raven": "~0.11",
         "ruflin/elastica": ">=0.90 <3.0",
         "doctrine/couchdb": "~1.0@dev",
         "aws/aws-sdk-php": "^2.4.9",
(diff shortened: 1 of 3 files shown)


Seldaek:
Thanks

greg0ire:
@prgTW wouldn't it make more sense to specify the release in the handler constructor than on each log method call ?

prgTW:
maybe so, although I was using monolog with Symfony and there all handler options are defined via bundle configuration in compile time. While I'm get my version number dynamically from git commit ref I would have to alter this configuration somehow dynamically ... but I didn't want to play with this and solve the problem (compat. with Pareto principle) and deliver 80% of value in 20% of time.

Feel free to refactor :)

greg0ire:
I'm using Symfony too, and I think it would be possible to have a class reponsible for getting the version number (from a VERSION.txt file created when deploying, for instance). This class could be used in the monolog bundle extension class to set a container parameter, which could in turn be fed to the constructor of the handler. But good point about the compatibility with other software, which may not have big DI capabilities. I'll refactor if I find the time and motivation :)

prgTW:
@greg0ire You can always introduce a setter method to be backward compatible instead of altering constructor. Using it would require almost no changes to 3rd parties

greg0ire:
Indeed, a setter makes more sense, also because the release number is an optional dependency of the handler.

greg0ire:
@prgTW see #727

PR #635, opened 2015-08-22
Allow extra_data and payload to be sent, too

mbrevda:
Any thoughts @paulstatezny? Did you specifically exclude this data?

### DIFF
--- a/src/Monolog/Handler/RollbarHandler.php
+++ b/src/Monolog/Handler/RollbarHandler.php
@@ -55,14 +55,15 @@ class RollbarHandler extends AbstractProcessingHandler
     {
         if (isset($record['context']['exception']) && $record['context']['exception'] instanceof Exception) {
             $context = $record['context'];
+            $exception = $context['exception']
             unset($context['exception']);
 
             $payload = isset($context['payload']) ? $context['payload'] : [];
-            if ($context['payload']) {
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/RollbarHandler.php
```diff
@@ -54,7 +54,15 @@ public function __construct(RollbarNotifier $rollbarNotifier, $level = Logger::E
     protected function write(array $record)
     {
         if (isset($record['context']['exception']) && $record['context']['exception'] instanceof Exception) {
-            $this->rollbarNotifier->report_exception($record['context']['exception']);
+            $context = $record['context'];
+            unset($context['exception']);
+
+            $payload = isset($context['payload']) ? $context['payload'] : [];
+            if ($context['payload']) {
```
paulstatezny: Should this be `if (isset($context['payload']))`?

When the `isset` check above fails, `if ($context['payload'])` will cause an `undefined index` error, correct?

paulstatezny:
@mbrevda Sorry for the delay. I did not intentionally exclude this data. Just wasn't this thorough. Thanks for building on it. :smile:

mbrevda:
@Seldaek - can we get this merged?

Seldaek:
Looks good so let's merge, I'm in holidays so not watching github too closely

PR #636, opened 2015-08-24
Add getUid() to UidProcessor

jaredm4:
Would be nice to have an easy way to get Uid from UidProcessor. I have a scenario where I want to pass a parent's Uid to children processes for log grouping.

### DIFF
--- a/src/Monolog/Processor/UidProcessor.php
+++ b/src/Monolog/Processor/UidProcessor.php
@@ -35,4 +35,12 @@ class UidProcessor
 
         return $record;
     }
+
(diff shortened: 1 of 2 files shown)


jaredm4:
Another option if deemed better is to change $uid to protected. Then I can extend the UidProcessor class and add the getter there.

stof:
@jaredm4 making it protected is actually not better: it means we have to support the extension forever (well, at least forever in 1.x) and it also makes it possible for a child class to mutate it, which is opening it more than adding a getter

PR #641, opened 2015-09-07
Added the capability of handling a message multiple times

mjacobus:
What if I want error messages to be logged, but critical messages to be logged, mailed and sent to slack channel?

If I understood correctly this is currently not possible. This is why I created this PR, to add such capability.

```php
$log = new Logger('name');
$log->pushHandler(new StreamHandler('path/to/your.log', Logger::ERROR));
$log->pushHandler(new StreamHandler('path/to/your.log', Logger::CRITICAL));
$log->pushHandler(new MyEmailHandler(Logger::CRITICAL));
$log->pushHandler(new SlackHandler($connectionString, Logger::CRITICAL));
```

Note that 2 tests are now failing. I did not change them because I was not sure if that was testing the expected behavior or if it was just very strict expectations, since the [commit](https://github.com/Seldaek/monolog/commit/64009a58) that introduced those tests seemed just to be increasing the test coverage.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -285,9 +285,11 @@ class Logger implements LoggerInterface
         foreach ($this->processors as $processor) {
             $record = call_user_func($processor, $record);
         }
-        while (isset($this->handlers[$handlerKey]) &&
-            false === $this->handlers[$handlerKey]->handle($record)) {
-            $handlerKey++;
+
+        foreach ($this->handlers as $handler) {
+            if ($handler->isHandling(array('level' => $level))) {
+                $handler->handle($record);
+            }
         }
 
         return true;
--- a/tests/Monolog/LoggerTest.php
+++ b/tests/Monolog/LoggerTest.php
@@ -393,6 +393,46 @@ class LoggerTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * @test
+     */
+    public function multipleHandlersCanHandleTheSameRecord()
+    {
+        $logger = new Logger(__METHOD__);
+
+        $handler1 = $this->getMock('Monolog\Handler\HandlerInterface');
+        $handler1->expects($this->any())
+            ->method('isHandling')
+            ->will($this->returnValue(false));
+
+        $handler1->expects($this->never())
+            ->method('handle');
+
+        $handler2 = $this->getMock('Monolog\Handler\HandlerInterface');
+        $handler2->expects($this->any())
+            ->method('isHandling')
+            ->will($this->returnValue(true));
+
+        $handler2->expects($this->once())
+            ->method('handle')
+            ->will($this->returnValue(true))
+            ;
+
+        $handler3 = $this->getMock('Monolog\Handler\HandlerInterface');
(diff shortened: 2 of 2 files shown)


src/Monolog/Logger.php
```diff
@@ -285,9 +285,11 @@ public function addRecord($level, $message, array $context = array())
         foreach ($this->processors as $processor) {
             $record = call_user_func($processor, $record);
         }
-        while (isset($this->handlers[$handlerKey]) &&
-            false === $this->handlers[$handlerKey]->handle($record)) {
-            $handlerKey++;
+
```
staabm: this change kills the optimization which line https://github.com/Seldaek/monolog/pull/641/files#diff-dce565f403e044caa5e6a0d988339430R258 tries to achieve

src/Monolog/Logger.php
```diff
@@ -285,9 +285,11 @@ public function addRecord($level, $message, array $context = array())
         foreach ($this->processors as $processor) {
             $record = call_user_func($processor, $record);
         }
-        while (isset($this->handlers[$handlerKey]) &&
```
staabm: the old code let all loggers - starting at the first which supports the current error level - handle the logmessage as long as they `return false`..?

this in turn is returns `false` as long as your handler defines `bubble=true`.
not sure what is missing to get it running for your case?

src/Monolog/Logger.php
mjacobus: Do you mean line 258? It still returns earlier if no handler handles that message. Any suggestion on how to merge both functionalities?

src/Monolog/Logger.php
mjacobus: Hum, I was confused by this bubble thing. Maybe that does it.

src/Monolog/Logger.php
mjacobus: IMHO, 

1 - That bubble param is not intention revealing. That should be named 'breakeChain' or something

2- If this bubble thing is what defines this behavior, there should be a method for this property in the [interface](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/HandlerInterface.php#L21)

src/Monolog/Logger.php
mjacobus: IMHO, 

1 - That bubble param is not intention revealing. That should be named 'breakeChain' or something

2- If this bubble thing is what defines this behavior, there should be a method for this property in the [interface](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/HandlerInterface.php#L21)

staabm: the expected behaviour is documented in the interface:
https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/HandlerInterface.php#L49-L50

mjacobus: Yes, I had read that. What is "bubbling"? Maybe it is just because I am no English native speaker, but I think that is far from clear =)

staabm: google can tell:

http://stackoverflow.com/questions/4616694/what-is-event-bubbling-and-capturing

PR #646, opened 2015-09-16
Prevent HipChat API failure from throwing an exception

alextartan:
When a connection to HipChat API cannot be established, the 'createSocketResource' method from the 'SocketHandler' class throws an exception. 

To prevent this, I've wrapped the "write" method from the 'HipChatHandler' class in a try-catch block. While the call specifically throws an 'UnexpectedValueException', the catch is using '\Exception' since no action is actually needed and the exception type doesn't really matter is this particular case.

### DIFF
--- a/src/Monolog/Handler/HipChatHandler.php
+++ b/src/Monolog/Handler/HipChatHandler.php
@@ -205,8 +205,12 @@ class HipChatHandler extends SocketHandler
      */
     protected function write(array $record)
     {
-        parent::write($record);
-        $this->closeSocket();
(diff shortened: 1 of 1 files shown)


Seldaek:
You can use the WhatFailureGroupHandler for this, as handlers should by default throw if they can not write otherwise misconfigurations can go unnoticed, and log data loss can be quite critical in some contexts. See https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/WhatFailureGroupHandler.php

PR #647, opened 2015-09-17
Use `text` and `title` fields for SlackHandler

rap2hpoutre:
According to slack documentation (https://api.slack.com/docs/attachments),  an attachment may have a title and a text. The log message and title should use these fields. One of the benefits of using these two properties instead of adding a field is that it's shortened (with a "show more" button) and expandable by slack. It adds a bit of hierarchy in the attachement. Message is the content. Extra and context are additional fields.

### DIFF
--- a/src/Monolog/Handler/SlackHandler.php
+++ b/src/Monolog/Handler/SlackHandler.php
@@ -145,29 +145,20 @@ class SlackHandler extends SocketHandler
         if ($this->useAttachment) {
             $attachment = array(
                 'fallback' => $record['message'],
-                'color'    => $this->getAttachmentColor($record['level'])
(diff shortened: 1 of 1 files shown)


rap2hpoutre:
Checks have failed but they failed too on master so I think it's not this PR, right?

Seldaek:
Would you have before/after screenshots by any chance?

rap2hpoutre:
Here are the screenshots.
Before (with a stack trace):
capture d ecran 2015-09-21 a 09 46 50

After (same error, you can see the "show more" button added by slack):
capture d ecran 2015-09-21 a 09 48 07

Seldaek:
OK thanks, that looks like an improvement yes :)

PR #654, opened 2015-09-26
Reduce memory consumption of the debug_backtrace call

dracony:
Since the code does not need neither the 'object' not 'args' keys from backtrace, we can save some memory by not requesting them.

I have a better idea planned too: first attempt getting the trace with limited back trace (limit to only about 5 results), and get the full one only if those were not enough

### DIFF
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -52,7 +52,7 @@ class IntrospectionProcessor
             return $record;
         }
 
-        $trace = debug_backtrace();
+        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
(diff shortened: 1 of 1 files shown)


staabm:
How much does this save on a real world call?

dracony:
Depends on your real world of course. Currently Im trying to solve the
problem of monologs debug_backtrace eating up to 6 mb ram in a long running
process ( with deep recursion). This parmeter cut it by half. Furthemore
there is no downside to this.


A bigger improvement is to limit the backtrace to 5 elements instead of
getting the full one, and if a proper caller cannot be found in that limit
only then get the full one. Realistically the entire debug_backtrace() in a
big app results in a huuuuge array ( especially withing recursive function
calls)

Seldaek:
Thanks

PR #656, opened 2015-09-27
Add linting to Travis tests

bd808:
Protect against syntax compatibility issues such as the non-5.3
compatible array syntax corrected in 5785a9a by running a linter during
Travis CI testing for pull requests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,4 @@ before_script:
   - if [ "$deps" == "low" ]; then composer update --prefer-source --prefer-lowest --prefer-stable; fi
   - if [ "$deps" != "low" ]; then composer install --prefer-source; fi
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

.travis.yml
```diff
@@ -21,4 +21,4 @@ before_script:
   - if [ "$deps" == "low" ]; then composer update --prefer-source --prefer-lowest --prefer-stable; fi
   - if [ "$deps" != "low" ]; then composer install --prefer-source; fi
 
-script: ./vendor/bin/phpunit
+script: composer test
```
stof: I think it would be more readable to have multiple commands here on Travis, so that they all run all the time, instead of running the next one only when the first one works

PR #658, opened 2015-09-27
Optionally reduce timestamp resolution for performance

### LINKED ISSUE #657: Microsecond timestamps cost milliseconds of runtime with heavy log usage
The use of `DateTime::createFromFormat` to create timestamps with microsecond resolution can incur milliseconds of runtime overhead when handling a request which generates a significant number of logging events. This cost is paid even if the events are eventually discarded as the `$record['datetime']` member for each event is populated even if no handlers are interested in recording the event durably.

This simplistic benchmark shows the relative costs of collecting microsecond level timing data:

```php
<?php
$limit = 1000;
$tz = new DateTimeZone('UTC');

$start = microtime(true);
for ($i = 0; $i < $limit; $i++) {
        $dt = DateTime::createFromFormat('U.u', sprintf('%.6F', microtime(true)), $tz)->
                setTimezone($tz);
}
echo 'createFromFormat: ', microtime(true) - $start, "\n";

$start = microtime(true);
for ($i = 0; $i < $limit; $i++) {
        $dt = new DateTime(null, $tz);
}
echo 'DateTime:         ', microtime(true) - $start, "\n";
```

```sh
$ php benchmark.php
createFromFormat: 0.010702133178711
DateTime:         0.0037209987640381
```

Use of `DateTime::createFromFormat` was introduced in 61a4945 replacing a simple usage of `new DateTime`. In a application that generates a few dozens or even hundreds of log events per request the runtime cost difference of `DateTime::createFromFormat` is probably not noticeable. The costs add up however on a heavily loaded system handling hundreds of requests per second with hundreds or even thousands of log events generated per request. Under any conditions the benefit of having microsecond resolution for timestamps verses millisecond level resolution is debatable. When having this level of granularity requires milliseconds of wall clock time to compute, the benefit is even more tenuous.

The easy fix for this issue would be to simply replace `\DateTime::createFromFormat('U.u', sprintf('%.6F', microtime(true)), static::$timezone)->setTimezone(static::$timezone)` with `'datetime' => new \DateTime(null, static::$timezone)` in `Monolog\Logger`. A more complicated approach would be to introduce a feature flag that allows control of millisecond/microsecond timestamp resolution. If I was the benevolent dictator for this project I would chose the simple solution under the basic assumption that logging should seek to be as low impact as possible, especially for `debug` events that are likely to be discarded anyway.

bd808:
Introduce a new `useMicrosecondTimestamps()` static method to Logger
that allows runtime selection of whether or not to create microsecond
resolution timestamps for log records.

Generating microsecond resolution timestamps by calling
`microtime(true)`, formatting the result via `sprintf()` and then
parsing the resulting string via `DateTime::createFromFormat` can incur
a measurable runtime overhead vs simple usage of `new DateTime` to
capture a second resolution timestamp in systems which generate a large
number of log events.

The default behavior of generating high precision timestamps remains the
same, but may be changed in a future release. Users requiring high
precision timestamps are encouraged to explicitly call
`Monolog\Logger::useMicrosecondTimestamps(true)` in their setup code.

Closes #657

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -129,6 +129,11 @@ class Logger implements LoggerInterface
     protected $processors;
 
     /**
+     * @var bool
+     */
+    protected $microsecondTimestamps = true;
+
+    /**
      * @param string             $name       The logging channel
      * @param HandlerInterface[] $handlers   Optional stack of handlers, the first one in the array is called first, etc.
      * @param callable[]         $processors Optional array of processors
@@ -239,6 +244,25 @@ class Logger implements LoggerInterface
         return $this->processors;
     }
 
+
+    /**
+     * Control the use of microsecond resolution timestamps in the 'datetime'
+     * member of new records.
+     *
+     * Generating microsecond resolution timestamps by calling
+     * microtime(true), formatting the result via sprintf() and then parsing
+     * the resulting string via \DateTime::createFromFormat() can incur
+     * a measurable runtime overhead vs simple usage of DateTime to capture
+     * a second resolution timestamp in systems which generate a large number
+     * of log events.
+     *
+     * @param bool $micro True to use microtime() to create timestamps
+     */
+    public function useMicrosecondTimestamps($micro)
+    {
+        $this->microsecondTimestamps = (bool)$micro;
+    }
+
     /**
      * Adds a log record.
      *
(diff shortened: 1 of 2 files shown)


staabm:
the reason the build failes is a issue with raven, not related to this PR which is fixed with https://github.com/Seldaek/monolog/pull/660

atdt:
Ping! :)

bd808:
Updated patch and description based on feedback from @Seldaek in https://github.com/Seldaek/monolog/issues/657#issuecomment-148472673.

Seldaek:
@bd808 Sorry but would you mind making this flag and setter non-static? I think it was a mistake to have the timezone statically to be honest, and it'll be changed in 2.0, so let's not generate more BC breaks :) Aside from that it looks good so I can merge as soon as you let me know it's done.

bd808:
@Seldaek changed from a static to a member variable.

Seldaek:
Thanks!

atdt:
Thank you very much, @bd808 and @Seldaek.

PR #668, opened 2015-10-19
Avoid infinite loops when no data is written on a socket for a time 

silvadanilo:
greater than writingTimeout settings.

When a remote socket is closed without the FIN packet, it may be that the local socket is not closed but the data is not sent anymore so php entering in an infinite loop.

To reproduce the issue you can:
- start netcat in listening mode on 1234 port:
    `nc -l 1234`

- run the script locateted here: https://gist.github.com/silvadanilo/5043fc2d16b5346b1db1

- after a few logs entry, interrupt netcat with Ctrl+c

### DIFF
--- a/src/Monolog/Handler/SocketHandler.php
+++ b/src/Monolog/Handler/SocketHandler.php
@@ -25,6 +25,8 @@ class SocketHandler extends AbstractProcessingHandler
     private $connectionTimeout;
     private $resource;
     private $timeout = 0;
+    private $writingTimeout = 0;
+    private $lastSentBytes = null;
     private $persistent = false;
     private $errno;
     private $errstr;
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/SocketHandler.php
```diff
@@ -276,9 +301,35 @@ private function writeToSocket($data)
...
+    private function writingIsTimedOut($sent)
+    {
+        $writingTimeout = (int) floor($this->writingTimeout);
+        if (0 === $writingTimeout) {
+            return false;
+        }
+
+        if ($sent !== $this->lastSentBytes) {
+            $this->lastWritingAt = time();
+            $this->lastSentBytes = $sent;
```
Seldaek: Shouldn't you `return false` here? If it seems like it wrote successfully I don't think we should do the time check?

Seldaek:
Thanks, however I am wondering if it should really default to zero or if we should default to something like 10 as a safety measure? Is there a scenario where this could cause problems though? Like if you don't write any log for 10seconds in a long running process and then write something, it should be fine still right?

silvadanilo:
I agree with you that the default value should not be zero, I put zero for backward compatibility, but if all goes well with you I will gladly put something like 10.

No problem when you don't write any log for 10 seconds in a long running process because the `timeout check` is while data is written on socket, not outside writing.

src/Monolog/Handler/SocketHandler.php
silvadanilo: yes, you're right!

Seldaek:
Great thanks!

silvadanilo:
thanks to you!

PR #669, opened 2015-10-19
[TASK] CGL fix

afoeder:
This replaces a double-quote occurrence with
a single-quote one to comply with the CGL
as well as fixing an escaping mistake.


Seldaek:
Thanks

PR #671, opened 2015-10-22
added param in InstrospectionProcessor __construct to allow stacktrac…

mtopolski:
…e offset

Wanted to be able to have my handlers wrapped without compromising the value of the stack trace, also needed to be able to adjust stack trace on the fly depending on from where I am logging. Example usage--

```
function monolog_error_log($message = '', $extra_stack = 0) {
    $error_logger = new Logger('FOO_BAR');
    $error_stream = new StreamHandler('var/log/nginx/error.log');
    $formatter = new LineFormatter();
    $introspectionProcessor = new IntrospectionProcessor(Logger::DEBUG, array(), $extra_stack);
    $error_stream->setFormatter($formatter);
    $error_logger->pushProcessor($introspectionProcessor);
    $error_logger->pushHandler($error_stream);
    $error_logger->addError($message);
}
```

### DIFF
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -30,18 +30,18 @@ class IntrospectionProcessor
 
     private $skipClassesPartials;
 
-    private $extra_stack;
+    private $extraStack;
 
     private $skipFunctions = array(
         'call_user_func',
         'call_user_func_array',
     );
 
-    public function __construct($level = Logger::DEBUG, array $skipClassesPartials = array(), $extra_stack = 0)
+    public function __construct($level = Logger::DEBUG, array $skipClassesPartials = array(), $extraStack = 0)
     {
         $this->level = Logger::toMonologLevel($level);
         $this->skipClassesPartials = array_merge(array('Monolog\\'), $skipClassesPartials);
-        $this->extra_stack = $extra_stack;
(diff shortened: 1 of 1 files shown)


src/Monolog/Processor/IntrospectionProcessor.php
```diff
@@ -30,15 +30,18 @@ class IntrospectionProcessor
 
     private $skipClassesPartials;
 
+    private $extra_stack;
```
stof: the name should be camelcased

stof:
Couldn't you handle your use case with the class/function ignore list instead ?

mtopolski:
I definitely could; I worry about changes to the code causing more trouble that way though. If the function name changes my trace is off, in some cases I'm actually skipping 3 or 4 functions, and in some cases I could see myself logging an error where I know I want to go one level deeper anyway. It feels more flexible to me this way?

mpatnode-si:
The ignore list seems a little fragile.  What happens when method names change?   Also, += should be faster than multiple string compares.

Seldaek:
Not really against this but I feel like $extraStack is not exactly self-documenting. From the name alone I'd expect it adds more stack frames than one. Could you rename it? Maybe $skipStackFrames? Just an idea.

mtopolski:
That sounds like a good idea to me; now it matches the $skipClassesPartials too.

Seldaek:
Thanks!

Issue #673, opened 2015-10-25
Handle fatal error on exception in __toString methods


xapon:
When a NormalizerFormatter tries to normalize an object that implements '__toString' method, and this method throws an exception (which is exceptionally bad, but still), it results in a messy fatal error, which is hard to debug. This is especially painful if happens in the process of logging another error or exception, so we end up not knowing what the original error was.

The solution, i think, is to simply call '$data->__toString()' instead of '(string) $data', which will correctly throw an exception instead of triggering fatal error.

PR #674, opened 2015-10-25
#673 Fix fatal error while normalizing with __toString method

xapon:
See issue at https://github.com/Seldaek/monolog/issues/673


Seldaek:
Oh that's interesting, I didn't realize there was an actual behavioral difference between (string) and an explicit call to __toString.

Thanks for the fix!

PR #676, opened 2015-11-03
mongodb/mongodb support

kornrunner:
This should add support for new https://github.com/mongodb/mongo-php-library that deprecates old one (`ext-mongo`).

New library + extension is the only one that will get support for PHP7. It recently got released as beta / release candidate - and I believe this patch should help / ease everyone who's looking forward to PHP7 upgrade.

New client class has the required [`selectCollection`] (http://mongodb.github.io/mongo-php-library/api/class-MongoDB.Client.html#_selectCollection) method that is invoked, however I had to add a case for `MongoDB\Collection` which no longer has `save` method, but [`insertOne`] (http://mongodb.github.io/mongo-php-library/api/class-MongoDB.Collection.html#_insertOne) / [`insertMany`] (http://mongodb.github.io/mongo-php-library/api/class-MongoDB.Collection.html#_insertMany) instead. 

I hope I hadn't missed anything, but I'll happy to rework it if required. Thanks!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,6 +38,7 @@
         "videlalvaro/php-amqplib": "Allow sending log messages to an AMQP server using php-amqplib",
         "ext-amqp": "Allow sending log messages to an AMQP server (1.0+ required)",
         "ext-mongo": "Allow sending log messages to a MongoDB server",
+        "mongodb/mongodb": "Allow sending log messages to a MongoDB server via PHP Driver",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
         "rollbar/rollbar": "Allow sending log messages to Rollbar",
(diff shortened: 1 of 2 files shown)


Seldaek:
Seems fine to me thanks, ping @jmikola just in case you spot something weird here.

jmikola:
@kornrunner: In hindsight, I think it would be preferable to simply depend on `ext-mongodb`. Since the handler class is constructed with the `$database` and `$collection` names as strings, we can receive a `MongoDB\Driver\Manager` instance and log messages with:

```
$bulk = new MongoDB\Driver\BulkWrite;
$bulk->insert($document);
$manager->executeBulkWrite($database . '.' . $collection, $bulk);
```

`executeBulkWrite()` can take an optional WriteConcern as its third argument, but it doesn't look like Monoglog is using that anyway.

PR #679, opened 2015-11-09
Update NativeMailerHandler.php

glensc:
don't calculate parameters in a loop

### DIFF
--- a/src/Monolog/Handler/NativeMailerHandler.php
+++ b/src/Monolog/Handler/NativeMailerHandler.php
@@ -122,8 +122,9 @@ class NativeMailerHandler extends MailHandler
         if ($this->getContentType() == 'text/html' && false === strpos($headers, 'MIME-Version:')) {
             $headers .= 'MIME-Version: 1.0' . "\r\n";
         }
+        $parameters = implode(' ', $this->parameters);
         foreach ($this->to as $to) {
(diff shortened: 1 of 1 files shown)


glensc:
note, i initially committed with a typo (64fc8534849ab728b6d16960d03a3cfa20928527), seems testcase did not catch this!

```
All checks have passed
1 successful check
```

Seldaek:
Thanks, it's a minor improvement really as it's unlikely that people send to so many addresses that this would really matter, but why not I guess.

glensc:
it's about code style. this code will be likely source for many other projects. so it's good idea to have excellent code ;)

but what about testing, that the typo was not catched? i think the part is not tested at all.

Seldaek:
Yeah that's well possible, we don't have tests for everything, some 
handlers are pretty hard to test and pretty low return as the code is 
simple and changes rare.

glensc:
as i understand this can be easily tested to avoid calling native `mail()`?

i.e declare `mail()` in `Monolog\Handler` namespace, and the handler will call `\Monolog\Handler\mail()` function not `\mail()`, right?

Seldaek:
Yup I think so as well, feel free to give it a shot if you like, the 
more tests the better I am not gonna argue with that :)

PR #682, opened 2015-11-10
Add HTTP_X_FORWARDED_IP (if set) when user wants to log IP

stefandoorn:
Add HTTP_X_FORWARDED_IP (if set) when user wants to log IP from $_SERVER array, useful behind loadbalancer eg.

In our current setup the loadbalancer IP is in $_SERVER['REMOTE_ADDR'], which is not particularly useful in my case. Checking for the existence of HTTP_X_FORWARDED_IP can be useful in this case, and will only be added if user was requesting for the IP already.

### DIFF
--- a/src/Monolog/Processor/WebProcessor.php
+++ b/src/Monolog/Processor/WebProcessor.php
@@ -108,6 +108,13 @@ class WebProcessor
             $extra['unique_id'] = $this->serverData['UNIQUE_ID'];
         }
 
+        // Add forwarded IP, useful eg when application runs behind loadbalancer
+        if (array_key_exists('ip', $this->extraFields)) {
+            if (array_key_exists('HTTP_X_FORWARDED_FOR', $this->serverData)) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Problem is `x-forwarded-ip` isn't very standard, and it can be sent by clients. There are `Client-IP` and `x-forwarded-for` headers as well, and the somewhat-newly standardized `forwarded` https://tools.ietf.org/html/rfc7239 - and as I said they can all be sent by clients, so you should only accept their content up to the trusted loadbalancer/proxy IP, which is complex logic I really don't want to take care of in there.

Symfony for example has a bunch of code doing this https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/HttpFoundation/Request.php#L792-L838 that is based on a configured list of trusted IPs.

I think the better way really is to handle this at the application level and add your own processor to do this, it's very easy to do and means we don't open pandora's box in monolog :)

stefandoorn:
Hm, that makes sense. Maybe best indeed to solve it at application level for a specific use case. Thanks for the extensive reply :)

Issue #616, opened 2015-07-24
JSON Encode Failed 


austinh:
JSON encoding failed: Malformed UTF-8 characters, possibly incorrectly encoded. Encoding: array (

I upgraded my symfony composer files lately, and got this error. This does not happen in 1.14.0. The latest change in "Fixed JSON encoding errors being silently suppressed (PHP 5.5+ only)" caused this error.

I am using this in conjunction with SNCRedis Bundle to store sessions in redis.

Is there any way sncredisbundle can utf8_encode the log message?
The session is saved fine and everything else works, its purely the logging.
I have also created a bug at SncRedisBundle git because it may be an issue with their code.

Seldaek:
I believe this is usually indicating a  malformed UTF-8 message yes, but maybe I made a mistake, can you link to that SncRedis issue?

austinh:
https://github.com/snc/SncRedisBundle/issues/193

ckmaresca:
I don't know if this is relevant, but there are a bunch of options that you can use with `json_encode` to deal with odd characters that cause encoding to fail.  They are described here: http://php.net/manual/en/json.constants.php - specifically `JSON_HEX_*` and `JSON_UNESCAPED_*`

I would note that this is a pain in PHP as it's not obvious that you need these options for things to work and `json_encode` has a habit of failing silently with no exception to trap....

Seldaek:
I'm gonna assume this gets fixed by https://github.com/Seldaek/monolog/pull/683 as there is no clear repro case I can't verify but it most likely is covered too.

@ckmaresca for the record, JSON_HEX and JSON_UNESCAPED do nothing to prevent these errors unfortunately, they merely tweak the way things are encoded by default, but that only concerns *valid* UTF-8 strings. Non-UTF-8 can not be encoded to json as the format mandates Unicode strings.

PR #683, opened 2015-11-12
Attempt to recover from json encoding errors

### LINKED ISSUE #545: No error on non UTF8 input in NormalizerFormatter
I'm using the `LogstashFormatter`.When I log non UTF8 stuff the `NormalizerFormatter::toJson` returns false which cause finally just an \n in the log and no error message at all. Is that expected? Would it perhaps not be better to throw an error in `toJson()`?

bd808:
Detect and attempt to recover from json_encode errors triggered by
strings containing invalid UTF-8 sequences. Recovery will only be
attempted when encoding strings or arrays. If recovery fails then
a RuntimeException will be thrown.

The recovery process will convert invalid UTF-8 codepoints as though the
input string was encoded using the ISO-8859-15 character encoding. This
conversion may result in incorrect string output if the original
encoding was not ISO-8859-15, but it will be a valid UTF-8 string.

Closes #545 and possibly #616

### DIFF
--- a/src/Monolog/Formatter/NormalizerFormatter.php
+++ b/src/Monolog/Formatter/NormalizerFormatter.php
@@ -138,25 +138,76 @@ class NormalizerFormatter implements FormatterInterface
         return $data;
     }
 
+    /**
+     * Return the JSON representation of a value
+     *
+     * @param mixed $data
+     * @param bool  $ignoreErrors
+     * @return string
+     * @throws \RuntimeException if encoding fails and errors are not ignored
+     */
     protected function toJson($data, $ignoreErrors = false)
     {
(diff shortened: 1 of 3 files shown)


Seldaek:
Thanks, fixed a bunch of CS nitpicks (see above) just in case you wanna have a look but no biggie. Otherwise I  am sure it won't work for everyone but it's still good to make sure most things can be logged even if it leads to a bit of data loss.

bd808:
Thanks @Seldaek and sorry for leaving you those code style issues.

Seldaek:
No worries I just run php-cs-fixer so it's no work but I like to point it
out for the record :)

PR #686, opened 2015-11-17
[HipChatHandler] Always send the "from" name

apfelbox:
Although the docs for API version v2 are very sparse, a quick test confirmed, that the name should always be sent:

Without the patch:
2015-11-17 at 18 52

With the patch (sorry for obfuscation, client name...):
2015-11-17 at 18 53

(same config)

### DIFF
--- a/src/Monolog/Handler/HipChatHandler.php
+++ b/src/Monolog/Handler/HipChatHandler.php
@@ -141,12 +141,17 @@ class HipChatHandler extends SocketHandler
             'message' => $record['formatted'],
             'message_format' => $this->format,
             'color' => $this->getAlertColor($record['level']),
-            'from' => $this->name,
         );
 
         // if we are using the legacy API then we need to send some additional information
         if ($this->version == self::API_V1) {
             $dataArray['room_id'] = $this->room;
+        }
+
+        // append the sender name if it is set
+        // always append it if we use the v1 api (it is required in v1)
+        if ($this->version == self::API_V1 || $this->name !== null) {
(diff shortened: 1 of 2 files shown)


apfelbox:
@esalter do you remember, why the `from` field should only be sent for API v2? Does it produce issues in older clients?

esalter:
Hi @apfelbox , I'm sorry, I don't remember, it's been quite a while since I looked at that.

andrioli:
@apfelbox I don't know the reason the field `form` was removed (@esalter changed this behavior some time ago). But in fact, this field is only required in [API v1](https://www.hipchat.com/docs/api/method/rooms/message). For [API v2](https://www.hipchat.com/docs/apiv2/method/send_room_notification) it is optional.

 I'm currently using *room tokens* with my app and the sender of the messages is the **label** from the token. As the documentation says, The field `form` is `A label to be shown in addition to the sender's name`.

IMO, we must send it only when its value is not `null`.

apfelbox:
@andrioli you are right! The "Reporting Apps" in my screenshots is in fact the label of the token and the name is an optional *additional* label.

I agree, we should only send it if it is non-null. I will change that tomorrow.

apfelbox:
I see that I also need to update the tests accordingly to the new order of the parameters. I will do that tomorrow, too.

apfelbox:
I updated the pull requests.

For v1 it now always sends the from name (casting it to string, so that a `null` name isn't omitted from the query string when building with `http_build_query`).

I also updated the tests.

Seldaek:
Thanks

PR #688, opened 2015-11-19
Add php-cs-fixer to tests

bd808:
Add a step using `php-cs-fixer -vv fix --dry-run` to the `composer test`
script command used by Travis CI tests.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "swiftmailer/swiftmailer": "~5.3",
         "php-console/php-console": "^3.1.3",
         "phpunit/phpunit-mock-objects": "2.3.0",
-        "jakub-onderka/php-parallel-lint": "0.9"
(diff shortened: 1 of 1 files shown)


bd808:
The PHP7 error looks like an unrelated problem in the Travis CI VM. The PHP 5.3.3 error is however caused by php-cs-fixer's requirement of PHP >=5.3.6. That requirement is not arbitrary; see https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/4e4a50aa25a2e8b6a320ccf672ad8386323d2e30

Bumping the minimum version for Monolog to 5.3.6 as well (or just dropping the explicit 5.3.3 test suite) might not be the end of the world, but that's not my call to make.

Seldaek:
Thanks but I don't want to enforce this in PRs, it adds friction to the PR submitting process IMO, I just pointed it out so you know about it but it's not a huge deal either way :)

Issue #691, opened 2015-11-25
Logger not handling an associate array of handlers with named keys


mttjohnson:
I ran into an issue with Magento 2, *using Monolog out of the box* and the way that Magento 2 is injecting an array of handlers into the \Monolog\Logger instances they are all getting passed in as an associative array with named keys.

I entered an [issue on the Magento repo regarding this on their end](https://github.com/magento/magento2/issues/2529) in case there are any alternative to passing in the list of handlers, but this seems to be primarily surrounding the way arrays are getting managed in the \Monolog\Logger::addRecord() function.

When addRecord() iterates over the list of handlers it needs to process it inadvertantly is expecting it to be iterating over an indexed array, and fails to process any additional handlers besides the first handler it found when iterating over an associative array containing named keys.

```
        while (isset($this->handlers[$handlerKey]) &&
            false === $this->handlers[$handlerKey]->handle($record)) {
            $handlerKey++;
        }
```

This causes it to correctly process the first handler, but fails to call any additional handlers if it is expecting to bubble the record to additional handlers.

PR #692, opened 2015-11-26
Logger iterating over an associate array of handlers

mttjohnson:
Changed iterating over the $handlers array to allow for correct sequential processing regardless of array containing indexed numerical keys, or associative string keys in the array.

It iterates over the entire array until it finds the $handlerKey that was identified earlier using isHandling(). Once the starting position is found it will execute handle() on each handler unless a handler returns true indicating the handler completed the processing of the record, bubbling to the next handler should not occur, and the loop breaks.

This fixes an issue where an associative array of handlers is passed into the Logger constructor when instantiated.
magento/magento2#2529
Seldaek/monolog#691

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -315,9 +315,17 @@ class Logger implements LoggerInterface
         foreach ($this->processors as $processor) {
             $record = call_user_func($processor, $record);
         }
-        while (isset($this->handlers[$handlerKey]) &&
-            false === $this->handlers[$handlerKey]->handle($record)) {
-            $handlerKey++;
+        $foundStartingKey = false;
+        foreach ($this->handlers as $key => $handler) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, I tweaked things a bit though to only loop over the array once and not search for the key again, see the commit linked above if you like :)

mttjohnson:
Yes, the use of array pointers and a single loop is very nice. I was wanting to figure out how to incorporate that into the change, but was trying to keep my change isolated to the section that was causing the problem, and wasn't sure how much outside of that scope I should mess with. I also appreciate the test, as I still need to dive into building/running unit tests more.

Seldaek:
No worries it wasn't a bad fix per se but since it's quite a critical 
path for the library I try to keep it as fast as possible.

PR #698, opened 2015-12-06
Add handleBatch implementation for PhpAmqpLib

rdrenth:
I've implemented batch handling using the [PhpAmqpLib batch methods](https://github.com/videlalvaro/php-amqplib/blob/master/demo/batch_publish.php).
The PHP AMQP extension does not support batch handling, so this is only implemented for PhpAmqpLib.

### DIFF
--- a/src/Monolog/Handler/AmqpHandler.php
+++ b/src/Monolog/Handler/AmqpHandler.php
@@ -55,18 +55,12 @@ class AmqpHandler extends AbstractProcessingHandler
     protected function write(array $record)
     {
         $data = $record["formatted"];
-
-        $routingKey = sprintf(
-            '%s.%s',
-            // TODO 2.0 remove substr call
-            substr($record['level_name'], 0, 4),
-            $record['channel']
-        );
+        $routingKey = $this->getRoutingKey($record);
 
         if ($this->exchange instanceof AMQPExchange) {
             $this->exchange->publish(
                 $data,
-                strtolower($routingKey),
+                $routingKey,
                 0,
                 array(
                     'delivery_mode' => 2,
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/AmqpHandler.php
```diff
@@ -75,17 +69,72 @@ protected function write(array $record)
...
         }
+
+        $this->exchange->publish_batch();
+    }
+
+    /**
+     * Gets the routing key for the AMQP exchange
+     *
+     * @param array $record
+     * @returns string
```
stof: typo: ``@return``

src/Monolog/Handler/AmqpHandler.php
```diff
@@ -75,17 +69,72 @@ protected function write(array $record)
...
+     */
+    protected function getRoutingKey(array $record)
```
stof: this should be private IMO

src/Monolog/Handler/AmqpHandler.php
```diff
@@ -75,17 +69,72 @@ protected function write(array $record)
...
+        );
+
+        return strtolower($routingKey);
+    }
+
+    /**
+     * @param string $data
+     * @return AMQPMessage
+     */
+    protected function createAmqpMessage($data)
```
stof: same here

rdrenth:
@stof I've updated the PR according to you feedback. And yeah, making functions protected by default is an old habbit i need to get rid of :)

Seldaek:
Thanks, I merged it in the 1.x  branch, see https://github.com/Seldaek/monolog/commit/0d2bef057998dd3294808d906aa6b3f71dc59439

rdrenth:
Awesome, thanks!

Issue #699, opened 2015-12-07
Handler for Logging via consuming an API


paragdakle:
Hello,

I have been using Monolog with Laravel 5.1 and I really like the simplicity of the logger.

However, in my project we had a central logging server which exposed an API for logging. I failed to find any Logging Handler for the same purpose.

I think such kind of an handler will be useful as it will provide the developer the flexibility to consume any API for logging.

Cheers,

xabbuh:
Are you sure that a handler in Monolog is the best solution? I mean the actual API endpoints will probably differ with each use case and you can quite easily implement a custom handler in your application using the [cURL Util class](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/Curl/Util.php).

paragdakle:
Yes. I agree, that actual API endpoints will differ, but if we have a generic basic Handler which does the task and will work for majority of the cases, from a developers perspective, writing a new handler is always more tedious then adding a few tweaks to an existing one. I have added a basic generic handler and pull request for the same also has been initiated.

Seldaek:
Closing as per https://github.com/Seldaek/monolog/pull/700#issuecomment-190781061

PR #700, opened 2015-12-07
Handler for Logging via consumption of an API

paragdakle:
Handler for the #699 issue created.


Seldaek:
This doesn't really strike me as a general purpose handler that I'd like to include. Please don't take it the wrong way but I just can't include everything as I then have to maintain it forever.

You are very welcome to publish your handler as a package however, there are already a bunch that exist, see https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

Issue #197, opened 2013-05-21
2.0 Code cleanups / tasks


Seldaek:
- [x] ~~Clean up arg orders of handlers that have had added args pushed at the end (after bubbling etc), maybe group them in an $options array or parameter objects? Needs RFC. Check #801 too about fluent setters~~
- [x] add scalar hints and return type hints where relevant ("only" handlers left to go through)
- [x] ~~figure out how to handle network timeouts globally somehow, see #733~~
- [x] Add a close() method on the Logger to close all its handlers? What about resetting FingersCrossedHandler?
- [x] Define which php version should be supported (=> 7.1)
- [x] Add a timezone arg to Logger instead of the static timezone?
- [x] Turn static::getLevelName into static::
- [x] Set Monolog\Logger::API = 2 to allow users to adjust
- [x] Remove non PSR-3 methods to add record
- [x] Remove processor/formatter setters and getters from the HandlerInterface
- [x] Remove logstash V0 support or at least make V1 default
- [x] Remove legacy GELF handler support, c.f. #326
- [x] Remove datetime TODO in loggly handler
- [x] Move `Monolog\TestCase` to src/, see https://github.com/Seldaek/monolog/pull/356#issuecomment-40899938
- [x] ~~Split handlers w/ dependencies into packages i.e. http://php-and-symfony.matthiasnoback.nl/2014/04/theres-no-such-thing-as-an-optional-dependency/~~
- [x] ~~Change Logger constants to have RFC values (https://github.com/php-fig/log/issues/8)~~
- [x] Remove substr TODO in AmqpHandler
- [x] Remove default handler (`$this->pushHandler(new StreamHandler('php://stderr', static::DEBUG));`) from Monolog\Logger
- [x] ~~Consider adding a public method (send or write or so) on handlers to make reusing them out of monolog easier, c.f.~~ https://github.com/Seldaek/monolog/commit/f3b61cf48af2c0730f149ae48f930e12a3bec1f2#commitcomment-8237858
- [x] Consider using MongoDBFormatter as default in MongoDBHandler
- [x] Drop support for outdated mongo exts in MongoDBHandler
- [x] Remove BC __get() in SwiftMailerHandler
- [x] Update HipChat to API v2 using 01549ec4d05e2d3f33227531d9fc9454d5ef6049 and `room_id` and `from` in buildContent (+ `nickname` everywhere) should be removed. See if we can't detect the token format and warn users using a v1 token.
[...]

PR #706, opened 2015-12-20
Removed substr call in amqp routing key generator

acrobat:
Fixed todo from #197

> Remove substr TODO in AmqpHandler


src/Monolog/Handler/AmqpHandler.php
```diff
@@ -58,8 +58,7 @@ protected function write(array $record)
 
         $routingKey = sprintf(
```
dosten: IMO this should be in a single line

acrobat: Fixed!

acrobat:
Rebase branch and changed fix after changes in master

Seldaek:
Thanks

PR #707, opened 2015-12-20
Removed unnecessary php version checks after php dependency bump to 7.0

acrobat:
Fixed todo from #197 

> Check all PHP_VERSION and version_compare checks for relevance


Seldaek:
Great thanks

PR #713, opened 2016-01-05
Changed LOG_LOCALx constants to their numeric values in AbstractSyslogHandler.php

pulzarraider:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | -
| License       | MIT

LOG_LOCALx constants are unavailable in Windows so we cannot used them in cross-platform code, but the local facilities has to be defined to avoid throwing `Unknown facility value...` exceptions on Windows.

Our use case: We are using Symfony with enabled syslog logging in `config_prod.yml`:

```yml
monolog:
    handlers:
        syslog:
            type: syslog
            level: error
            ident: php-project
            facility: local3
```
Everything works OK on Linux, but in Windows (developer's machine) it ends with error: `Unknown facility value "local3" given`.

### DIFF
--- a/src/Monolog/Handler/AbstractSyslogHandler.php
+++ b/src/Monolog/Handler/AbstractSyslogHandler.php
@@ -61,16 +61,14 @@ abstract class AbstractSyslogHandler extends AbstractProcessingHandler
     {
         parent::__construct($level, $bubble);
 
-        if (!defined('PHP_WINDOWS_VERSION_BUILD')) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Fixed in https://github.com/Seldaek/monolog/commit/bf61d9ff3fba3bbed2abce486b175f31a807f384 - I rather not remove the constants on linux in case they change in the future for some reason.

Issue #366, opened 2014-05-13
Line breaks in ErrorLogHandler


anlutro:
It seems like on most systems, linebreaks in error_log calls are not supported, instead linebreaks are output as literal "\n" strings.

It seems like some sort of logic should happen in the ErrorLogHandler where it asks the Formatter to explode the log message into an array and then call error_log once on each line. Or maybe the handler could just `explode(PHP_EOL, $str)`, so as to not add a lot of complexity to the formatter(s).

Is this something that is doable or would it be out of the library's scope or simply not worth the effort?

PR #715, opened 2016-01-11
Fix \n characters in the ErrorHandler stack trace.

earncef:
The stack trace generated by the LineFormatter contains \n characters instead of newlines with ErrorHandler errors. This should fix it.


earncef:
On some further checking, this might also be related to #366

Seldaek:
AFAIK monolog never generates literal \n.. So unless you can figure out exactly where they come from in monolog, I would say that they must come from somewhere in your code. I don't really think replacing stuff in the output is a good idea. We do it for real new lines because those are special but I'm rather againts replacing other characters so I'll close this. If you find more details as to where they come from please do tell and we can always reopen.

PR #719, opened 2016-01-14
Add Understand.io handler, formatter & tests

aivis:
Add handler for [Understand.io](https://www.understand.io)

Handler uses `curl` command line tool to deliver data.

### DIFF
--- a/src/Monolog/Handler/UnderstandHandler.php
+++ b/src/Monolog/Handler/UnderstandHandler.php
@@ -1,0 +1,125 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Logger;
+use Monolog\Formatter\UnderstandFormatter;
+
+/**
+ * Sends records to Understand.io
+ *
+ * Requirement is that CURL command line tool is installed and functioning correctly. 
+ * To check whether CURL is available on your system, 
+ * execute the following command in your console "curl -h"
+ * If you see instructions on how to use CURL 
+ * then your system has the CURL binary installed and you can use the async handler.
+ * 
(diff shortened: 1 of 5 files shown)


src/Monolog/Handler/UnderstandHandler.php
```diff
@@ -0,0 +1,125 @@
...
+     * @param string $requestData
+     * @return void
+     */
+    protected function send($requestData)
+    {
+        $parts = array(
+            'curl',
+            '-X POST',
+            '-d',
+            escapeshellarg($requestData),
```
xabbuh: Why do you use the command-line here instead of using the [CURL Util class](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/Curl/Util.php)?

src/Monolog/Handler/UnderstandHandler.php
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * Execute call
+     * 
+     * @param string $cmd
+     * @return void
+     */
+    protected function exec($cmd)
+    {
+        exec($cmd);
```
xabbuh: Why do you use the command-line here instead of using the [CURL Util class](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/Curl/Util.php)?

aivis: Following code forks a separate curl process and allows PHP process to complete execution without waiting on HTTP response. It's faster because it doesn't block PHP process.

```php
exec('curl -XPOST -d \'{"message": "test"}\' https://api.understand.io/input-key > /dev/null 2>&1 &');
```
https://github.com/Seldaek/monolog/pull/719/files#diff-0473cdde8bb7699cd878c4c22c98bc5dR91

Seldaek:
Please publish this as a third party package and then add it to https://github.com/seldaek/monolog/wiki/Third-Party-Packages.

Taking more handlers in means I have to maintain them all which isn't really sustainable..

As for the curl fork, the right way to do this is to use UDP packets IMO, not forking processes for every log being sent. I don't know if understand.io supports that but that would be a better approach.

Issue #452, opened 2014-11-07
StreamHandler Pipe to Process?


holybit:
StreamHandler uses [fopen](http://php.net/manual/en/function.fopen.php) which is great for logging to files, but what about logging to another process? For example:

```
$log->pushHandler(new StreamHandler('|/usr/bin/mylogger.py', Logger::DEBUG));
```

Is there a way to do this currently in Monolog?

anlutro:
Have you tried creating a FIFO (named pipe) and use the StreamHandler to write to it?

jackminh:
I want to use monolog to log write redis, there is an error, how can I solve?

Seldaek:
@jackminh first of all you can report a new issue for things that don't have any connection with this one. And then if you have an error you might want to post the error, because otherwise we have no clue how to help.

Seldaek:
Fixed by #720 (only in the upcoming 2.0)

PR #720, opened 2016-01-15
Created "ProcessHandler", that logs records to the STDIN of a process.

ArSn:
This implements #452, I hope you find it useful.

**Some points/questions that I'd like to add here:**

1. I did intentionally not include an option to pass the already opened process to the handler for now, but rather require the consumer to pass a command. This is due to the circumstance that the `proc_open()` call has to be quite specific in order for the handler to work properly.
2. I am not entirely sure about the name. Although it was suggested in the linked issue, it might be true that "ProcessHandler" could cause some confusion, as there are classes like "ProcessableHandlerInterface" and "ProcessableHandlerTrait" in the codebase, which are unrelated. I could not think of anything better and therefore went with it for now, but please, suggestions are welcome!
3. Is there any documentation that I should update, like "How to use this handler" or such?

**Apart from that, I have a few ideas of how to extend this handler, but was not sure about the demand or use cases for that, and hence, left them out for now:**

1. There could be something like a "strategy" (probably also implementing the Strategy Pattern), in how the process is handled, e.g.: Open new process on every write; OR keep process open (like it does now); OR use a time based transient handling
2. There could be even more options in the constructor, as [`proc_open()`](http://php.net/proc_open) also:
    1. allows an `$env` parameter to pass environment variables to the process
    2. allows an `$other_options` parameter, that seems to be only doing anything on windows right now (I don't see very much use in implementing this one)
3. The [`proc_close()`](http://php.net/proc_close) function actually returns the termination status of the handled process. This could be returned from `ProcessHandler::close()`, as it *may* be helpful for some.

### DIFF
--- a/src/Monolog/Handler/ProcessHandler.php
+++ b/src/Monolog/Handler/ProcessHandler.php
@@ -115,7 +115,7 @@ class ProcessHandler extends AbstractProcessingHandler
 
 		$errors = $this->readProcessErrors();
 		if (empty($errors) === false) {
-			throw new \UnexpectedValueException('Errors while writing to process: ' . var_export($errors, true));
+			throw new \UnexpectedValueException(sprintf('Errors while writing to process: %s', $errors));
 		}
 	}
 
--- a/src/Monolog/Handler/ProcessHandler.php
+++ b/src/Monolog/Handler/ProcessHandler.php
@@ -26,197 +26,197 @@ use Monolog\Logger;
  */
 class ProcessHandler extends AbstractProcessingHandler
 {
-	/**
-	 * Holds the process to receive data on its STDIN.
-	 *
-	 * @var resource
-	 */
-	private $process;
+    /**
+     * Holds the process to receive data on its STDIN.
+     *
+     * @var resource
+     */
+    private $process;
 
-	/**
-	 * @var string
-	 */
-	private $command;
+    /**
+     * @var string
+     */
+    private $command;
 
-	/**
-	 * @var string
-	 */
-	private $cwd;
+    /**
+     * @var string
+     */
+    private $cwd;
 
-	/**
-	 * @var array
-	 */
-	private $pipes = [];
+    /**
+     * @var array
+     */
+    private $pipes = [];
 
-	/**
-	 * @var array
-	 */
-	const DESCRIPTOR_SPEC = [
-		0 => ['pipe', 'r'],  // STDIN is a pipe that the child will read from
-		1 => ['pipe', 'w'],  // STDOUT is a pipe that the child will write to
-		2 => ['pipe', 'w'],  // STDERR is a pipe to catch the any errors
-	];
+    /**
+     * @var array
+     */
+    const DESCRIPTOR_SPEC = [
+        0 => ['pipe', 'r'],  // STDIN is a pipe that the child will read from
+        1 => ['pipe', 'w'],  // STDOUT is a pipe that the child will write to
+        2 => ['pipe', 'w'],  // STDERR is a pipe to catch the any errors
+    ];
 
-	/**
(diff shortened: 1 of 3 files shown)


stof:
> This could be returned from ``ProcessHandler::close()``, as it may be helpful for some.

this is a bad idea, as it would be inconsistent with the interface

src/Monolog/Handler/ProcessHandler.php
```diff
@@ -0,0 +1,222 @@
...
+	 */
+	protected function write(array $record)
+	{
+		$this->ensureProcessIsStarted();
+
+		$this->writeProcessInput($record['formatted']);
+
+		$errors = $this->readProcessErrors();
+		if (empty($errors) === false) {
+			throw new \UnexpectedValueException('Errors while writing to process: ' . var_export($errors, true));
```
stof: why using var_export ? this is not human readable as it escape things

tests/Monolog/Handler/ProcessHandlerTest.php
```diff
@@ -0,0 +1,206 @@
...
+	 */
+	public function testConstructWithInvalidCwdThrowsInvalidArgumentException($invalidCwd)
+	{
+		$mockBuilder = $this->getMockBuilder('Monolog\Handler\ProcessHandler');
+		$mockBuilder->disableOriginalConstructor();
+		$handler = $mockBuilder->getMock();
+
+		$this->setExpectedException('\InvalidArgumentException');
+		/** @var ProcessHandler $handler */
+		$handler->__construct('fake command', Logger::DEBUG, true, $invalidCwd);
```
stof: don't use a partial mock to test the exception. Just perform a normal instantiation. It will be much more readable, and more reliable (as it will test the real thing it want to test)

src/Monolog/Handler/ProcessHandler.php
ArSn: @stof This was a leftover from when it was still an array, thanks for pointing out! Fixed.

Seldaek:
Thanks for the PR! Please note that we use 4 spaces for indenting here, it'd be great if you can update the files.

tests/Monolog/Handler/ProcessHandlerTest.php
ArSn: @stof I have to agree again, thanks for pointing that out. For some reason it also lead to letting the first travis CI build fail: https://travis-ci.org/Seldaek/monolog/jobs/102658703

I tried to debug that, however, it complained about the setLevel() not matching the signature - as far as I could see this method was not even implemented in the mock code, which is why I find this quite odd - any idea why that error occured?

ArSn:
@Seldaek Certainly! Done. :)

Seldaek:
Also could you please re-open this PR against the 1.x branch? Because master only supports php7.

ArSn:
@Seldaek Hmm. I am using the bracket syntax for some arrays, which only became available in PHP 5.4., the 1.x branch specifies `"php": ">=5.3.0",` however. Should I really backport that? I mean, just so one would maybe change it in the future back again? What is the point of having it in the 1.x branch already, too? Of course, it would be faster available in a stable release, but, as the issue has been inactive since march of 2015, I don't feel too much pressure here.

What do you think?

Seldaek:
If it's fine for you it's also ok for me. I don't think it's urgent but if
you do want to use the code in a release then I'd recommend backporting it.
A few square brackets aren't such a big deal :)

ArSn:
> If it's fine for you it's also ok for me.

Ok, it's fine for me :)

Let's leave it at that then. Nothing I need in production urgently either. Maybe this is even one more feature encouraging people to upgrade to a stable `>=php7` version of monolog when it comes. ;)

PR #723, opened 2016-01-20
Raven client fingerprint support

damour:
Raven client [fingerprint] (https://github.com/getsentry/raven-php/blob/master/docs/usage.rst#optional-attributes) support.


Seldaek:
Thanks

Issue #695, opened 2015-12-02
MongoDBHandler can use ext-mongodb instead of userland library


jmikola:
See: https://github.com/Seldaek/monolog/pull/676#issuecomment-161344479

Since the handler class is constructed with the `$database` and `$collection` names as strings, we can receive a `MongoDB\Driver\Manager` instance and log messages with:

```
$bulk = new MongoDB\Driver\BulkWrite;
$bulk->insert($document);
$manager->executeBulkWrite($database . '.' . $collection, $bulk);
```

`executeBulkWrite()` can take an optional WriteConcern as its third argument, but it doesn't look like Monoglog is using that anyway.

Seldaek:
ping @jmikola I can't assign you to the issue as you're not a contributor so.. just leave this notification unread as a TODO :p

jmikola:
Too late. I already had the tab open. Please try again!

/me closes tab

Seldaek:
ping @jmikola (this is getting absurd) :p

jmikola:
Related PR: #724

PR #724, opened 2016-01-20
Address issue #695

billybanfield:
Add manager as a possible argument to the MongoDBHandler so that writing can be done directly through the extension.


src/Monolog/Handler/MongoDBHandler.php
```diff
@@ -28,14 +28,20 @@
...
     public function __construct($mongo, $database, $collection, $level = Logger::DEBUG, $bubble = true)
     {
-        if (!($mongo instanceof \MongoClient || $mongo instanceof \Mongo || $mongo instanceof \MongoDB\Client)) {
+        if (!($mongo instanceof \MongoClient || $mongo instanceof \Mongo || $mongo instanceof \MongoDB\Client || $mongo instanceof \MongoDB\Driver\Manager)) {
             throw new \InvalidArgumentException('MongoClient, Mongo or MongoDB\Client instance required');
         }
-
-        $this->mongoCollection = $mongo->selectCollection($database, $collection);
+        $this->namespace = "$database.$collection";
+        if($mongo instanceof \MongoDB\Driver\Manger) {
```
jmikola: CS: space between `if` and `(`

src/Monolog/Handler/MongoDBHandler.php
```diff
@@ -28,14 +28,20 @@
...
 
     public function __construct($mongo, $database, $collection, $level = Logger::DEBUG, $bubble = true)
     {
-        if (!($mongo instanceof \MongoClient || $mongo instanceof \Mongo || $mongo instanceof \MongoDB\Client)) {
+        if (!($mongo instanceof \MongoClient || $mongo instanceof \Mongo || $mongo instanceof \MongoDB\Client || $mongo instanceof \MongoDB\Driver\Manager)) {
             throw new \InvalidArgumentException('MongoClient, Mongo or MongoDB\Client instance required');
         }
-
-        $this->mongoCollection = $mongo->selectCollection($database, $collection);
+        $this->namespace = "$database.$collection";
```
jmikola: The namespace is only applicable for the `MongoDB\Driver\Manager` case, so we can probably move this within the conditional below. Typically, `$database . '.' . $collection` is preferred over variable name expansion in double-quoted strings.

src/Monolog/Handler/MongoDBHandler.php
```diff
@@ -44,8 +50,12 @@ protected function write(array $record)
     {
         if ($this->mongoCollection instanceof \MongoDB\Collection) {
             $this->mongoCollection->insertOne($record["formatted"]);
+        } else if($this->mongoCollection instanceof \MongoCollection) {
+            $this->mongoCollection->insert($record["formatted"]);
```
jmikola: Rather than `else if`, I'd suggest a series of `if` blocks that return:

```php
if ($this->mongoCollection instanceof \MongoDB\Collection) {
    $this->mongoCollection->insertOne($record['formatted']);

    return;
}

// $this->manager instanceof  \MongoDB\Driver\Manager
if ($this->mongoCollection instanceof \MongoCollection) {
    $this->mongoCollection->insert($record['formatted']);

    return;
}

$bulk = new \MongoDB\Driver\BulkWrite;
$bulk->insert($record['formatted']);
$this->$manager->executeBulkWrite($this->namespace, $bulk);
```

src/Monolog/Handler/MongoDBHandler.php
```diff
@@ -28,14 +28,20 @@
 class MongoDBHandler extends AbstractProcessingHandler
 {
     protected $mongoCollection;
+    protected $namespace;
+    protected $manager;
```
jmikola: I'd suggest making all three of these private. If needed, they can be opened up later.

src/Monolog/Handler/MongoDBHandler.php
```diff
@@ -28,14 +28,20 @@
...
 
     public function __construct($mongo, $database, $collection, $level = Logger::DEBUG, $bubble = true)
```
jmikola: We can add a doc block here:

```
/**
 * Constructor.
 *
 * @param Client|Manager|Mongo|MongoClient $mongo      MongoDB driver or library instance
 * @param string                           $database   Database name
 * @param string                           $collection Collection name
 * @param int                              $level      The minimum logging level at which this handler will be triggered
 * @param Boolean                          $bubble     Whether the messages that are handled can bubble up the stack or not
 */
```

jmikola:
General feedback. The various MongoDB driver/library classes can be given `use` statements at the top of the file, so we can avoid full namespaces in the code. This will also ensure the API doc generator picks them up:

```
use MongoDB\Driver\Manager;
use MongoDB\Client;
use Mongo;
use MongoClient;
```

The class-level doc block can also be improved:

```
/**
 * Logs to a MongoDB database.
 *
 * Usage example:
 *
 *   $handler = new MongoDBHandler(new MongoDB\Client, 'logs', 'prod');
 *   $logger = new Logger('application');
 *   $logger->pushHandler($handler);
 *
 * The above examples uses the MongoDB PHP library's client class; however,
 * classes from ext-mongodb (MongoDB\Driver\Manager) and ext-mongo (Mongo and
 * MongoClient) are also supported.
 */
```

----

@Seldaek: Are you actively using `@author` tags? Just asking because we recently decided to remove them from the ODM (they're prone to getting outdated, and git already tells us who worked on a file). If so, should @billybanfield add himself as an author?

src/Monolog/Handler/MongoDBHandler.php
billybanfield: - [x]

src/Monolog/Handler/MongoDBHandler.php
billybanfield: done.

src/Monolog/Handler/MongoDBHandler.php
billybanfield: done

src/Monolog/Handler/MongoDBHandler.php
billybanfield: done.

src/Monolog/Handler/MongoDBHandler.php
billybanfield: done

src/Monolog/Handler/MongoDBHandler.php
Seldaek: @billybanfield if it's done you forgot to push?! Because here it's not done :)

Seldaek:
No we don't actively use @author, it's usually just a nod at the initial author of the file and anyone having done major contributions maybe, but I don't think anyone doing minor stuff should go there, and yes git history is the most valuable source really.

billybanfield:
moved tests to correct commit

src/Monolog/Formatter/MongoDBFormatter.php
```diff
@@ -100,6 +102,13 @@ protected function formatException(\Exception $exception, $nestingLevel)
...
     {
-        return new \MongoDate($value->getTimestamp());
+        $seconds = (int) $value->format('U');
+        $milliseconds = (int) $value->format('u') / 1000;
+
+        if ($seconds < 0) {
+            return new UTCDateTime($seconds * 1000 - $milliseconds);
+        } else {
+            return new UTCDateTime($seconds * 1000 + $milliseconds);
+        }
```
jmikola: @Seldaek: I'm not sure what the point of this class was, as it was never the default formatter for the MongoDB handler. I suppose people may be using it manually?

In any event, this update was required for the new driver's date class. This math is prone to causing integer overflow on 32-bit systems, but that shouldn't be an issue for PHP 7, since it supports 64-bit ints on all platforms.

src/Monolog/Handler/MongoDBHandler.php
```diff
@@ -11,41 +11,66 @@
...
+ * Usage example:
  *
- *   $log = new Logger('application');
- *   $mongodb = new MongoDBHandler(new \Mongo("mongodb://localhost:27017"), "logs", "prod");
+ *   $log = new \Monolog\Logger('application');
+ *   $client = new \MongoDB\Client('mongodb://localhost:27017');
+ *   $mongodb = new \Monolog\Handler\MongoDBHandler($client, 'logs', 'prod');
  *   $log->pushHandler($mongodb);
  *
- * @author Thomas Tourlourat <thomas@tourlourat.com>
```
jmikola: This class was essentially rewritten from the 1.x version, so I didn't see much reason to keep `@author` around.

tests/Monolog/Handler/MongoDBHandlerTest.php
```diff
@@ -21,45 +22,57 @@ class MongoDBHandlerTest extends TestCase
...
     {
-        public function selectCollection()
-        {
+        if (!(class_exists('MongoDB\Driver\Manager'))) {
+            $this->markTestSkipped('ext-mongodb not installed');
+        }
+
+        /* This can become a unit test once ManagerInterface can be mocked.
+         * See: https://jira.mongodb.org/browse/PHPC-378
+         */
```
jmikola: I noticed that some of the other test (e.g. CouchDB) were functional/integration instead of unit. We'll need to do the same here until we have mockable interfaces in the extension, since the core classes themselves are all final.

src/Monolog/Formatter/MongoDBFormatter.php
Seldaek: Yeah it was an optional formatter.. I'm not sure if it is still relevant with the latest mongo clients or not. I know nothing about mongo :)

Seldaek:
So this is only for the master branch and drops support for old mongo client versions right? It now only supports latest client + extension AFAIK. Fine by  me as master targets 2.0.

jmikola:
@Seldaek: Correct. The legacy driver doesn't (and won't) support PHP 7 and HHVM, so it made sense to remove it.

src/Monolog/Formatter/MongoDBFormatter.php
jmikola: It was obvious that this date formatting needed an update, but I didn't read into the rest of the class enough to see how this differs from NormalizerFormatter. All we're doing here is storing MongoDB's native date type instead of a string date produced by NormalizerFormatter.

ZergMark:
@billybanfield I made a small change and [a PR](https://github.com/billybanfield/monolog/pull/2) on your branch to make this a little more PHP7 specific.

tests/Monolog/Handler/MongoDBHandlerTest.php
stof: Well, technically, having a functional test is better for them. Otherwise, you test the handler against your assumption of the mongodb behavior, but there is nothing guaranteeing that the handler actually works with a real mongo

tests/Monolog/Handler/MongoDBHandlerTest.php
```diff
@@ -21,45 +22,57 @@ class MongoDBHandlerTest extends TestCase
...
+        $mongodb = new Manager('mongodb://localhost:27017');
+        $handler = new MongoDBHandler($mongodb, 'test', 'monolog');
+        $record = $this->getRecord();
+
+        try {
+            $handler->handle($record);
+        } catch (\RuntimeException $e) {
+            $this->markTestSkipped('Could not connect to MongoDB server on mongodb://localhost:27017');
```
stof: this should rather be checked earlier than the ``->handle`` call, to avoid catching exceptions coming from Monolog.

And the test should have an assertion in case of success (checking that the mongo collection contains the record) ?

PR #725, opened 2016-01-21
Allow GroupHandler to set a formatter to all nested handlers

eexit:
Hello,

I would like to know if it's possible to add such shortcut for the `GroupHandler`?

Thanks,


Seldaek:
Why not I guess..

eexit:
Thanks @Seldaek!

eexit:
Should I port the change to the master branch?

Seldaek:
No I'll merge 1.x into master myself it's easier to keep the branches consistent that way I think but thanks :)

PR #727, opened 2016-01-23
add release number to every log

greg0ire:
This adds an internal release number to the raven handler. The release
number is added to what is sent to sentry unless something already is
present because a release number was sent via "context" or "extra".

An abstraction might be introduced somewhere to represent release number generators, so that different generators might be implemented, but I am not sure where, and I am also unsure whether such a class should be injected instead of the release number. In short, should it work like this : 

```
$releaseNumberGenerator = new GitDescribeReleaseNumberGenerator();
$ravenHandler->setRelease($releaseNumberGenerator->generate());
```

or like this : 

```
$ravenHandler->setReleaseGenerator($releaseNumberGenerator);
// the handler calls generate() internally
```

or if both things should be possible.

Finally, I named the internal variable "release" instead of "releaseNumber" because of what already existed, but I am ok to change that.

### DIFF
--- a/src/Monolog/Handler/RavenHandler.php
+++ b/src/Monolog/Handler/RavenHandler.php
@@ -39,6 +39,12 @@ class RavenHandler extends AbstractProcessingHandler
     );
 
     /**
+     * @var string should represent the current version of the calling
+     *             software. Can be any string (git commit, version number)
+     */
+    private $release;
+
+    /**
      * @var Raven_Client the client object that sends the message to the server
      */
     protected $ravenClient;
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/RavenHandler.php
```diff
@@ -208,4 +218,11 @@ protected function getExtraParameters()
     {
         return array('checksum', 'release');
     }
+
+    public function setRelease($value)
```
prgTW:
```
/**
 * @param string $value
 */
```
?

greg0ire: Sure, why not? Fixed :)

prgTW:
:+1

greg0ire:
ping @Seldaek

Seldaek:
Thanks

PR #728, opened 2016-02-03
Instantiate LineFormatter only when configured to use short attachment

danizord:
We don't need LineFormatter when not configured to use short attachment.
I think current tests will be enough to cover this behavior.


Seldaek:
Good catch thanks

PR #731, opened 2016-02-14
Added new Formatter to support Bunyan logging

Lunatic666:
I extended the JsonFormatter with some extra fields

### DIFF
--- a/src/Monolog/Formatter/BunyanFormatter.php
+++ b/src/Monolog/Formatter/BunyanFormatter.php
@@ -1,0 +1,136 @@
+<?php
+
+/*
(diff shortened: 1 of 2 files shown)


Seldaek:
Please release this as a standalone package instead. I'd rather not maintain too esoteric things forever as I have no domain knowledge and it makes me a bottleneck which isn't good.

You can add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages once released to allow people to find it more easily.

Lunatic666:
Thanks, no problem. I just added it to the Wiki.

PR #734, opened 2016-03-02
GearmanHandler

VBelozyorov:
Simple handler for [Gearman](http://php.net/manual/ru/book.gearman.php) support.

### DIFF
--- a/src/Monolog/Handler/GearmanHandler.php
+++ b/src/Monolog/Handler/GearmanHandler.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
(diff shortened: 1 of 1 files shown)


VBelozyorov:
TODO: tests and docs :)

Seldaek:
Thanks. However please release this as a standalone package instead. I'd rather not maintain too esoteric things forever as I have no domain knowledge and it makes me a bottleneck which isn't good.

You can add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages once released to allow people to find it more easily!

PR #738, opened 2016-03-04
Fixed small syntax bug with BrowserConsoleHandler preventing console …

johnhunt:
…output

Replaced commas in send() method when calling generateScript() - this was
preventing any javascript from being output


johnhunt:
Done in a bit of a rush as I now have family commitments, you get the idea anyway.

Seldaek:
Thanks, I'll pull it in 1.x no worries :)

PR #740, opened 2016-03-08
Fix content_type for AMQP Message

macintoshplus:
If use the AmqpHandler with the php-ext amqp, the content type is `plain/text` for each message in the queue because the property name is wrong written.

This pull request, change the property name `Content-type` by correct property name `content_type`.

The patch work fine with RabbitMQ.

### DIFF
--- a/src/Monolog/Handler/AmqpHandler.php
+++ b/src/Monolog/Handler/AmqpHandler.php
@@ -65,7 +65,7 @@ class AmqpHandler extends AbstractProcessingHandler
                 0,
                 array(
                     'delivery_mode' => 2,
-                    'Content-type' => 'application/json',
+                    'content_type' => 'application/json',
                 )
             );
         } else {
--- a/tests/Monolog/Handler/AmqpHandlerTest.php
+++ b/tests/Monolog/Handler/AmqpHandlerTest.php
@@ -65,7 +65,7 @@ class AmqpHandlerTest extends TestCase
             0,
             array(
                 'delivery_mode' => 2,
-                'Content-type' => 'application/json',
+                'content_type' => 'application/json',
             ),
         );
 


Seldaek:
https://www.rabbitmq.com/tutorials/amqp-concepts.html says:

> AMQP peers typically use the "content-type" and "content-encoding" fields to communicate this information, but this is by convention only.

This does not really reassure me that this PR is correct.. Any clue @videlalvaro? :)

videlalvaro:
Per protocol spec it should be `content-type`, but not sure how the library maps this internally. See https://www.rabbitmq.com/resources/specs/amqp-xml-doc0-9-1.pdf

The other PHP library uses underscores as well: https://github.com/php-amqplib/php-amqplib/blob/master/PhpAmqpLib/Message/AMQPMessage.php#L28 so if by mistake your message is created using `content-type` then it won't be encoded AFAIK, and maybe that's the problem @macintoshplus is referring to

Seldaek:
OK thanks. It seems AMQPExchange from the amqp extension uses Content-type http://php.net/manual/sr/amqpexchange.publish.php

Depends on the docs language tohugh, http://php.net/manual/pl/amqpexchange.publish.php shows content_type.. So maybe it changed? @macintoshplus could you try what happens if we send both? Would that be a good way to work with all versions? Ping @lstrojny any help here would be welcome, if we can check for the extension version or something that'd be the best I guess.

macintoshplus:
Hi,

RabbitMQ ignore all unknown properties. I don't know other AMQP implementation.

Send twice key for the content-type, should be a good idea.

videlalvaro:
Why send it twice and not send the proper one, assuming RabbitMQ is the
target broker?

Seldaek:
My point was that I don't think the code is wrong, but just outdated. If the php AMQP extension indeed changed this then depending on the installed version we should send A or B. That's why I want to hear from the maintainers there before taking action either way.

lstrojny:
@Seldaek just go with `content_type`, `Content-type`, according to the history, never worked.

Seldaek:
OK thanks :)

PR #741, opened 2016-03-09
Update RedisHandler.php

shiyangwu520:
Should be the exec method, not the execute method


Seldaek:
Thanks, merged in d04afcb5d0a59b95f98a55502e078a7d6fcb557f

PR #742, opened 2016-03-12
Add milliseconds to ElasticaFormatter as default

mablae:
Milliseconds are more precise. The same change was already made for the `LogstashFormatter` in 899f7f52e5014feac069319df22e25cbd3b34d5f

Tested against ES 1.7.x and Kibana 4.1


Seldaek:
Thanks, merged in 1.x as well as master!

PR #744, opened 2016-03-15
move from abandoned package to maintained package

greg0ire:


stloyd:
:+1:

greg0ire:
Wow that was quick

Seldaek:
No brainer so no problem to review ;)

ismailbaskin:
@Seldaek could you tag this commit?

stof:
@ismailbaskin this is only a dev requirement, so it does not affect any project using Monolog. This does not justify a tag by itself

stof:
@greg0ire you forgot to update the ``suggest`` section, which is also referencing this package

greg0ire:
Good catch @stof ! See #750

gries:
Hi,
Could you merge it into the 1.x branch too? Also why not use the current 2.6.x Version instead of 2.4.x?

greg0ire:
@gries `~` means the last number can change. So `~2.4` actually means `2.x`, with `4 <= x`

gries:
thx for clearing that up, any chance of merging it in the 1.x branch anyway? I'm currently getting problems because we use the "php-amqplib/php-amqplib" version in our projects but monolog still requires the videlalvaro version.

stof:
@gries monolog only has a dev requirements and a suggestion, so this does not impact the projects using monolog

gries:
In this case it affects the projects if I install it with the --dev switch.
"php-amqplib/php-amqplib" and "videlalvaro/php-amqplib" use the same Namespace **PhpAmqpLib\**
So right now using monolog and php-amqplib/php-amqplib when installing --dev dependencies installs two different Versions of the class: **PhpAmqpLib\Message\AMQPMessage** for example.

greg0ire:
Do you really think `--dev` implies installing the dev dependencies of all your dependencies?

gries:
hmm good point my bad, maybe I should have more coffee before diving into dependency hell bugs :S

greg0ire:
You can use `composer depends videlalvaro/php-amqplib` to see what exactly requires the legacy lib.

PR #756, opened 2016-03-25
missing property

remicollet:
This property is used, but not declared, in writingIsTimedOut

Don't know why it pass on travis, but locally I got

1) Monolog\Handler\SocketHandlerTest::testAvoidInfiniteLoopWhenNoDataIsWrittenForAWritingTimeoutSeconds
Undefined property: Mock_SocketHandler_4c030aea::$lastWritingAt

This trivial fix fix this

### DIFF
--- a/src/Monolog/Handler/SocketHandler.php
+++ b/src/Monolog/Handler/SocketHandler.php
@@ -30,6 +30,7 @@ class SocketHandler extends AbstractProcessingHandler
     private $persistent = false;
     private $errno;
     private $errstr;
+    private $lastWritingAt;
 
     /**
      * @param string  $connectionString Socket connection string


remicollet:
Travis failure is not related to this change (related to phpunit/mockobject 3.1.0)

Please consider for both master and 1.x

stof:
> Please consider for both master and 1.x

in this case, it is easier to send the PR to the 1.x branch, as we can then merge branches together. I don't know whether @Seldaek has the tool created for Symfony to rebase PRs on another branch automatically.

stof:
> Travis failure is not related to this change (related to phpunit/mockobject 3.1.0)

These failures indicate bugs in the testsuite though (which were silently accepted in older versions), so it would be worth sending a PR to fix them if you have some time for it.

remicollet:
=> PR #757

PR #765, opened 2016-04-07
(slack) fix for missing messages

CyberLine:
while debugging why slack sometimes receives a message but does not send it to the chat i found out, that the solution is, to read the response from the buffer.


Seldaek:
We fixed that by adding https://github.com/Seldaek/monolog/blob/1.x/src/Monolog/Handler/SlackHandler.php#L246  in 5085630fd606c50d9b09efa0f93c0bdd83603d24 as far as I can tell. Are you still experiencing issues with the latest release?

CyberLine:
Yeah, that fixed it! Thx.

PR #769, opened 2016-04-14
Provide means to manually activate a FingersCrossedHandler

naderman:
This is useful if you have a FingersCrossedHandler set up and some where in your application are interested in all output regardless of whether the threshold is actually reached. For example to temporarily debug some problem. One can replace the handler in that spot, but this is signifcantly easier to use.

### DIFF
--- a/src/Monolog/Handler/FingersCrossedHandler.php
+++ b/src/Monolog/Handler/FingersCrossedHandler.php
@@ -36,6 +36,7 @@ class FingersCrossedHandler extends AbstractHandler
     protected $buffer = array();
     protected $stopBuffering;
     protected $passthruLevel;
+    protected $overrideActivated = false;
 
     /**
      * @param callable|HandlerInterface       $handler            Handler or factory callable($record, $fingersCrossedHandler).
@@ -80,6 +81,14 @@ class FingersCrossedHandler extends AbstractHandler
     }
 
     /**
+     * Manually activate this logger regardless of the activation strategy
+     */
+    public function activate()
+    {
+        $this->overrideActivated = true;
+    }
+
+    /**
      * {@inheritdoc}
      */
     public function handle(array $record)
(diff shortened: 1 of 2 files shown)


Seldaek:
You still need to push out some log though after calling this, so I wonder if it's really much better than to just send a dummy error to force it to write out? Considering the latter doesn't add edge-casey code that is.

naderman:
@Seldaek well I have alerting on my logs, and these definitely are not errors, so I'd rather not get paged over it. I just want to force the output, so I can track down background information in logs.

naderman:
I'd be fine with automatically writing through on `activate()` and stopping buffering depending on the internal settings though.

Seldaek:
Yeah I think it'd be better to have activate() basically act more like the BufferHandler flush(), which I guess is what you described. It'd remove the hackish flag, and make it also not need a new log record to trigger. Win/win.

src/Monolog/Handler/FingersCrossedHandler.php
```diff
@@ -96,17 +119,7 @@ public function handle(array $record)
...
-                }
-                if (!$this->handler instanceof HandlerInterface) {
-                    $this->handler = call_user_func($this->handler, $record, $this);
-                    if (!$this->handler instanceof HandlerInterface) {
-                        throw new \RuntimeException("The factory callable should return a HandlerInterface");
-                    }
-                }
-                $this->handler->handleBatch($this->buffer);
-                $this->buffer = array();
+                $this->activate($record);
```
Seldaek: $record shouldn't be there

src/Monolog/Handler/FingersCrossedHandler.php
```diff
@@ -80,6 +80,29 @@ public function isHandling(array $record)
...
     /**
+     * Manually activate this logger regardless of the activation strategy
+     */
+    public function activate()
+    {
+        if ($this->stopBuffering) {
+            $this->buffering = false;
+        }
+        if (!$this->handler instanceof HandlerInterface) {
+            $record = end($this->buffer);
```
Seldaek: `$record = end($this->buffer) ?: null;` could replace those 4 lines but no biggie.

naderman:
@Seldaek fixed the two things you mentioned

Seldaek:
Yay

PR #770, opened 2016-04-15
Fixed $handler property visibility in the HandlerWrapper

andrzejkupczyk:
According to the given example, $handler property should not be private.


Seldaek:
Absolutely right, thanks.

Issue #771, opened 2016-04-15
Invalid assumption about date sorting in RotatingFileHandler


rpkamp:
I was looking at the `RotatingFileHandler` and in lines 111-114 there is the following code:

```php
// Sorting the files by name to remove the older ones
usort($logFiles, function ($a, $b) {
    return strcmp($b, $a);
});
```

The problem with this code is that it makes the assumption that the chronological order of dates is the same as the lexicographical order of the textual representation of those dates, which is indeed the case for the default date format of the class, `Y-m-d`, but is certainly not true for all date formats.
For example here in the Netherlands the default date format is `m-d-Y`. Using that format and assuming we still have some files from the end of last month and some files from the beginning of this month, when rotating the class would keep the files from last month and throw away the files from this month, which is certainly not what you'd want.

Combining this with https://github.com/Seldaek/monolog/issues/753 (which is why I was looking at the file in the first place) it may be worth it to ask people if they actually use the class with a different date format than the default. If not, it would remove a lot of complexity from the class if `Y-m-d` could be set as a hard coded default.

Alternatively, if you don't want to force a default on people, they should be able to supply their own sorting algorithm for the lines given above so they can inject a sorting algorithm for their custom date format.
Inspecting the date format and generating a sorted specifically tailored for that sounds like a framework on its own :)

stof:
The idea of making the format configurable was to allow rotating per month instead of rotating per day for instance. So removing the feature entirely may cause issues.

But I agree that we could provide a simpler feature for that rather than making the format configurable, which would simplify solving #753 too

rpkamp:
Okay, so maybe we could solve it like so:

1) Stop exposing the dateFormat to the outside world (remove parameter from `setFilenameFormat` method), but instead introduce a "rotation strategy" that can be set to "days", "months" or "years", using class constants.

```
const ROTATE_DAYS = 'days';
const ROTATE_MONTHS = 'months';
const ROTATE_YEARS = 'years';
```

2) Add a method `setRotateStrategy` that allows one of the constants above and sets the internal $dateFormat according to the chosen strategy.

3, optional) Make $dateFormat private so people can't override it in subclasses and fall in the same traps described in this ticket.

How does that sound?

NB. The 'years' option is probably overkill

rpkamp:
Come to think of it I think the naming `ROTATE_*` in my earlier comment is quite bad and would instead suggest an interface like

```
[...]

PR #773, opened 2016-04-18
Lock down RotateFileHandler to prevent errors with rotation

rpkamp:
- Require the dateFormat to be one of three presets ('Y-m-d', 'Y-m' or 'Y') to
  ensure that dates can be sorted lexographically
- Require the filenameFormat to contain the (sub)string `{date}` so we will
  actually create new files instead of the same file over and over again.

Fixes https://github.com/Seldaek/monolog/issues/771

This PR breaks BC in a major way so should be set for 2.0
If accepted, the checks added to `setFilenameFormat` could be added to the 1.x branch as well, but raise `E_USER_DEPRECATED` errors instead of throwing `InvalidArgumentException`s as @stof pointed out in https://github.com/Seldaek/monolog/issues/771

### DIFF
--- a/src/Monolog/Handler/RotatingFileHandler.php
+++ b/src/Monolog/Handler/RotatingFileHandler.php
@@ -11,6 +11,7 @@
 
 namespace Monolog\Handler;
 
+use InvalidArgumentException;
 use Monolog\Logger;
 
 /**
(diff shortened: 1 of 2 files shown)


rpkamp:
This PR breaks BC in a major way so should be set for 2.0
If accepted, the checks added to `setFilenameFormat` could be added to the 1.x branch as well, but raise `E_USER_DEPRECATED` errors instead of throwing `InvalidArgumentException`s as @stof pointed out in https://github.com/Seldaek/monolog/issues/771

stof:
Well, I think it would be worth adding this in 1.x first with the deprecation. It will allow warning the users earlier.

Seldaek:
Thanks

PR #774, opened 2016-04-19
Add deprecation errors on RotatingFileHandler

rpkamp:
Add deprecation errors on RotatingFileHandler when attempting to set dateFormats of fileFormats that break the possibility of rotating easily in RotatingFileHandler. Version 2.x
of Monolog will throw `\InvalidArgumentException`s in these cases.

- Require the dateFormat to be one of three presets ('Y-m-d', 'Y-m' or 'Y') to
  ensure that dates can be sorted lexographically
- Require the filenameFormat to contain the (sub)string `{date}` so we will
  actually create new files instead of the same file over and over again.

Related to https://github.com/Seldaek/monolog/pull/773

### DIFF
--- a/src/Monolog/Handler/RotatingFileHandler.php
+++ b/src/Monolog/Handler/RotatingFileHandler.php
@@ -11,7 +11,6 @@
 
 namespace Monolog\Handler;
 
-use InvalidArgumentException;
 use Monolog\Logger;
 
 /**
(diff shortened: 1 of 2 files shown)


rpkamp:
As the stack of commits above proves I haven't worked with PHP 5.3 in quite a while :)
But it's working now.

Seldaek:
Thanks!

PR #775, opened 2016-04-20
Fix NewRelicHandler error when using LineFormatter

michaelmoussa:
My application uses a customized `LineFormatter`, which sets the `formatted` key value to a string. This is a problem for the existing `NewRelicHandler`, which expects `formatted` to be an array with a `context` and `extra` key.

This change just does an `isset` check before looping over those elements.

### DIFF
--- a/src/Monolog/Handler/NewRelicHandler.php
+++ b/src/Monolog/Handler/NewRelicHandler.php
@@ -92,23 +92,27 @@ class NewRelicHandler extends AbstractProcessingHandler
             newrelic_notice_error($record['message']);
         }
 
-        foreach ($record['formatted']['context'] as $key => $parameter) {
-            if (is_array($parameter) && $this->explodeArrays) {
-                foreach ($parameter as $paramKey => $paramValue) {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, merged in 1.x

michaelmoussa:
👍  thank you!

moderndeveloperllc:
@michaelmoussa I definitely didn't think of that situation when I rewrote that bit of code a while back. If `$record['formatted']` is a string, are you sending it to New Relic? I can't see where in your fix you are actually doing any kind of `$this->setNewRelicParameter()` for that parameter.

Seldaek:
Yeah I don't really see the use case as making a ton of sense either, but the guards are harmless.

michaelmoussa:
@moderndeveloperllc I do not need to send any custom parameters to New Relic for this use case, so no, it wouldn't do any `->setNewRelicParameter(...)` calls. I'm just sending plain text log messages that were formatted using a `LineFormatter`.

PR #777, opened 2016-04-25
Add parent logger to 1.x

PhiloEpisteme:
## Overview
This was my first pass as this feature. I have filed a feature Issue ticket to better describe the use case. #778 

In general,developers may find parent loggers useful when a service wants to have a defined logging behavior across all of its parts; however, a developer may also want individual parts of this service to have some additional behavior. For example:

```php
class Foo
{
    protected $logger;

    public function __construct(Logger $parent_logger = null){
        // We want our logger to always log to the injected logger
        // If we have a warning level log, lets do something special with it.
        $special_handler = new MySpecialHandler(Logger::WARN, false);
        // If we don't have a warning level log, lets do something boring.
        $boring_handler = new MyBoringHandler(Logger::DEBUG);

        $logger = new Logger(__CLASS__);
        $logger->setHandlers(array($special_handler, $boring_handler));

        // Lets always log to our parent logger
        $logger->setParent($parent_logger);
        $this->logger = $logger;
    }
}
```
```
$handler_1 = new StreamHandler('/var/log/my_log.log');
$handler_2 = new StreamHandler('/var/log/my_important_log.log', Logger::WARNING, false);
$logger = new Logger('routine_1');
$logger->pushHandler($handler_2);
$logger->pushHandler($handler_1);

// No matter how Foo logs, I always want it to log to my stream handlers from above.
$foo = new Foo($logger);
```

### Possible Additions
One may want to pass the record instead of the message to the parent so that any context or extras added by the processors gets passed to the parent. For example . . .

```php


    /**
     * Adds a log record.
     *
     * @param  int     $level   The logging level
     * @param  string  $message The log message
     * @param  array   $context The log context
     * @return Boolean Whether the record has been processed
     */
    public function addRecord($level, $message, array $context = array(), $message_as_record = false)
    {
        . . .    
        $record = $message_as_record ?
            $message :
            array(
                'message' => (string) $message,
                'context' => $context,
                'level' => $level,
                'level_name' => $levelName,
                'channel' => $this->name,
                'datetime' => $ts,
                'extra' => array(),
        );
        . . . 
        // Lets also propagate this to the parent, if it exists.
        if ($this->parent != null) {
            $this->parent->addRecord($level, $record, $context, true);
            // We return true here because at least one logger (this one) handled this record.
            return true;
        }
        . . . 
    }
```

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -120,6 +120,13 @@ class Logger implements LoggerInterface
     protected $handlers;
 
     /**
+     * The logger's parent logger
+     *
+     * @var Logger
+     */
+    protected $parent;
+
+    /**
      * Processors that will process all log records
      *
      * To process records of a single handler instead, add the processor on that specific handler
@@ -137,12 +144,18 @@ class Logger implements LoggerInterface
      * @param string             $name       The logging channel
      * @param HandlerInterface[] $handlers   Optional stack of handlers, the first one in the array is called first, etc.
      * @param callable[]         $processors Optional array of processors
+     * @param Logger             $parent     Optional logger parent
      */
-    public function __construct($name, array $handlers = array(), array $processors = array())
+    public function __construct(
+        $name,
+        array $handlers = array(),
+        array $processors = array(),
+        Logger $parent = null)
     {
         $this->name = $name;
         $this->handlers = $handlers;
         $this->processors = $processors;
+        $this->parent = $parent;
     }
 
     /**
@@ -217,6 +230,28 @@ class Logger implements LoggerInterface
     public function getHandlers()
     {
         return $this->handlers;
+    }
+
+    /**
+     * Set the parent logger. Will override any prior set parent logger.
+     *
+     * @param Logger $parent    Parent logger
+     * @return $this
+     */
+    public function setParent(Logger $parent)
+    {
+        $this->parent = $parent;
+        return $this;
+    }
+
+    /**
+     * Get the parent logger, if it exists.
+     *
+     * @return Logger
+     */
+    public function getParent()
+    {
+        return $this->parent;
     }
 
     /**
@@ -339,6 +374,12 @@ class Logger implements LoggerInterface
 
             next($this->handlers);
         }
+
+        // Lets also propagate this to the parent, if it exists.
+        if ($this->parent != null) {
+            $this->parent->addRecord($level, $message, $context);
+        }
+
 
         return true;
     }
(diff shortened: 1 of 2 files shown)


rantonmattei:
This is super handy when you want to set up bubble behavior at the logger level.

Something like "_I want all my loggers to send logs to a logging service such as [Loggly](https://www.loggly.com/) or [Logstash](https://www.elastic.co/products/logstash) but then I also want  my children loggers to log locally on a server._" Of course you can use handlers in a similar way. But the major difference is that you can use this in combination with the channel names and start using those as namespaces. i.e. `new Logger('parent.childLogger')`.

Then it gets very powerful because you can you can setup loggers in a parent applications or packages and have them used or extended by other packages using the Registry for instance. I can imagine something like this happening:

In component A, you log using the registry: `Registry::getInstance('hello.com')->info()`. This would go to some log service for instance assuming you set up the right handler for it. Then you use another component B to build your app and work on the `hello.com/signup` feature. Assuming you use namespacing with the `/` as separator you can setup something pretty powerful where anything logged using the `Registry::getInstance('hello.com/signup')` logger gets:

1. sent to your log service for free (using the parent/root feature)
2. then sent to another destination specific to the signup feature.

You can imagine mapping namespaces for various things such as url (like this example), class names, tree structures, or any hierarchy really.
As long as you define parent loggers with handlers, all the inheriting children gets the ability to log to the parent destination automatically. That gives you the ability to separate the action of logging from the actual handlers.

@Seldaek I'm all for it. See above ^^^^

rpkamp:
I have several questions/doubts about this PR.

1) I don't see what this adds in value when compared to the `Monolog\Handler\PsrHandler`. If you create such a `PsrHandler` as the first handler for the "child" logger and pass it the parent logger as you call it, wouldn't that have the exact same effect?
2) The behaviour of `addRecord` with respect to the parent logger differs depending on whether the "child" logger has handlers.
- If it does _not_ have handlers it will _not_ apply the processors before sending the message to the parent
- If it _does_ have handlers it _will_ apply processors before sending the message to the parent.

This kind of behaviour changes are extremely hard to debug if you ever walk in to them.

3) The name `parent` doesn't truly convey the relation between the two loggers to me because it is not a true parent-child relationship. I have a hard time coming up with a name for the relation that does exist between the two classes, but parent-child does not fit imho.

Seldaek:
Sorry but I don't see enough value in this to include it in monolog itself, it can easily be done by extending the Logger class.

PR #782, opened 2016-05-10
fix endless recursion of normalizer formatter

m6w6:
If the NormalizerFormatter encounters an instance of Traversable it
calls itself on all children of that instance. If the Traversable
returns itself (once, sometimes or always -- doesn't matter) the
formatter goes into endless recursion.

Fix by checking the currently being normalized data to not be
identity-equal to its parent container.

### DIFF
--- a/src/Monolog/Formatter/NormalizerFormatter.php
+++ b/src/Monolog/Formatter/NormalizerFormatter.php
@@ -55,7 +55,7 @@ class NormalizerFormatter implements FormatterInterface
         return $records;
     }
 
-    protected function normalize($data)
+    protected function normalize($data, $parent = null)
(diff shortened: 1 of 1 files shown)


staabm:
sounds like a edge case which should be covered by a unit test to prevent regressions

stof:
this tries to handle cyclic object graphs, but only covers the case of cycles of length 2, not of longer cycles.
If we fix it to deal with circular references, it should deal with all of them

m6w6:
Maybe, but there are quite a few standard iterators, which may return itself first.

My actual problem, probably is, that the check for `Traversable` comes before the check for `::__toString`.

PR #783, opened 2016-05-10
fix endless recursion of normalizer formatter

m6w6:
If the NormalizerFormatter encounters an instance of Traversable it
calls itself on all children of that instance. If the Traversable
returns itself (once, sometimes or always -- doesn't matter) the
formatter goes into endless recursion.

Fix by checking the currently being normalized data to not be
identity-equal to its parent container.


staabm:
duplicate of  Seldaek/monolog#782?

m6w6:
Theoretically, yes. Technically, I don't know the merge strategy of this project, so I created a PR against 1.x and another one against master.

PR #785, opened 2016-05-11
Change out Exception to Throwable

moderndeveloperllc:
Looks like these were TODOs for 2.0, so might as well get these out of the way.


staabm:
After this change you need to uplift php version constraint to >=7.0

moderndeveloperllc:
@staabm The `master` branch is already at that.
```
"require": {
    "php": "^7.0",
    "psr/log": "^1.0"
},
```

staabm:
Then its fine :)

Seldaek:
Thanks

PR #788, opened 2016-05-12
Catch Throwable

fprochazka:
I'm not 100% sure about adding the catch everywhere, but here it is, let's have a discussion about it.

Related #790

### DIFF
--- a/src/Monolog/Handler/Handler.php
+++ b/src/Monolog/Handler/Handler.php
@@ -39,7 +39,7 @@ abstract class Handler implements HandlerInterface
     {
         try {
             $this->close();
-        } catch (\Exception $e) {
+        } catch (\Throwable $e) {
             // do nothing
         }
     }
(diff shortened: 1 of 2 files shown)


stof:
makes sense to me

moderndeveloperllc:
With `master` being PHP 7 only (check `composer.json`), you can take out the `\Exception` call and only call `\Throwable`. I don't thing @Seldaek, or anyone else, has gone through the codebase yet to change all these over. I have #785 that does it in another file.

xabbuh:
Though the change as it is right now could make sense in 1.x too.

fprochazka:
@moderndeveloperllc cool, I didn't notice that :+1:

fprochazka:
So, should I also open a PR against `1.x` ?

moderndeveloperllc:
@xabbuh Putting PHP 7 code into 1.x doesn't make much sense as the main reason (I believe) for going to a 2.x release is PHP 7. I'm not even sure the 1.x library works in 7. By bifurcating the code, it eliminates the need to have this kind of duplication or version checking.

@Seldaek, can we get a definitive ruling on putting PHP 7 code in the 1.x branch?

stof:
@moderndeveloperllc the 1.x branch is compatible with PHP 7 yes (it cannot depend on new features added in PHP 7).
Making this handler aimed at hiding errors hide all of them looks sensible to me.

stof:
We would have to catch both Exception and Throwable in 1.x though (which was done in this PR initially)

PR #789, opened 2016-05-13
add is_null check to avoid to set the string "null" instead of null

lenybernard:


src/Monolog/Formatter/GelfMessageFormatter.php
```diff
@@ -106,7 +106,7 @@ public function format(array $record)
         }
 
         foreach ($record['extra'] as $key => $val) {
-            $val = is_scalar($val) ? $val : $this->toJson($val);
+            $val = is_scalar($val) || is_null($val) ? $val : $this->toJson($val);
```
stof: please use ``null === $val`` for consistency

lenybernard: done

Seldaek:
Thanks, merged in 1.x

PR #790, opened 2016-05-13
Catch Throwable

fprochazka:
Related #788


fprochazka:
@Seldaek this shouldn't have been merged into master https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/WhatFailureGroupHandler.php

Seldaek:
Well it had to but now it needs to be cleaned up yeah :)

Seldaek:
I'll do it BTW :)

fprochazka:
@Seldaek Ok, thanks :)

PR #792, opened 2016-05-19
Exclude development stuff from repository autogenerated ZIP archives

mlocati:
People that install this library via composer don't need the test/sample/doc stuff, they just need the library itself.
This PR removes the development directories and files from the repository auto-generated ZIP archives.
Since these development files/directories will still be available via `git clone`, Travis is happy, developers are happy, and so are end-users (they won't have unneeded stuff in production machines).


mlocati:
Thanks!

PR #793, opened 2016-05-20
Datetime changes

### LINKED ISSUE #736: DateTime related changes in 2.0
Things to consider:

- Switch to DateTimeImmutable?
- Extend to make datetimes json serializable (#402) and make sure the format is configurable somehow?
- Disable microsecond precision by default for performance?

Seldaek:
Fixes #736 

- Uses DateTimeImmutable for everything
- Added our own class extending it that means datetimes get serialized in json as `"Y-m-d\TH:i:s.uP"` or `"Y-m-d\TH:i:sP"` depending if microsecond precision is enabled.

Feedback welcome :)


Seldaek:
BTW build fails due to https://github.com/bzikarsky/gelf-php/pull/62

PR #807, opened 2016-06-15
Fix for TypeError within SocketHandler

martinkacmar:
Proposed change fixes `TypeError` thrown while calling [`stream_set_timeout()`](stream_set_timeout) which expects both 2nd and 3rd arguments to be integers while [`floor()`](https://secure.php.net/manual/en/function.floor.php) returns `float` / `bool` and [`round()`](https://secure.php.net/manual/en/function.round.php) returns `float` as well.

**Warning**: This may not be the best solution, I didn't go through internals of class in depth, but I've noticed this breaks Slack logging for us after updating to most recent commit so it may be useful for others too.


Seldaek:
That sounds alright, I guess php returns doubles only because they might be large integer that overflow otherwise. Thanks

PR #808, opened 2016-06-22
Normalization of arrays containing self references

naderman:
This is an issue in particular when $GLOBALS gets stuck into $context. This is something Symfony's ErrorHandler does when an error is thrown in global scope. The result is infinite recusion resulting in stack overflow on normalize().

Currently only adds a failing test.

Options are
- Limit nesting, by introducing depth counting to `normalize()`
- Reference detection (more complicated and uses more memory)

### DIFF
--- a/src/Monolog/Formatter/JsonFormatter.php
+++ b/src/Monolog/Formatter/JsonFormatter.php
@@ -135,8 +135,12 @@ class JsonFormatter extends NormalizerFormatter
      *
      * @return mixed
      */
-    protected function normalize($data)
+    protected function normalize($data, $depth = 0)
     {
+        if ($depth > 9) {
+            return 'Over 9 levels deep, aborting normalization';
+        }
+
         if (is_array($data) || $data instanceof \Traversable) {
             $normalized = array();
 
@@ -146,7 +150,7 @@ class JsonFormatter extends NormalizerFormatter
                     $normalized['...'] = 'Over 1000 items, aborting normalization';
                     break;
                 }
-                $normalized[$key] = $this->normalize($value);
+                $normalized[$key] = $this->normalize($value, $depth+1);
             }
 
             return $normalized;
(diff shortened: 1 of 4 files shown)


naderman:
Limited depth to a hardcoded value of 9, as an implementation of option 1.

Seldaek:
Thanks

src/Monolog/Formatter/JsonFormatter.php
```diff
@@ -135,8 +135,12 @@ protected function formatBatchNewlines(array $records)
      *
      * @return mixed
      */
-    protected function normalize($data)
+    protected function normalize($data, $depth = 0)
```
gmponos: @Seldaek I understand the intention behind this PR reasoning behind although I am not 100% I agree with this.

Regardless what ever point of view anyone might isn't this  a BC?

```
Warning: Declaration of MyClassFormatter::normalize($data) should be compatible with Monolog\Formatter\NormalizerFormatter::normalize($data, $depth = 0)                                                                              
```

Maybe adding it in [changelog](https://github.com/Seldaek/monolog/blob/master/CHANGELOG.md#1240-2018-11-05) with a **BC* suffix?

Seldaek: Well.. if it is just a warning IMO that's not really a BC Break. Do you have your own normalizer extending this? It's an easy enough fix and IMO a rare enough occurrence..

Seldaek: That isn't to say it's not inconvenient.. Sorry I didn't notice this would cause a warning, but at least it's not strictly breaking anything.

gmponos: Yea.. ok.. it is an easy fix.. that's why I suggested it at least add it on readme with instructions :)

Seldaek: Yup, added a notice now at https://github.com/Seldaek/monolog/releases/tag/1.24.0

gmponos: Just wanted to point out that these things could be mentioned to readme file. I don't want to appear ungrateful. You are doing a great job.

Thank you anyway.

PR #811, opened 2016-06-28
Add Logmatic.io provider.

JulienBreux:
Add Logmatic.io provider.

❤️ Monolog

### DIFF
--- a/src/Monolog/Formatter/LogmaticFormatter.php
+++ b/src/Monolog/Formatter/LogmaticFormatter.php
@@ -33,7 +33,7 @@ class LogmaticFormatter extends JsonFormatter
      *
      * @param string $hostname
      */
-    public function setHostname(String $hostname) {
+    public function setHostname(string $hostname) {
       $this->hostname = $hostname;
     }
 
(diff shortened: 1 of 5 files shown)


src/Monolog/Formatter/LogmaticFormatter.php
```diff
@@ -0,0 +1,66 @@
...
+     * @param string
+     */
+    protected $appname = '';
+
+    /**
+     * Set hostname
+     *
+     * @param string $hostname
+     */
+    public function setHostname(String $hostname) {
```
stof: the scalar typehint is ``string``, not ``String``

src/Monolog/Handler/LogmaticHandler.php
```diff
@@ -0,0 +1,89 @@
...
+use Monolog\Logger;
+use Monolog\Formatter\FormatterInterface;
+use Monolog\Formatter\LogmaticFormatter;
+
+/**
+ * @author Julien Breux <julien.breux@gmail.com>
+ */
+class LogmaticHandler extends SocketHandler
+{
+    const LOGMATIC_URI = '/v1/';
```
stof: no need for this constant IMO (and its name is bad anyway, as this is not a URI)

gpolaert:
@JulienBreux It's very helpful. Thanks!

Seldaek:
Thanks

PR #813, opened 2016-06-29
HtmlFormatter methods accessibility

dmitry-ivanov:
Please, make them protected instead of private to enable comfortable overriding of HtmlFormatter class.


Seldaek:
Merged in 1.x, thanks

PR #816, opened 2016-07-06
Added option to log errors regardless of error_reporting setting

masterada:


Seldaek:
Thanks, I merged it in 1.x and master. It's now available in 1.21.0

PR #817, opened 2016-07-07
with graylog2/gelf-php wrong class in tests mock

Hawk6:
"graylog2/gelf-php": "~1.5.0"
Attempted to load class "MessagePublisher" from namespace "Gelf".
Did you forget a "use" statement for another namespace?

### DIFF
--- a/tests/Monolog/Handler/GelfMockMessagePublisher.php
+++ b/tests/Monolog/Handler/GelfMockMessagePublisher.php
@@ -11,7 +11,7 @@
 
 namespace Monolog\Handler;
 
-use Gelf\MessagePublisher;
+use Gelf\Publisher;
 use Gelf\Message;
 
 class GelfMockMessagePublisher extends MessagePublisher


tests/Monolog/Handler/GelfMockMessagePublisher.php
```diff
@@ -11,7 +11,7 @@
 
 namespace Monolog\Handler;
 
-use Gelf\MessagePublisher;
+use Gelf\Publisher;
 use Gelf\Message;
 
 class GelfMockMessagePublisher extends MessagePublisher
```
xabbuh: Then the class name must be changed here too.

xabbuh: and the argument type hint for the `publish()` method must be changed too

xabbuh:
Actually, this change is wrong. The `GelfMockMessagePublisher` class is only used in the `GelfHandlerLegacyTest` which doesn't make use of the `graylog2/gelf-php` package, but is meant to test the legacy `mlehner/gelf-php` package which has different classes.

Hawk6:
ok you're right

PR #821, opened 2016-07-14
Allow alternative date separators for RotatingFileHandler

rpkamp:
Currently only dashes are allowed in date formats (i.e., "Y-m-d").
This commit also allows slashes, underscores and dots to be used instead
of only dashes for more flexibility.


rpkamp:
ping @Seldaek

Seldaek:
Thanks

Seldaek:
It's available in 1.21 now btw.

PR #822, opened 2016-07-18
Fixed incompatible interfaces with MailHandler

ondrejmirtes:
These two classes crash when I try to include them on PHP 7.

BTW: I found this issue using my new static analysis tool [PHPStan](https://github.com/phpstan/phpstan). I can send you a PR with integrating the tool in the Monolog build script.


Seldaek:
Thanks but I usually use etsy's phan, the master branch is just kind of in flux at the moment so not paying too much attention to those minor glitches yet. Anyway thanks for the fix :)

PR #824, opened 2016-07-20
Throw RuntimeException if socket resource is missing

adambro:
The socket resource can be closed, purged or lost during the runtime of an application.
It is valid concern, because the resource is acquired and used at different times during request.

LogicException is more for static preconditions that have to be valid before function call.

I realize that it's BC break, but it still makes sense. Having that I can handle all RuntimeExceptions the same way, instead of using catch-all Exception handling.

### DIFF
--- a/src/Monolog/Handler/SyslogUdp/UdpSocket.php
+++ b/src/Monolog/Handler/SyslogUdp/UdpSocket.php
@@ -42,7 +42,7 @@ class UdpSocket
     protected function send($chunk)
     {
         if (!is_resource($this->socket)) {
-            throw new \LogicException('The UdpSocket to '.$this->ip.':'.$this->port.' has been closed and can not be written to anymore');
+            throw new \RuntimeException('The UdpSocket to '.$this->ip.':'.$this->port.' has been closed and can not be written to anymore');
         }
         socket_sendto($this->socket, $chunk, strlen($chunk), $flags = 0, $this->ip, $this->port);
     }


staabm:
The RuntimeExcepion feels more apropriate as the LogicException (as there is no logical error, just the runtime assumption that the socket will work is no longer valid).

In case you wanne handle this case in a dedicated try-catch block you could just add a `SocketGoneException extends RuntimeException` and throw this instead (which would also preserve BC)

adambro:
@staabm I know I could extend exception, but it's more about the semantics of those exceptions. Socker resource can be gone at any time in runtime, hence relevant exception. Here's nice explanation of Logic vs Runtime exception:
http://stackoverflow.com/a/5586510

staabm:
I agree that RuntimeException is a better fit then LogicException (as I wrote above).

I just proposed the BC version of this patch as IMO in a minor release such types shouldn't change.

adambro:
Agree, definitely BC change not for minor version release. I've noticed @Seldaek is working on major release, because I see a lot of BC breaks in master (including PHP7 requirement). Maybe it's the time to re-think the structure and semantics of exceptions in handlers?

staabm:
> Agree, definitely BC change not for minor version release. I've noticed @Seldaek is working on major release, because I see a lot of BC breaks in master (including PHP7 requirement). Maybe it's the time to re-think the structure and semantics of exceptions in handlers?

good call. lets wait for jordi.

adambro:
Similar kind of problem is in StreamHandler. In this case LogicException makes sense, but it should be rather thrown at initialization step (i.e. constructor) not during writing:
https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/StreamHandler.php#L94

Seldaek:
LogicException was used there IMO as a UDP socket shouldn't just stop working out of the blue AFAIK, as it's not really an established connection but rather a fire and forget protocol.. Maybe that's a wrong assumption though, in which case I'm ok changing it for 2.0 to RuntimeException

Regarding StreamHandler, it's not done in constructor because in most requests nothing is logged if you only log errors, so doing i/o checks for the stream validity would be prohibitive in performance terms.

Seldaek:
Ping? :) I'd be happy to get a definite answer to:

> LogicException was used there IMO as a UDP socket shouldn't just stop working out of the blue AFAIK, as it's not really an established connection but rather a fire and forget protocol.. Maybe that's a wrong assumption though, in which case I'm ok changing it for 2.0 to RuntimeException

Then we can move forward.

adambro:
True, UDP is fire&forget protocol. Here we're checking for PHP resource rather than if the resource is UDP socket or can make UDP connections. If I understand the check correctly it is just to have nice, exception based, error handling, rather than bubble internal PHP error from `socket_sendto()` call (not a resource or something similar).

Idea: how about having public API to check if handler can perform `write()` operation? Such API would allow for handler user to make verification at init or runtime - whatever fits them better. It would be also nice framework for handlers contributors IMHO. Currently the code is a bit unreadable, i.e. take StreamHandler class.

Seldaek:
OK I think it's not a really big deal either way but it sounds alright so merging :)

PR #826, opened 2016-07-22
Throw exception in case the log file has been deleted

jjanvier:
When you create a StreamHandler based on a file, absolutely no errors is raised if the file has been deleted.

Which is a real problem, as you don't know that you are not able to log.

This PR aims to fix this bug.

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -108,6 +108,10 @@ class StreamHandler extends AbstractProcessingHandler
             }
         }
 
+        if (null !== $this->url && null !== $this->getDirFromStream($this->url) && !is_writable($this->url)) {
+            throw new \UnexpectedValueException(sprintf('Impossible to write in the stream. This may be caused by the deletion of "%s".', $this->url));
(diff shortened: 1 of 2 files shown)


Seldaek:
The problem here is that this fix basically adds i/o ops to *every* log write to detect a rare edge case. Unfortunately IIRC it's not really feasible to reliably check if fwrite() succeeded or not, and only act on failures. PHP is really not helping us here :/ But as it stands I'd rather not merge this for performance reasons. If it really matters to you though you could do it by extending the class and overriding `write()`.

jjanvier:
@Seldaek yes you're right, PHP is not helping at all. `fwrite` does not return false even if the resource's file does not exist anymore.

I've been looking for a more elegant solution but I didn't find anything. Maybe somebody will have an idea.

Otherwise, would it make sense to create a  dedicated `FileStreamHandler` in monolog? Maybe it has a big impact for the lib, I don't know, it's up to you.

Seldaek:
I don't know if it's really worth having a specialized handler for that, but you can always do one and publish it as its own package :)

Issue #835, opened 2016-08-27
JsonFormatter::normalize() should pass \Throwable to ::normalizeException() on PHP7


eimanavicius:
JsonFormatter::normalizeException() handles \Throwable:
https://github.com/Seldaek/monolog/blob/1.x/src/Monolog/Formatter/JsonFormatter.php#L173

but JsonFormatter::normalize() do not passes it to normalizeException():
https://github.com/Seldaek/monolog/blob/1.x/src/Monolog/Formatter/JsonFormatter.php#L155

eimanavicius:
same in master:
https://github.com/Seldaek/monolog/blob/master/src/Monolog/Formatter/JsonFormatter.php#L159

PR #836, opened 2016-08-27
Better support for PHP7 \Throwable

eimanavicius:
#835 Better support for PHP7 \Throwable in JsonFormatter


Seldaek:
Thanks, I merged this in 1.x and then into master so will close the other PR (#837).

PR #843, opened 2016-09-02
Add ConstStub awareness to NormalizerFormatter

nicolas-grekas:
By wrapping context values into [`ConstStub`](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/VarDumper/Caster/ConstStub.php) objects, one can add precise type information to log contexts. Yet, this results in pure noise when the context is dumped into log files.
Since ConstStub is basically a value wrapper, NormalizerFormatter can remove it to get clean logs back.

This would be of paramount usefulness for the upcoming Symfony 3.2 profiler panel where we'd use this feature immediately to get a better log panel.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "php-amqplib/php-amqplib": "~2.4",
         "swiftmailer/swiftmailer": "~5.3",
         "php-console/php-console": "^3.1.3",
+        "symfony/var-dumper": ">=2.7",
         "jakub-onderka/php-parallel-lint": "^0.9"
     },
     "suggest": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -26,6 +26,7 @@
         "php-amqplib/php-amqplib": "~2.4",
         "swiftmailer/swiftmailer": "~5.3",
         "php-console/php-console": "^3.1.3",
+        "symfony/var-dumper": ">=2.7",
```
nicolas-grekas: 3.x & 4.x is guaranteed by Symfony BC policy regarding major versions and by the fact that a simple `instanceof` is used.

nicolas-grekas:
Closing as I'm taking another approach in the profiler. This might still be useful, but let's have someone actually need it.

Issue #831, opened 2016-07-29
JsonFormatter (json_encode) should properly encode float


gpolaert:
Hi everyone,

When I use JsonFormatter, float are not properly encoded. 
For instance, 

```php
<?php

$z = [
  'stage1' => [
    'stage2.1'              =>  'Alexia',
    'stage2.2'    => [
      'cpl' =>  5.00
   ]
  ]
];

$a = json_encode($z);
$b = json_encode($z, JSON_NUMERIC_CHECK);
$c = json_encode($z, JSON_NUMERIC_CHECK | JSON_PRESERVE_ZERO_FRACTION);

var_dump($a);//string(53) "{"stage1":{"stage2.1":"Alexia","stage2.2":{"cpl":5}}}"
var_dump($b); //string(53) "{"stage1":{"stage2.1":"Alexia","stage2.2":{"cpl":5}}}"
var_dump($c); //string(55) "{"stage1":{"stage2.1":"Alexia","stage2.2":{"cpl":5.0}}}"




?>
```
IMO, I think the correct json string is `$c`.
What do you think?

PR #844, opened 2016-09-05
Preserved typing for float and integers values

gpolaert:
Related #831
Logs stored in a ELK-based solution need to keep the value typing, especially with float vs integer types.
Adding `JSON_NUMERIC_CHECK` and `JSON_PRESERVE_ZERO_FRACTION` allow to avoid issue.

### DIFF
--- a/src/Monolog/Formatter/NormalizerFormatter.php
+++ b/src/Monolog/Formatter/NormalizerFormatter.php
@@ -196,7 +196,7 @@ class NormalizerFormatter implements FormatterInterface
      */
     private function jsonEncode($data)
     {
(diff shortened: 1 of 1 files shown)


Seldaek:
Hey, thanks for the PR. I am fine with JSON_PRESERVE_ZERO_FRACTION as it is fairly harmless and preserves the float/int types, but JSON_NUMERIC_CHECK has potential nasty side effects as is_numeric is a bit trigger happy, things like `1e3`, `+4179000000` like phone numbers, probably octals too, will just be blindly converted in many cases I can see this happening accidentally. I'll merge but without the NUMERIC_CHECK

Issue #743, opened 2016-03-14
Getting token to use Monolog with Slack


nisbeti:
It is not straight forward to get a token to use monolog with Slack. The expectation is that you write an OAuth client to get the token, but it is possible to get it manually. These are the steps I followed today:

You should already have an account and a channel set up that you want to send error messages to.
1. Create a new Slack App - https://api.slack.com/applications/new which will give you a client_id and a client_secret which you need for step two. You will need to enter a redirect URI, which can be any page for this purpose.
2. Got to https://slack.com/oauth/authorize?client_id=[client_id]&scope=chat:write:bot
3. Authorise App on the page that is returned.
4. From the URI copy the code that is returned.
5. Go to https://slack.com/api/oauth.access?client_id=[client_id]&client_secret=[client_secret]&code=[code]
6. Your token is in the JSON that is displayed.

TO debug errors insert echo fread($this->resource, 1000); at the end of the writeToSocket method of Monolog/Handler/SocketHandler (see https://github.com/Seldaek/monolog/issues/514)

johnhunt:
This method doesn't work for me, I just get the following json response from oauth.access:

```
{
"ok": false,
"error": "invalid_code"
}
```

Wouldn't it make more sense to just use the incoming webhook functionality? This is far easier to set up than having to do a manual oauth dance..

damienalexandre:
It worked for me. To bad this handler is not using a simple but effective web-hook, it would be much simpler to configure.

jsafield:
omg. 
I don't want to open the page; https://nimbustest.slack.com/oauth/reflow/ and return back after clicking Authorization.
 https://slack.com/oauth/authorize?client_id=[client_id]&scope=chat:write:bot
Is it only way to get code?? How can I get code in the server?

Gummibeer:
I'm working with a test token https://api.slack.com/docs/oauth-test-tokens but a web-hook would be much better and simpler! :)

hkdobrev:
I also think that this is not about documentation. The Slack handler should use incoming webhooks. They are made for things like that and are quite simpler to configure. The Slack OAuth apps get published and you describe how other users could install them. I'm sure almost all cases for the Monolog Slack handler are about logging something to your team Slack.

[...]

Issue #745, opened 2016-03-16
SlackHandler ignores includeStacktraces


websirnik:
I'm using MonologBundle to include stacktraces in logs. It works for StreamHandler, but not for Slack.

``` yaml
monolog:
    handlers:
        main:
            type:  stream
            path:  %kernel.logs_dir%/%kernel.environment%.log
            level: error
            include_stacktraces: true
        slack:
            type: slack
            token: xxx
            channel: "#errors"
            bot_name: ErrorBot
            icon_emoji: :ghost:
            level: critical
            include_stacktraces: true
```

Seldaek:
Yeah, the problem seems to be SlackHandler uses LineFormatter::stringify which doesn't handle exceptions, and it also doesn't use this->getFormatter() like the include_stacktrace does.. lots of small problems which make it not so easy to fix, but it sure isn't impossible :)

LeniM:
I think you can do it with a formater.
In my case I just used a global handler that does $e->__toString() before sending it to slack. works fine

gmponos:
Hi @websirnik does the problem still occur?

Can you check this #1127 and tell me if this happens also for you?

Seldaek:
Closing as no activity here in a while.

PR #846, opened 2016-09-06
Add new Slack handlers using Slackbot and webhooks

hkdobrev:
See https://github.com/Seldaek/monolog/issues/743#issuecomment-242994019

This is keeping the previous `SlackHandler`, but extracts the message formatting logic and introduces two new Slack handlers:
- `SlackbotHandler` - this is the simplest way to log to Slack and most people could use that. It doesn't allow fancy formatting, but has autolinking and the simplest configuration with two clicks in the Slack settings and getting a token.
- `SlackWebhookHandler` - This is a bit more advanced as it allows for the same formatting and adding attachements as the `SlackHandler`, but it uses the Slack Webhooks which are quite easier to set up than the OAuth API. The minimum it requires is a webhook URL which is a retrievable with about two clicks in the Slack settings. Even the channel is optional as it could be configured in the integration.

This is probably not the leanest way to implement what we've discussed in https://github.com/Seldaek/monolog/issues/743, but after working at this I believe this is the simplest and the right approach.

While we could implement everything in one handler the API would become horrible and it would be harder to understand from users. Providing choice with simple to use separate classes is better IMHO. Apart from this the `SlackHandler` extends the `SocketHandler` and the webhooks and the Slackbot wouldn't work nicely with that.

What do you think?

@Seldaek @gkedzierski @Gummibeer @nisbeti

---
- [x] Extracting Slack formatting logic in `SlackRecord`
- [x] `SlackbotHandler`
- [x] `SlackWebhookHandler`
- [x] Refactor `SlackHandler` using the new `SlackRecord`
- [x] Add new handlers to the docs
- [x] Unit tests of `SlackRecord`
- [x] Unit tests for `SlackbotHandler`
- [x] Unit tests for `SlackWebhookHandler`
- [x] Update unit tests for `SlackHandler`

### DIFF
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -99,7 +99,6 @@ class SlackRecord
         $dataArray = array(
             'username'    => $this->username,
             'text'        => '',
-            'attachments' => array(),
         );
 
         if ($this->channel) {
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -179,7 +179,7 @@ class SlackRecord
                 }
             }
 
-            $dataArray['attachments'] = json_encode(array($attachment));
+            $dataArray['attachments'] = array($attachment);
         } else {
             $dataArray['text'] = $message;
         }
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -96,11 +96,14 @@ class SlackRecord
     public function getSlackData(array $record)
     {
         $dataArray = array(
-            'channel'     => $this->channel,
             'username'    => $this->username,
             'text'        => '',
             'attachments' => array(),
         );
+
+        if ($this->channel) {
+            $dataArray['channel'] = $this->channel;
+        }
 
         if ($this->formatter) {
             $message = $this->formatter->format($record);
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -216,10 +216,14 @@ class SlackRecord
      * Stringifies an array of key/value pairs to be used in attachment fields
      *
      * @param  array  $fields
-     * @return string
+     * @return string|null
      */
     public function stringify($fields)
     {
+        if (!$this->lineFormatter) {
+            return null;
+        }
+
         $string = '';
         foreach ($fields as $var => $val) {
             $string .= $var.': '.$this->lineFormatter->stringify($val)." | ";
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -217,9 +217,9 @@ class SlackRecord
     /**
      * Generates attachment field
      *
-     * @param $title
-     * @param $value
-     * @param $short
+     * @param string $title
+     * @param string|array $value
+     * @param bool $short
      * @return array
      */
     private function generateAttachmentField($title, $value, $short)
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -226,4 +226,14 @@ class SlackRecord
 
         return $string;
     }
+
+    /**
+     * Sets the formatter
+     *
+     * @param FormatterInterface $formatter
+     */
+    public function setFormatter(FormatterInterface $formatter)
+    {
+        $this->formatter = $formatter;
+    }
 }
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -96,7 +96,7 @@ class SlackRecord
         $this->includeContextAndExtra = $includeContextAndExtra;
         $this->formatter = $formatter;
 
-        if ($this->includeContextAndExtra && $this->useShortAttachment) {
+        if ($this->includeContextAndExtra) {
             $this->lineFormatter = new LineFormatter();
         }
     }
@@ -152,7 +152,7 @@ class SlackRecord
                         foreach ($record['extra'] as $var => $val) {
                             $attachment['fields'][] = array(
                                 'title' => $var,
-                                'value' => $val,
+                                'value' => is_array($val) ? $this->lineFormatter->stringify($val) : $val,
                                 'short' => $this->useShortAttachment,
                             );
                         }
@@ -171,7 +171,7 @@ class SlackRecord
                         foreach ($record['context'] as $var => $val) {
                             $attachment['fields'][] = array(
                                 'title' => $var,
-                                'value' => $val,
+                                'value' => is_array($val) ? $this->lineFormatter->stringify($val) : $val,
                                 'short' => $this->useShortAttachment,
                             );
                         }
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -25,6 +25,14 @@ use Monolog\Formatter\FormatterInterface;
  */
 class SlackRecord
 {
+    const COLOR_DANGER = 'danger';
+
+    const COLOR_WARNING = 'warning';
+
+    const COLOR_GOOD = 'good';
+
+    const COLOR_DEFAULT = '#e3e4e6';
+
     /**
      * Slack channel (encoded ID or name)
      * @var string
@@ -194,13 +202,13 @@ class SlackRecord
     {
         switch (true) {
             case $level >= Logger::ERROR:
-                return 'danger';
+                return self::COLOR_DANGER;
             case $level >= Logger::WARNING:
-                return 'warning';
+                return self::COLOR_WARNING;
             case $level >= Logger::INFO:
-                return 'good';
+                return self::COLOR_GOOD;
             default:
-                return '#e3e4e6';
+                return self::COLOR_DEFAULT;
         }
     }
 
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -1,0 +1,221 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler\Slack;
+
+use Monolog\Logger;
+use Monolog\Formatter\LineFormatter;
+use Monolog\Formatter\FormatterInterface;
+
+/**
+ * Slack record utility helping to log to Slack webhooks or API.
+ *
+ * @author Greg Kedzierski <greg@gregkedzierski.com>
+ * @author Haralan Dobrev <hkdobrev@gmail.com>
+ * @see    https://api.slack.com/incoming-webhooks
+ * @see    https://api.slack.com/docs/message-attachments
+ */
+class SlackRecord
+{
+    /**
+     * Slack channel (encoded ID or name)
+     * @var string
+     */
+    private $channel;
+
+    /**
+     * Name of a bot
+     * @var string
+     */
+    private $username;
+
+    /**
+     * Emoji icon name
+     * @var string
+     */
+    private $iconEmoji;
+
+    /**
+     * Whether the message should be added to Slack as attachment (plain text otherwise)
+     * @var bool
+     */
+    private $useAttachment;
+
+    /**
+     * Whether the the context/extra messages added to Slack as attachments are in a short style
+     * @var bool
+     */
+    private $useShortAttachment;
+
+    /**
+     * Whether the attachment should include context and extra data
+     * @var bool
+     */
+    private $includeContextAndExtra;
+
+    /**
+     * @var FormatterInterface
+     */
+    private $formatter;
+
+    /**
+     * @var LineFormatter
+     */
+    private $lineFormatter;
+
+    public function __construct(
+        $channel,
+        $username = 'Monolog',
+        $useAttachment = true,
+        $iconEmoji = null,
+        $useShortAttachment = false,
+        $includeContextAndExtra = false,
+        FormatterInterface $formatter = null
+    ) {
+        $this->channel = $channel;
+        $this->username = $username;
+        $this->iconEmoji = trim($iconEmoji, ':');
+        $this->useAttachment = $useAttachment;
+        $this->useShortAttachment = $useShortAttachment;
+        $this->includeContextAndExtra = $includeContextAndExtra;
+        $this->formatter = $formatter;
+
+        if ($this->includeContextAndExtra && $this->useShortAttachment) {
+            $this->lineFormatter = new LineFormatter();
+        }
+    }
+
+    public function getSlackData(array $record)
+    {
+        $dataArray = array(
+            'channel'     => $this->channel,
+            'username'    => $this->username,
+            'text'        => '',
+            'attachments' => array(),
+        );
+
+        if ($this->formatter) {
+            $message = $this->formatter->format($record);
+        } else {
+            $message = $record['message'];
+        }
+
+        if ($this->useAttachment) {
+            $attachment = array(
+                'fallback' => $message,
+                'color'    => $this->getAttachmentColor($record['level']),
+                'fields'   => array(),
+            );
+
+            if ($this->useShortAttachment) {
+                $attachment['title'] = $record['level_name'];
+                $attachment['text'] = $message;
+            } else {
+                $attachment['title'] = 'Message';
+                $attachment['text'] = $message;
+                $attachment['fields'][] = array(
+                    'title' => 'Level',
+                    'value' => $record['level_name'],
+                    'short' => true,
+                );
+            }
+
+            if ($this->includeContextAndExtra) {
+                if (!empty($record['extra'])) {
+                    if ($this->useShortAttachment) {
+                        $attachment['fields'][] = array(
+                            'title' => "Extra",
+                            'value' => $this->stringify($record['extra']),
+                            'short' => $this->useShortAttachment,
+                        );
+                    } else {
+                        // Add all extra fields as individual fields in attachment
+                        foreach ($record['extra'] as $var => $val) {
+                            $attachment['fields'][] = array(
+                                'title' => $var,
+                                'value' => $val,
+                                'short' => $this->useShortAttachment,
+                            );
+                        }
+                    }
+                }
+
+                if (!empty($record['context'])) {
+                    if ($this->useShortAttachment) {
+                        $attachment['fields'][] = array(
+                            'title' => "Context",
+                            'value' => $this->stringify($record['context']),
+                            'short' => $this->useShortAttachment,
+                        );
+                    } else {
+                        // Add all context fields as individual fields in attachment
+                        foreach ($record['context'] as $var => $val) {
+                            $attachment['fields'][] = array(
+                                'title' => $var,
+                                'value' => $val,
+                                'short' => $this->useShortAttachment,
+                            );
+                        }
+                    }
+                }
+            }
+
+            $dataArray['attachments'] = json_encode(array($attachment));
+        } else {
+            $dataArray['text'] = $message;
+        }
+
+        if ($this->iconEmoji) {
+            $dataArray['icon_emoji'] = ":{$this->iconEmoji}:";
+        }
+
+        return $dataArray;
+    }
+
+    /**
+     * Returned a Slack message attachment color associated with
+     * provided level.
+     *
+     * @param  int    $level
+     * @return string
+     */
+    public function getAttachmentColor($level)
+    {
+        switch (true) {
+            case $level >= Logger::ERROR:
+                return 'danger';
+            case $level >= Logger::WARNING:
+                return 'warning';
+            case $level >= Logger::INFO:
+                return 'good';
+            default:
+                return '#e3e4e6';
+        }
+    }
+
+    /**
+     * Stringifies an array of key/value pairs to be used in attachment fields
+     *
+     * @param  array  $fields
+     * @return string
+     */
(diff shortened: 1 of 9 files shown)


Gummibeer:
I will Test it tomorrow and give you Feedback.

Seldaek:
This seems fine to me implementation wise, if you can write tests for SlackRecord at least that'd be nice, testing the curl calls isn't really easy nor worth it IMO.

I'd also be happy to get feedback from people actually using this.. :)

Seldaek:
BTW it might be worth looking at this at the same time if the Slack stuff gets a rewrite.. https://github.com/Seldaek/monolog/issues/745

And also marking the original SlackHandler as deprecated would be good I think (I will remove it from master/2.0 then), or do you think there is really a point in having 3?

greeny:
I know that this might be a case for all handlers, not just this one, but what if I have disabled curl extension? There is also way to send requests through `file_get_contents` (you can check my implementation in https://github.com/greeny/nette-slack-logger/blob/master/src/SlackLogger.php#L77). Are we going to support this?

hkdobrev:
@Seldaek I've started working on the tests and while doing so I've deprecated a couple of protected methods in `SlackHandler` which are better accessed through `SlackRecord`. However I don't see why we'd deprecate the `SlackHandler`. It's the only handler using the Slack API directly.

@greeny Could you point me to Monolog handlers which employ either `cURL` and if not available fall back to `file_get_contents()`? I don't think using only `file_put_contents()` is nice. I guess we could use Guzzle instead which would do the check and the fallback for us.

greeny:
@hkdobrev I did not check monolog for handlers, but I think many of them are firing requests through curl. Guzzle is fine solution, I was just asking, if servers without curl can use this handler too.

<offtopic>Regarding my implementation, that was first only internal service which i published later.</offtopic>

Seldaek:
Most handlers use curl or raw sockets yes. I think that's fine to be honest, it's the first time anyone even asks about it in all these years. I wouldn't add a dependency on guzzle just for that.

greeny:
@Seldaek it is not mine case (i have curl everywhere), but I know some hostings, that have curl disabled. Thats why I have `file_get_contents` in my implementation.

But I agree curl is very common nowdays.

@hkdobrev what about support for non-emoji icons? (`"icon_url": "https://slack.com/img/icons/app-57.png"`)

src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -0,0 +1,224 @@
...
+                );
+            }
+
+            if ($this->includeContextAndExtra) {
+                if (!empty($record['extra'])) {
+                    if ($this->useShortAttachment) {
+                        $attachment['fields'][] = array(
+                            'title' => "Extra",
+                            'value' => $this->stringify($record['extra']),
+                            'short' => $this->useShortAttachment,
```
stof: could be ``true`` here, as we know the value. It may be more readable

src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -0,0 +1,224 @@
...
+                        );
+                    } else {
+                        // Add all extra fields as individual fields in attachment
+                        foreach ($record['extra'] as $var => $val) {
+                            $attachment['fields'][] = array(
+                                'title' => $var,
+                                'value' => $val,
+                                'short' => $this->useShortAttachment,
```
stof: could be ``false`` directly

src/Monolog/Handler/SlackHandler.php
```diff
@@ -81,25 +46,36 @@ class SlackHandler extends SocketHandler
...
+        $iconEmoji = null,
+        $level = Logger::CRITICAL,
+        $bubble = true,
+        $useShortAttachment = false,
+        $includeContextAndExtra = false
+    ) {
         if (!extension_loaded('openssl')) {
-            throw new MissingExtensionException('The OpenSSL PHP extension is required to use the SlackHandler');
+            throw new MissingExtensionException(
+                'The OpenSSL PHP extension is required to use the SlackHandler'
```
stof: this change should be reverted

src/Monolog/Handler/SlackHandler.php
```diff
@@ -81,25 +46,36 @@ class SlackHandler extends SocketHandler
...
+    public function __construct(
+        $token,
+        $channel,
+        $username = 'Monolog',
+        $useAttachment = true,
+        $iconEmoji = null,
+        $level = Logger::CRITICAL,
+        $bubble = true,
+        $useShortAttachment = false,
+        $includeContextAndExtra = false
```
stof: Monolog always has the signature on one line

src/Monolog/Handler/SlackbotHandler.php
```diff
@@ -0,0 +1,84 @@
...
+    private $token;
+
+    /**
+     * Slack channel name
+     * @var string
+     */
+    private $channel;
+
+    /**
+     * @param  string $token   Slackbot token
```
stof: missing phpdoc for ``$slackTeam``

src/Monolog/Handler/SlackWebhookHandler.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Instance of the SlackRecord util class preparing data for Slack API.
+     * @var SlackRecord
+     */
+    private $slackRecord;
+
+    /**
+     * @param  string      $webhookUrl             Slack Webhook URL
+     * @param  string      $channel                Slack channel (encoded ID or name)
```
stof: ``string|null``

src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -0,0 +1,224 @@
...
+                            );
+                        }
+                    }
+                }
+            }
+
+            $dataArray['attachments'] = json_encode(array($attachment));
```
martijncalker: Not sure if I use it the right way, but now I've got:
```json
{"username":"TestBot","text":"","attachments":"[{\\"fallback\\":\\"Testing Slackhandler!\\",\\"color\\":\\"good\\",\\"fields\\":[{\\"title\\":\\"Level\\",\\"value\\":\\"INFO\\",\\"short\\":true}],\\"title\\":\\"Message\\",\\"text\\":\\"Testing Slackhandler!\\"}]","channel":"development-logging"}
```

Changing this line to:
```php
$dataArray['attachments'] = [$attachment];
```
Output:
```json
{"username":"TestBot","text":"","attachments":[{"fallback":"Testing Slackhandler!","color":"good","fields":[{"title":"Level","value":"INFO","short":true}],"title":"Message","text":"Testing Slackhandler!"}],"channel":"development-logging"}
```

an1zhegorodov:
@Seldaek 
> I'd also be happy to get feedback from people actually using this.. :)

Happy to see webhooks implementation, because it's very simple to configure.

Using webhooks and curl as a temporary solution. Tried using current SlackHandler, but it silently failed to send messages. The weird thing though, it worked like a charm when I was xdebugging. Could be env issue on my side, anyone faced this behavior?

Seldaek:
@hkdobrev any update here? Should I wait a couple days to get a chance to include this in the next 1.x release or will you need more time to finalize it? No stress I just need to know :)

an1zhegorodov:
@Seldaek Wait couple of days pls. I'm using custom handler on my project, but would love this implementation. I've just finished tests for SlackRecord, planning to fix tests for SlackHandler and fix thing related to @stof comments.

an1zhegorodov:
@Seldaek [done](https://github.com/hkdobrev/monolog/pull/1). Now waiting for feedback from @hkdobrev

hkdobrev:
@an1zhegorodov Thank you for your work! `SlackRecord` wasn't really my code as I just extracted it from the previous `SlackHandler` and I wasn't really keen on refactoring it. I have now cherry-picked your commits and fixing conflicts in the meantime as I had some things locally which I haven't pushed. For example I've moved colours to constants, deprecated old methods and exposed the `SlackRecord` instance instead. I've started with a couple of tests and I've merged our `SlackRecordTest`.

Could you please confirm that your changes didn't incur any backwards incompatible changes? Not just to the PHP API, but also to the way the pre-existing `SlackHandler` is configured?

src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -115,59 +115,36 @@ public function getSlackData(array $record)
...
-                $attachment['text'] = $message;
             } else {
                 $attachment['title'] = 'Message';
-                $attachment['text'] = $message;
-                $attachment['fields'][] = array(
-                    'title' => 'Level',
-                    'value' => $record['level_name'],
-                    'short' => true,
-                );
+                $attachment['fields'][] = $this->generateAttachmentField('Level', $record['level_name'], true);
```
hkdobrev: I'm sorry for commenting on an outdated diff, will copy my comment on the updated line as it is still valid and so the comment could be visible.

src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -0,0 +1,249 @@
...
+                'text'     => $message,
+                'color'    => $this->getAttachmentColor($record['level']),
+                'fields'   => array(),
+            );
+
+            if ($this->useShortAttachment) {
+                $attachment['title'] = $record['level_name'];
+            } else {
+                $attachment['title'] = 'Message';
+                $attachment['fields'][] = $this->generateAttachmentField('Level', $record['level_name'], true);
```
hkdobrev: @an1zhegorodov I think here in the `else` of `if ($this->useShortAttachment)` we should really pass `false` for the short attachment option instead of `true`, don't you think? Just like @stof's comment: https://github.com/Seldaek/monolog/pull/846#discussion_r83167917

@stof While I agree with your comment and would have changed it without thinking twice, this is an example where not using the property introduces a bug more easily.

hkdobrev:
@greeny 

> @hkdobrev what about support for non-emoji icons? (`"icon_url": "https://slack.com/img/icons/app-57.png"`)

This is a good suggestion. However, I wanted to limit the scope of this PR to provide the current functionality, but through the Slackbot or Webhooks without going through the Slack API. I think it should be extended to support everything Slack supports. But this could be done in a later Monolog release, ideally in 2.x when it could be refactored with breaking the public interface of the handler and its configuration so it could have a more flexible way to support everything Slack offers and even future features of Slack webhooks without changing the API and addition additional arguments to methods.

What do you think? Do you think this should be done as part of this PR? Cheers!

hkdobrev:
@Seldaek 

> And also marking the original `SlackHandler` as deprecated would be good I think (I will remove it from master/2.0 then), or do you think there is really a point in having 3?

I've marked a couple of methods as deprecated now. However, I disagree with marking the whole handler as deprecated as people who already use the Slack API could have a valid case and could continue to use it. Using the Slack API is not deprecated, so I don't see why we should deprecated the handler. Although, 2.x would be a good place to refactor the public interface and allow for more flexibility so the interface doesn't depend on every property Slack supports, I don't see why to deprecate the whole method of logging to the Slack API.

What do you think? Are you still considering removing the whole handler in 2.x?

src/Monolog/Handler/Slack/SlackRecord.php
hkdobrev: @martijncalker Thanks for your feedback! This is now fixed thanks to the work of @an1zhegorodov

hkdobrev:
Please let me know if you'd like commits squashed or partially squashed.

Thanks to all for all the feedback and help! I'm sorry for delaying it!

src/Monolog/Handler/Slack/SlackRecord.php
an1zhegorodov: @hkdobrev It's not about `useShortAttachment` in this case. Check out [current](https://github.com/Seldaek/monolog/blob/0fb49dcfdf7318c0bb01bf6876a44bac0a4a99d1/src/Monolog/Handler/SlackHandler.php#L170) implementation. It also has `'short' => true`. That's because `level_name` is really short: `INFO`, `DEBUG`, `ERROR`.

hkdobrev: @an1zhegorodov Thanks for pointing this out! I got confused from the if-else and for being away from this for a while.

an1zhegorodov:
> Could you please confirm that your changes didn't incur any backwards incompatible changes? Not just to the PHP API, but also to the way the pre-existing SlackHandler is configured?

@hkdobrev I didn't change public API, only made `SlackHandler` constructor one line. Any hints what I should look for? I think it's worth to try it out locally on `1.x`, what do you think?

src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -0,0 +1,249 @@
...
+                        // Add all extra fields as individual fields in attachment
+                        $attachment['fields'] = array_merge(
+                            $attachment['fields'],
+                            $this->generateAttachmentFields($record[$key])
+                        );
+                    }
+                }
+            }
+
+            $dataArray['attachments'] = array($attachment);
```
an1zhegorodov: Removed `json_encode` here. I think this will break `SlackHandler`, despite working fine with `SlackWebhookHandler`. `SlackHandler` sends data with `application/x-www-form-urlencoded` headers, and it looks like slack expects `attachments` field to be json_encoded in that case.

A few optoins I figured:
- json_encode `attachments` in `SlackHandler`
- change content-type to `application/json` and json_encode the whole record in `SlackHandler`

I like second more. What do you think @Seldaek @hkdobrev ?

hkdobrev: @an1zhegorodov Could you elaborate more on the second option? Would that work with the Slack API?

an1zhegorodov: @hkdobrev That should work according to [Slack API documentation](https://api.slack.com/methods/chat.postMessage)

> invalid_post_type	
> The method was called via a POST request, but the specified Content-Type was invalid. Valid types are: application/json application/x-www-form-urlencoded multipart/form-data text/plain.

I'll work on this tonight and test it out. Sorry for missing that in the first place.

hkdobrev: @an1zhegorodov That may be an outdated answer, though.

1. It's from 2015.
2. The Slack API documentation you have linked earlier says JSON is supported.
3. Slack API website now has a message builder tool and it uses JSON exclusively. See [an example](https://api.slack.com/docs/messages/builder?msg=%7B%22attachments%22%3A%5B%7B%22fallback%22%3A%22Required%20plain-text%20summary%20of%20the%20attachment.%22%2C%22color%22%3A%22%2336a64f%22%2C%22pretext%22%3A%22Optional%20text%20that%20appears%20above%20the%20attachment%20block%22%2C%22author_name%22%3A%22Bobby%20Tables%22%2C%22author_link%22%3A%22http%3A%2F%2Fflickr.com%2Fbobby%2F%22%2C%22author_icon%22%3A%22http%3A%2F%2Fflickr.com%2Ficons%2Fbobby.jpg%22%2C%22title%22%3A%22Slack%20API%20Documentation%22%2C%22title_link%22%3A%22https%3A%2F%2Fapi.slack.com%2F%22%2C%22text%22%3A%22Optional%20text%20that%20appears%20within%20the%20attachment%22%2C%22fields%22%3A%5B%7B%22title%22%3A%22Priority%22%2C%22value%22%3A%22High%22%2C%22short%22%3Afalse%7D%5D%2C%22image_url%22%3A%22http%3A%2F%2Fmy-website.com%2Fpath%2Fto%2Fimage.jpg%22%2C%22thumb_url%22%3A%22http%3A%2F%2Fexample.com%2Fpath%2Fto%2Fthumb.png%22%2C%22footer%22%3A%22Slack%20API%22%2C%22footer_icon%22%3A%22https%3A%2F%2Fplatform.slack-edge.com%2Fimg%2Fdefault_application_icon.png%22%2C%22ts%22%3A123456789%7D%5D%7D).

Have you tried it out on the API? I might be able to try it out tomorrow if you won't be able to.

src/Monolog/Handler/SlackWebhookHandler.php
```diff
@@ -0,0 +1,101 @@
...
+     * @param  string|null $channel                Slack channel (encoded ID or name)
+     * @param  string      $username               Name of a bot
+     * @param  bool        $useAttachment          Whether the message should be added to Slack as attachment (plain text otherwise)
+     * @param  string|null $iconEmoji              The emoji name to use (or null)
+     * @param  bool        $useShortAttachment     Whether the the context/extra messages added to Slack as attachments are in a short style
+     * @param  bool        $includeContextAndExtra Whether the attachment should include context and extra data
+     * @param  int         $level                  The minimum logging level at which this handler will be triggered
+     * @param  bool        $bubble                 Whether the messages that are handled can bubble up the stack or not
+     */
+    public function __construct(
```
an1zhegorodov: @hkdobrev Are we going to one-line arguments for consistency?
Also, we can make the order of arguments the same as in `SlackHandler` (except here we'll have `webhookUrl` instead of `token`). This makes switching between `SlackHandler` and `SlackWebhookHandler` much easier.

src/Monolog/Handler/SlackHandler.php
```diff
@@ -260,36 +151,40 @@ protected function write(array $record)
...
-                return 'good';
-            default:
-                return '#e3e4e6';
-        }
+        trigger_error(
+            'SlackHandler::getAttachmentColor() is deprecated. Use underlying SlackRecord instead.',
+            E_USER_DEPRECATED
+        );
+
+        return $this->slackRecord->getAttachmentColor($level);
```
an1zhegorodov: Just a thought. What if someone extended `SlackHanlder` just to make colors mapping different? He'll get default colors mapping after update, right? Looks like we're breaking backwards compatibility here. Same happens with `stringify`.

hkdobrev: Yes. We would be breaking extensions which adhere to the Liskov principle, because we are not calling this method anymore. The deprecation notice would help, but not to people not familiar with Monolog. However, this is only about people extending `SlackHandler` as we are not changing anything for people using the public API. This is only the `protected` API. So I think we could close our eyes for that BC change here, because we have a deprecation notice for people who extended it.

@Seldaek Do you think it's OK breaking the use of classes extended `SlackHandler`?

Seldaek: Yeah I think I can live with a minor break there.

Seldaek:
Note that the 2.0 constructor arguments are still very likely to change 
so don't attach too much importance to that.. Make it work for 1.x and 
we'll see.

an1zhegorodov:
@Seldaek Done that yesterday. @hkdobrev needs some time to try it out.

I'm using `SlackHandler` at work and have some improvements ideas, but I guess that'll be a separate PR.

mirfilip:
I haven't looked into the code yet, but I have a question. I'm currently using [this SlackWebhookHandler](https://github.com/Pageon/SlackWebhookMonolog). The one thing which is missing is the good formatter for `$context` and `$extra`. Yes, it's readable for the message, but when you pass context/extra, there is no good formatter available.

The formatter that would pretty print the array would be preferred. Back when my colleagues and I used our own logger, we used to turn on the markdown option (`mrkdwn`) and add an attachment with json serialized text. That's the prettiest way we come up with.

Any solutions to this problem here?

an1zhegorodov:
@mirfilip I guess the first thing is that `extra` and `context` should not really be included to the message, since we're adding them as attachment fields.

As for me, it requires much more work in terms of formatting, e.g. the user should be able to configure it to hide/show fields in context/extra.

Can you send us some screenshots of what you have in Slack using `SlackWebhookHandler`?

hkdobrev:
@Seldaek I've added some very basic unit tests for `SlackbotHandler` and `SlackWebhookHandler` as well. I know they look a bit like dummy tests, but I believe there should be at least one unit test in the test suite which would actually instantiate a handler. This would check for things like syntax errors and basic omissions.

With that I think this PR is ready. Thanks to @an1zhegorodov for picking up the slack (pun intended)! 🎉  💪

mirfilip:
@an1zhegorodov Stringify is indeed _some_ improvement, but with multidimensional arrays it's still pretty much useless.

One way I've seen this solved is either:
- Normalize (`NormalizerFormatter`) first, then json encode with pretty print
or 
-  `print_r` or similar

an1zhegorodov:
@mirfilip Once this is merged, I'll come up with a separate PR addressing this and a few other issues. Would appreciate your feedback.

hkdobrev:
@mirfilip @an1zhegorodov I also think it's better a new issue to be opened for this and tackled separately from the majority of the Slack handlers work.

wysow:
Hey guys, Thanks for this amazing work, can we see it merged soon @Seldaek ?

Seldaek:
Thanks everyone :)

Seldaek:
1.22 is out..

gmponos:
May I ask why cURL was preferred over Socket? Is it possible you could to move to socket instead of cURL?

hkdobrev:
The `SlackHandler` uses a socket with the streaming API. The simpler `SlackbotHandler` and the `SlackWebhookHandler` use cURL.

I'm not sure if these handlers could use a socket. cURL is simpler and based on their examples in the docs. I think only the streaming API would require a socket.

Why do you need to use a socket? If you use webhooks or the Slackbot, their examples give you just a webhook URL.

gmponos:
I hope I am not saying something incredibly stupid but I think using Sockets you can fire and forget am HTTP request from PHP.

https://stackoverflow.com/a/14668762/4158811
https://stackoverflow.com/a/14982587/4158811

The reason is that I don't want my application to break/stay halt if my Slack service is down.

I can see also that with the current implementation of SlackWebhook it retries the cURL request five times and I don't see a timeout setting so most probably my app will stay frozen while trying to call the webhook.

Not sure though if SocketHandler of monolog is also waiting for the response or it has a "fire and forget" logic.

mintobit:
"Fire and forget" would only work if there's no need to write data to the socket. We need to do that, so it would be locking operation with either Sockets or cURL.
To avoid application crash if Slack is down, you can use [WhatFailureGroupHandler](https://github.com/Seldaek/monolog/blob/master/doc/02-handlers-formatters-processors.md#wrappers--special-handlers).
Timeout setting is indeed missing, can be easily implemented the same way as in `SocketHandler` with fallback to ini settings.

If performance is critical, I'd also consider using `AmqpHandler` or `RedisHandler` and then sending messages to Slack using separate process.

PR #856, opened 2016-09-25
Fixing .gitattributes

arcanedev-maroc:


Seldaek:
I don't see how this fixes anything sorry..

Seldaek:
Ah well yeah the missing export-ignore I see now.. Thanks for that, I'll add it, but if you avoid reformatting the whole file it's a lot clearer :)

PR #865, opened 2016-10-05
Don't call to close in FingersCrossedHandler and GroupHandler of their handlers

### LINKED ISSUE #864: Don't call to close in FingersCrossedHandler and GroupHandler of their handlers
Currently the method close is not being called from `FingersCrossedHandler` 👉 [link](https://github.com/Seldaek/monolog/blob/21dde4dedbbbefe3f8055ce5c1d4aaf8516d9ec6/src/Monolog/Handler/FingersCrossedHandler.php#L131-L143) and it doesn't exist in `GroupHandler` 👉 [link](https://github.com/Seldaek/monolog/blob/21dde4dedbbbefe3f8055ce5c1d4aaf8516d9ec6/src/Monolog/Handler/GroupHandler.php)

For example, with a configuration:

``` yml
monolog:
    main:
            type:         fingers_crossed
            action_level: error
            handler:      group
    group:
            type: group
            members:
                - rollbar
                - file
    rollbar:
            type: rollbar
            level: info
            token: "%rollbar_token%"
            formatter: monolog.formatter.json
    file:
            type:  stream
            path:  "%kernel.logs_dir%/%kernel.environment%.log"
            level: debug
```

When there is a `Fatal Error`, it is not sent to Rollbar because they are sent in close method of `RollbarHandler` (https://github.com/Seldaek/monolog/blob/96f4fd718f8a547f5c5eeadf06383f097f4e3c08/src/Monolog/Handler/RollbarHandler.php#L111-L116)

Related: https://github.com/Seldaek/monolog/commit/2525053e8ef01d7fe703e58aba5300e7881f9fc6

I'm working in a PR...

rodrigm:
Closes #864

### DIFF
--- a/src/Monolog/Handler/FingersCrossedHandler.php
+++ b/src/Monolog/Handler/FingersCrossedHandler.php
@@ -140,6 +140,10 @@ class FingersCrossedHandler extends Handler implements ProcessableHandlerInterfa
                 $this->buffer = [];
             }
         }
+
+        if ($this->handler instanceof AbstractHandler) {
+            $this->handler->close();
+        }
     }
 
     /**
--- a/src/Monolog/Handler/FingersCrossedHandler.php
+++ b/src/Monolog/Handler/FingersCrossedHandler.php
@@ -141,9 +141,7 @@ class FingersCrossedHandler extends Handler implements ProcessableHandlerInterfa
             }
         }
 
-        if ($this->handler instanceof AbstractHandler) {
-            $this->handler->close();
-        }
+        $this->handler->close();
     }
 
     /**
--- a/src/Monolog/Handler/GroupHandler.php
+++ b/src/Monolog/Handler/GroupHandler.php
@@ -101,4 +101,16 @@ class GroupHandler extends Handler implements ProcessableHandlerInterface
 
         return $this;
     }
+
+    /**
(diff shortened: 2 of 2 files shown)


src/Monolog/Handler/FingersCrossedHandler.php
```diff
@@ -140,6 +140,10 @@ public function close()
                 $this->buffer = [];
             }
         }
+
+        if ($this->handler instanceof AbstractHandler) {
```
stof: this check is useless. ``->close()`` is part of the HandlerInterface itself

src/Monolog/Handler/GroupHandler.php
```diff
@@ -101,4 +101,16 @@ public function setFormatter(FormatterInterface $formatter)
...
         return $this;
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function close()
+    {
+        foreach ($this->handlers as $handler) {
+            if ($handler instanceof AbstractHandler) {
```
stof: useless here too

src/Monolog/Handler/FingersCrossedHandler.php
rodrigm: fix

PR #866, opened 2016-10-10
Remove substitution for not defined variables

p-golovin:
If variable in extra or context isn't defined, than placeholder will not not be replaced in format string. This patch remove all not replaced placeholders


src/Monolog/Formatter/LineFormatter.php
```diff
@@ -76,13 +76,17 @@ public function format(array $record)
             }
         }
 
+	    $output = preg_replace('/%extra\..+?%/', '', $output);
```
stof: wrong indentation

Seldaek:
Thanks

PR #868, opened 2016-10-12
Added the NoopHandler.

ComaVN:
This handler handles anything, but does nothing, and does not stop bubbling to the rest of the stack, unlike the NullHandler.

This can be used for testing, or to disable a handler when overriding a configuration without influencing the rest of the stack.

The specific use-case I need this for is in symfony, where `config_test.yml` imports `config_dev.yml`, but needs to disable some handlers and add some others:

`config_dev.yml`:

```
monolog:
    handlers:
        stderr:
            type:   stream
            path:   "php://stderr"
            level:  debug
```

`config_test.yml`:

```
imports:
    - { resource: config_dev.yml }
monolog:
    handlers:
        test:
            type:   test
            level:  debug
        stderr:
            type:   'null'
```

This fails, because the `stderr` handler is first in the stack, even after overriding it. Because the NullHandler doesn't bubble, the `test` handler is never called. This would be solved with `type: 'noop'`

### DIFF
--- a/doc/02-handlers-formatters-processors.md
+++ b/doc/02-handlers-formatters-processors.md
@@ -113,6 +113,8 @@
    to the wrapped handler.
 - [_SamplingHandler_](../src/Monolog/Handler/SamplingHandler.php): Wraps around another handler and lets you sample records
    if you only want to store some of them.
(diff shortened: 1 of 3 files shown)


Seldaek:
Thanks

PR #879, opened 2016-11-03
Add SqsHandler

martijncalker:


Seldaek:
Thanks, it's quite simple so I think we can include it.. Would you mind adding it to the docs though? To https://github.com/Seldaek/monolog/blob/master/doc/02-handlers-formatters-processors.md I mean

martijncalker:
Thanks you. I've added a link to the docs. Is that ok for you?

Seldaek:
Looks good thanks

PR #880, opened 2016-11-06
Fix tests for sentry/sentry >= 0.16.0 (and < 1.0)

siwinski:
From composer.json: `"sentry/sentry": "^0.13"`
Latest 0.x `sentry/sentry`: 0.22.0

sentry/sentry 0.16.0 changed the function signature of Raven_Client::capture():
0.16.0: `public function capture($data, $stack = null, $vars = null)`
0.15.0: `public function capture($data, $stack, $vars = null)`

While the test suite silently passes with PHP < 7.1, running the test suite with
PHP 7.1 fails with the following error:
`Declaration of Monolog\Handler\MockRavenClient::capture($data, $stack, $vars = NULL)
should be compatible with Raven_Client::capture($data, $stack = NULL, $vars = NULL)`

### DIFF
--- a/tests/Monolog/Handler/MockRavenClient-gte-0-16-0.php
+++ b/tests/Monolog/Handler/MockRavenClient-gte-0-16-0.php
@@ -1,0 +1,27 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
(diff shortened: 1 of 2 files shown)


siwinski:
Test suite failures are due to non-related failures due to PHP 7.1 as well:

```
Monolog\LoggerTest::testUseMicrosecondTimestamps with data set "without microseconds" (false, 'assertSame')

Failed asserting that two strings are identical.
```

Seldaek:
Thanks

PR #882, opened 2016-11-07
Add CsvHandler

jaymoulin:


Seldaek:
Thanks but I don't really see the point of this when we have JSON and other more powerful formats available.. So I would rather suggest that you publish this as a third party handler package and add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

jaymoulin:
The point to have a CsvHandler is to be able to store directly those files via a "load data infile". Files can be written down a SSD and consumed later to be treated in SQL database for analysis. I would do a Third party app but I need a fix in StreamHandler to keep it clean. May I create another PR with this fix only?

Seldaek:
Yeah I am fine with the StreamHandler patch, but I think you should rather pass $record to streamWrite() to give more flexibility instead of passing only the formatted part to it.

Seldaek:
Also create the PR against the 1.x branch otherwise it'll be only for monolog 2.0 which is not yet stable :)

Issue #883, opened 2016-11-09
MongoDBFormatter DateTime formatting fails


nikolaposa:
There's a bug in `MongoDBFormatter`, namely its `formatDate()` method, that results in following error: 

 `TypeError: MongoDB\BSON\UTCDateTime::__construct() expects parameter 1 to be integer, string given`. 

So, `UTCDateTime` class accepts integer instead of string, and fails when strict typing has been turned on.

I was able to reproduce this on PHP 7.0.8, ext-mongodb 1.1.9.

PR #884, opened 2016-11-09
Fixing date/time formatting in MongoDBFormatter

nikolaposa:
Fixes: https://github.com/Seldaek/monolog/issues/883

### DIFF
--- a/src/Monolog/Formatter/MongoDBFormatter.php
+++ b/src/Monolog/Formatter/MongoDBFormatter.php
@@ -105,6 +105,26 @@ class MongoDBFormatter implements FormatterInterface
 
     protected function formatDate(\DateTimeInterface $value, int $nestingLevel): UTCDateTime
     {
+        if (version_compare(phpversion('mongodb'), '1.1.9', '<=')) {
+            return $this->legacyGetMongoDbDateTime($value);
+        }
+
+        return $this->getMongoDbDateTime($value);
+    }
+
+    final protected function getMongoDbDateTime(\DateTimeInterface $value): UTCDateTime
+    {
         return new UTCDateTime((string) floor($value->format('U.u') * 1000));
+    }
+
+    final protected function legacyGetMongoDbDateTime(\DateTimeInterface $value): UTCDateTime
+    {
+        $milliseconds = floor($value->format('U.u') * 1000);
+
+        $milliseconds = (PHP_INT_SIZE == 8) //64-bit OS?
+            ? (int) $milliseconds
+            : (string) $milliseconds;
+
+        return new UTCDateTime($milliseconds);
     }
 }


nikolaposa:
This actually happens only on environments with `ext-mongodb <= 1.1.9`.

src/Monolog/Formatter/MongoDBFormatter.php
```diff
@@ -105,6 +105,12 @@ protected function formatException(\Throwable $exception, int $nestingLevel)
 
     protected function formatDate(\DateTimeInterface $value, int $nestingLevel): UTCDateTime
     {
-        return new UTCDateTime((string) floor($value->format('U.u') * 1000));
+        $miliseconds = floor($value->format('U.u') * 1000);
+
+        $miliseconds = version_compare(phpversion('mongodb'), '1.1.9', '>')
```
alcaeus: For versions after 1.1.9 this is correct, but needs refining for versions before 1.1.9: According to [the constructor](https://github.com/mongodb/mongo-php-driver/blob/1.1.9/src/BSON/UTCDateTime.c#L61..L83), on 64-bit systems it would parse the parameter as an integer while it expects a string on 32-bit systems. So, when running this code on a 32-bit system with `strict_types` enabled it would still error out. You can use the `PHP_INT_SIZE` constant to figure out the correct behavior for older driver versions.

src/Monolog/Formatter/MongoDBFormatter.php
```diff
@@ -105,6 +105,12 @@ protected function formatException(\Throwable $exception, int $nestingLevel)
 
     protected function formatDate(\DateTimeInterface $value, int $nestingLevel): UTCDateTime
     {
-        return new UTCDateTime((string) floor($value->format('U.u') * 1000));
+        $miliseconds = floor($value->format('U.u') * 1000);
```
nikolaposa: Keen eye, thanks, I'll fix that.

src/Monolog/Formatter/MongoDBFormatter.php
nikolaposa: Yeah, you're right. Wow, now I'm not even sure whether all this makes sense, because it's gonna be ugly fix, and it is ugly already.

nikolaposa:
@alcaeus I did the necessary changes, I squashed my commits, too, so your remarks should be covered now?

alcaeus:
@nikolaposa Looks good now. As "ugly" as it may seem, it's the only way to ensure the code works when using strict typing. Good job fixing this. 👍

Seldaek:
Thanks, hopefully 1.2 comes out soon and then we can remove this hackery at some point :)

jmikola:
> Thanks, hopefully 1.2 comes out soon and then we can remove this hackery at some point :)

[1.2.0](https://github.com/mongodb/mongo-php-driver/releases/tag/1.2.0) has been released, which includes the fix for this issue ([PHPC-804](https://jira.mongodb.org/browse/PHPC-804)). I imagine it is still present in the 1.1.x branch, though.

PR #885, opened 2016-11-14
Isolate fwrite to allow override

jaymoulin:


Seldaek:
Thanks, merged it in 1.x & master

PR #886, opened 2016-11-15
Public generic methods in TestHandler

julienfalque:
Makes `TestHandler::hasRecordRecords()` and `TestHandler::hasRecord()` public so they can be used in tests directly just like the other generic methods. Also renames `hasRecordRecords` to `hasRecords` because _hasRecordRecords_ is kinda weird.
Not sure the `protected` to `public` change would be a BC break for 1.x. If not, I can set the `hasRecordRecords` method back with a deprecation message. WDYT?


Seldaek:
Thanks, merged in 1.x too

PR #887, opened 2016-11-17
enable 7.1 on travis

remicollet:


Seldaek:
OK let's see, it worked for me locally with 7.1 but maybe travis will surface that same error.

Seldaek:
OK well anyway we should build against 7.1 here so thanks for the PR :)

PR #894, opened 2016-11-27
Slack improvements

an1zhegorodov:
While #846 was focused on providing the ability to communicate with Slack via webhooks/bot, there's a bunch of improvements that can be done before 2.x. Thanks to @greeny, @websirnik, @jewome62 and @mirfilip for ideas.

- [x] Exclude `extra`/`context`, `datetime`, `level` from message when attachment is used
- [x] Use `ts` attachment key to display `datetime` considering user timezone
- [x] [Support](https://github.com/Seldaek/monolog/pull/846#issuecomment-249528719) custom user images
- [x] [Allow](https://github.com/Seldaek/monolog/pull/894#issuecomment-263532399) to setup username from slack
- [x] [Improve](https://github.com/Seldaek/monolog/pull/846#issuecomment-261529198) array formatting within `context`/`extra`
- [x] [Support](https://github.com/Seldaek/monolog/issues/745) `include_stacktraces` option when attachment is not used and always include stacktraces when attachment is used
- [x] Support `extra`/`context` field exclusion
- [x] Update tests

### DIFF
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -46,10 +46,10 @@ class SlackRecord
     private $username;
 
     /**
-     * Emoji icon name
+     * User icon e.g. 'ghost', 'http://example.com/user.png'
      * @var string
      */
-    private $iconEmoji;
+    private $userIcon;
 
     /**
      * Whether the message should be added to Slack as attachment (plain text otherwise)
@@ -79,11 +79,11 @@ class SlackRecord
      */
     private $lineFormatter;
 
-    public function __construct($channel = null, $username = 'Monolog', $useAttachment = true, $iconEmoji = null, $useShortAttachment = false, $includeContextAndExtra = false, FormatterInterface $formatter = null)
+    public function __construct($channel = null, $username = 'Monolog', $useAttachment = true, $icon = null, $useShortAttachment = false, $includeContextAndExtra = false, FormatterInterface $formatter = null)
     {
         $this->channel = $channel;
         $this->username = $username;
-        $this->iconEmoji = trim($iconEmoji, ':');
+        $this->userIcon = trim($icon, ':');
         $this->useAttachment = $useAttachment;
         $this->useShortAttachment = $useShortAttachment;
         $this->includeContextAndExtra = $includeContextAndExtra;
@@ -154,8 +154,12 @@ class SlackRecord
             $dataArray['text'] = $message;
         }
 
-        if ($this->iconEmoji) {
-            $dataArray['icon_emoji'] = ":{$this->iconEmoji}:";
+        if ($this->userIcon) {
+            if (filter_var($this->userIcon, FILTER_VALIDATE_URL)) {
+                $dataArray['icon_url'] = $this->userIcon;
+            } else {
+                $dataArray['icon_emoji'] = ":{$this->userIcon}:";
+            }
         }
 
         return $dataArray;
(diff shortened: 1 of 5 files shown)


jewome62:
Can you remove default value for username into SlackWebhookHandler (in construct)
We can into Slack for webhook define default name.
But with monolog, the default name is never use because we send Monolog by default.

Thanks

an1zhegorodov:
@jewome62 Good idea. This will allow to use slack internal username settings. I'll check how it works for slack API and get that done. Thanks.

danizord:
Looking forward to the removal of context fields duplicated in message :+1:

an1zhegorodov:
@Seldaek Do you think we can do better to support `include_stacktraces`?
#846 implemented `getFormatter`, thus if handler has a formatter which supports `include_stacktraces` and  `useAttachment === false` - it works 
In case `useAttachment === true`, we now have `NormalizerFormatter` in `SlackRecord::stringify` which will always include stacktraces and will ignore `include_stacktraces`. This works for people who want traces in slack, but doesn't work for those who don't.

The only way I see to fully support `include_stacktraces` is `SlackFormatter`, which would handle both `message` and `attachment` and would have `includeStacktraces()` setter.

Seldaek:
@an1zhegorodov seems like you say a way to do it so I'd say do it if you think there is value in it :)

an1zhegorodov:
@Seldaek Just realized it would also require moving most of constructor arguments to the formatter... We might consider that for 2.x

an1zhegorodov:
Some visuals on these changes:
- [1.x_short_attachment](http://storage5.static.itmages.com/i/16/1209/h_1481326922_2832863_0dc40bce31.png) vs [slack_improvements_short_attachment](http://storage6.static.itmages.com/i/16/1209/h_1481326945_3938609_d4cbc39492.png)
- [1.x_long_attachment](http://storage3.static.itmages.com/i/16/1209/h_1481327078_8059829_c2aa12d209.png) vs [slack_improvements_long_attachment](http://storage1.static.itmages.com/i/16/1209/h_1481327011_5174376_c1e7538803.png)

@hkdobrev, @stof, would appreciate a code review

riccardomessineo:
Awaiting this merge to fix issue #902 :) 
Thanks @an1zhegorodov

Seldaek:
From looking at your screenshots @an1zhegorodov it seems to me one important thing that might be worth adding again is the `$record['datetime']`. It seems it got lost :)

Otherwise I agree it looks much better, thanks for the work.

an1zhegorodov:
@Seldaek Slack has a feature to show the date and time below the attachment based on user timezone. That's where `datetime` is now.

Seldaek:
OK sorry :) Looks fine to me then, thanks!

danizord:
@Seldaek can you release a tag with this one? :)

mirfilip:
@an1zhegorodov How is the `include_stacktraces` switch handled? Let's say I don't use the attachment but want to choose whether to include stacktraces.

an1zhegorodov:
@mirfilip Whether it is handled or not depends [on the formatter](https://github.com/symfony/monolog-bundle/blob/master/MonologBundle.php#L52). If you use default formatter (which is `LineFormatter`), then you can use `inlcude_stacktraces`. However, the attachment will always include stacktraces, because it uses `NormalizerFormatter`. Here's [a comment](https://github.com/Seldaek/monolog/pull/894#issuecomment-265594272) I wrote during implementation, might also give you and idea how this works.

mirfilip:
@an1zhegorodov Stupid me, I was searching for `include_stacktraces` :( Your comment makes it crystal clear. Again, thx for a good work.

Issue #516, opened 2015-02-28
Provide a interface and class to handle records as objects


dnaber-de:
While I was writing a processor recently, I stumbled over the array passed to the processors `__invoke` method. I was not sure about the arrays structure, so I quickly dumped it using `var_dump()`. It would be nice to have an object here with an interface to work against. 

Another benefit of an record object would be, that processors needn't to take care about each other when more of them add extra data for example.

Here's a simple example of an record interface:

``` php
<?php # -*- coding: utf-8 -*-

namespace Monolog\Type;

interface RecordInterface {

    /**
     * @param string $name
     */
    public function setMessage( $name );

    /**
     * @return string
     */
    public function getMessage();

    /**
     * @param string $context
     */
    public function setContext( $context );

    /**
     * @return string
     */
    public function getContext();

    /**
     * @param int
     */
    public function setLevel( $level );

    /**
     * @return int
     */
    public function getLevel();

    /**
     * @param string $levelName
     */
    public function setLevelName( $levelName );

    /**
     * @return string
     */
    public function getLevelName();

    /**
     * @param string $channel
     */
    public function setChannel( $channel );

    /**
     * @return string
     */
    public function getChannel();

    /**
     * @param \DateTime $dateTime
     */
    public function setDateTime( \DateTime $dateTime );

    /**
     * @return \DateTime
     */
    public function getDateTime();

    /**
     * @param array $extra
     */
    public function setExtra( array $extra );

    /**
     * @return array
     */
    public function getExtra();
}
```

It might be discussable whether both `setLevel()` and `setLevelName()` are necessary as this might lead to an inconsistency of the object. However It's just an idea for an improvement.

Seldaek:
This is not really feasible if only for BC reasons, although we could technically implement ArrayAccess but that'd make everything slower for no great benefit.

The concept of record is kind of the only thing to understand about monolog internals so I don't think it's too much to ask :) But I get what you are saying though and I do think it could be documented better for sure.

dnaber-de:
Well, I didn't had backward compatibility in mind when I thought about that. Of course that is more important.

PR #895, opened 2016-12-02
Add documentation on internal log message structure

rpkamp:
Addresses https://github.com/Seldaek/monolog/issues/516

### DIFF
--- a/doc/message-structure.md
+++ b/doc/message-structure.md
@@ -8,11 +8,11 @@ key        | type                      | description
 message    | string                    | The log message. This string may contain placeholders that will be replaced by variables from the context, e.g., "User %username% logged in" with `['username' => 'John']` as context will be written as "User John logged in".
 level      | int                       | Severity of the log message. See log levels described in [01-usage.md](01-usage.md).
 level_name | string                    | String representation of log level.
-context    | array                     | Arbitrary data passed with the constuction of the message. For example the username of the current user or their IP addres.
+context    | array                     | Arbitrary data passed with the construction of the message. For example the username of the current user or their IP address.
 channel    | string                    | The channel this message was logged to.
 datetime   | Monolog\DateTimeImmutable | Date and time when the message was logged. Class extends `\DateTimeImmutable`.
 extra      | array                     | A placeholder array where processors can put additional data. Always available, but empty if there are no processors registered.
 
-At first glance `context` and `extra` look very similar, and they are in the sense that they both carry arbitrary data that is related to log the log message somehow.
-The main difference is that `context` can be supplied in user land (it is the 3rd parameter to `Logger::addRecord()`) whereas `extra` is internal only and can be filled processors.
+At first glance `context` and `extra` look very similar, and they are in the sense that they both carry arbitrary data that is related to the log message somehow.
+The main difference is that `context` can be supplied in user land (it is the 3rd parameter to `Logger::addRecord()`) whereas `extra` is internal only and can be filled by processors.
 The reason processors write to `extra` and not to `context` is to prevent overriding any user provided data in `context`.
--- a/doc/message-structure.md
+++ b/doc/message-structure.md
@@ -8,13 +8,11 @@ key        | type                      | description
 message    | string                    | The log message. This string may contain placeholders that will be replaced by variables from the context, e.g., "User %username% logged in" with `['username' => 'John']` as context will be written as "User John logged in".
 level      | int                       | Severity of the log message. See log levels described in [01-usage.md](01-usage.md).
 level_name | string                    | String representation of log level.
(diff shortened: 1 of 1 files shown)


doc/message-structure.md
```diff
@@ -0,0 +1,20 @@
...
+
+Within monolog log messages are passed around as arrays, for example to processors or handlers.
+The table below describes which keys are always available for every log message.
+
+key        | type                      | description
+-----------|---------------------------|-------------------------------------------------------------------------------
+message    | string                    | The log message. This string may contain placeholders that will be replaced by variables from the context, e.g., "User %username% logged in" with `['username' => 'John']` as context will be written as "User John logged in".
+level      | int                       | Severity of the log message. See log levels described in [01-usage.md](01-usage.md).
+level_name | string                    | String representation of log level.
+context    | array                     | Arbirtrary data passed with the constuction of the message. For example the username of the current user or their IP addres.
```
stof: typo: ``arbitrary``

doc/message-structure.md
```diff
@@ -0,0 +1,20 @@
...
+channel    | string                    | The channel this message was logged to.
+datetime   | Monolog\DateTimeImmutable | Date and time when the message was logged. Class extends `\DateTimeImmutable`.
+extra      | array                     | A placeholder array where processors can put additional data. Always available, but empty if there are no processors registered.
+
+Some handlers and/or processors may add additional keys to the array.
```
stof: Processors should not add new keys. This is not a supported usage pattern. They should store them in ``extra``.

And handlers are not adding keys to the record either (they could be turning the record into another data structure needed by the target system though of course)

rpkamp:
Both good points, thank you!

I just fixed them.

How do you feel about lines 16 through 19? I couldn't find anywhere that this was the case, but it felt right intuitively. I mentioned it because I've been confused about the existence of two arrays that appear to serve the same purpose for quite some time.

an1zhegorodov:
@rpkamp I recall I had a feeling like something is missing in Monolog documentation. That's it! Thanks!
And lines 16-19 make perfect sense. I'd be very surprised if that's not that case. May be @Seldaek can give us hint on this?

doc/message-structure.md
```diff
@@ -0,0 +1,18 @@
...
+level      | int                       | Severity of the log message. See log levels described in [01-usage.md](01-usage.md).
+level_name | string                    | String representation of log level.
+context    | array                     | Arbitrary data passed with the constuction of the message. For example the username of the current user or their IP addres.
+channel    | string                    | The channel this message was logged to.
+datetime   | Monolog\DateTimeImmutable | Date and time when the message was logged. Class extends `\DateTimeImmutable`.
+extra      | array                     | A placeholder array where processors can put additional data. Always available, but empty if there are no processors registered.
+
+At first glance `context` and `extra` look very similar, and they are in the sense that they both carry arbitrary data that is related to log the log message somehow.
+The main difference is that `context` can be supplied in user land (it is the 3rd parameter to `Logger::addRecord()`) whereas `extra` is internal only and can be filled processors.
```
hkdobrev: I think the end of the sentence here should be something like "can be filled *by* processors`.

rpkamp:
@hkdobrev nice catch! fixed it and fixed a few more typos while I was at it :)

rpkamp:
ping @Seldaek

doc/message-structure.md
```diff
@@ -0,0 +1,18 @@
+# Log message structure
+
+Within monolog log messages are passed around as arrays, for example to processors or handlers.
+The table below describes which keys are always available for every log message.
+
+key        | type                      | description
+-----------|---------------------------|-------------------------------------------------------------------------------
+message    | string                    | The log message. This string may contain placeholders that will be replaced by variables from the context, e.g., "User %username% logged in" with `['username' => 'John']` as context will be written as "User John logged in".
```
Seldaek: This placeholder stuff is only true if you use the PsrLogMessageProcessor btw, maybe it should be mentioned.

doc/message-structure.md
```diff
@@ -0,0 +1,18 @@
...
+level      | int                       | Severity of the log message. See log levels described in [01-usage.md](01-usage.md).
+level_name | string                    | String representation of log level.
+context    | array                     | Arbitrary data passed with the construction of the message. For example the username of the current user or their IP address.
+channel    | string                    | The channel this message was logged to.
```
Seldaek: Might be good to clarify this is equal to the channel name passed to `new Logger($channel)`?

Seldaek:
Also would be cool to link to it from some of the other doc pages, otherwise looks good to me, thanks for the work :)

Seldaek:
Thanks, I'll add a link to it in the README at least :)

PR #896, opened 2016-12-02
Remove remaining artifacts from mlehner/gelf-php

rpkamp:
Support for mlehner/gelf-php seems to have been dropped (which makes sense, since the package is abandoned), but the tests were still there. They were skipped by default, but when I installed `mlehner/gelf-php` the tests failed because GelfHandler doesn't support the interface provided by mlehner/gelf-php anymore.

Also, GelfMessageFormatterTest.php suggested to install mlehner/gelf-php, which doesn't make sense since it's not supported anymore.


Seldaek:
Thanks

PR #900, opened 2016-12-10
Support for Apache Kafka

kozlice:
Hi,

This is a very simple support for logging into Apache Kafka using [rdkafka extension](https://github.com/arnaud-lb/php-rdkafka) (there are a few libraries for PHP, but this once seems to be most mature and supports Kafka versions 0.8, 0.9 & 0.10).

Tests included, phpunit's coverage report shows 100% for handler.
Travis config updated to install extension from PECL (and librdkafka headers which it requires).
Manually checked that it works with most recent Kafka release (0.10.1).


src/Monolog/Handler/KafkaHandler.php
```diff
@@ -0,0 +1,84 @@
...
+     */
+    protected $producer;
+
+    /**
+     * @param Producer $producer             Kafka message producer instance
+     * @param string   $topicName            Kafka topic name (if it doesn't exist yet, will be created)
+     * @param int      $level                The minimum logging level at which this handler will be triggered
+     * @param bool     $bubble               Whether the messages that are handled can bubble up the stack or not
+     */
+    public function __construct(Producer $producer, string $topicName, TopicConf $topicConfig = null, int $level = Logger::DEBUG, bool $bubble = true)
```
Seldaek: Docs are missing for $topicConfig here

src/Monolog/Handler/KafkaHandler.php
```diff
@@ -0,0 +1,84 @@
...
+{
+    /**
+     * @var ProducerTopic
+     */
+    protected $topic;
+
+    /**
+     * @var Producer
+     */
+    protected $producer;
```
stof: these should be private IMO, avoiding the need to maintain BC on them

tests/Monolog/Handler/KafkaHandlerTest.php
```diff
@@ -0,0 +1,76 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Test\TestCase;
+
+class KafkaHandlerTest extends TestCase
+{
+    public function setUp()
```
stof: should be protected

Seldaek:
Thanks, but I think this might be better as a standalone package like kozlice/monolog-kafka which can be listed at https://github.com/Seldaek/monolog/wiki/Third-Party-Packages too - that way you can maintain it and I don't have to maintain even more handlers because it's especially hard when there are third party dependencies like kafka which I don't know well. Also that way you could use it already without waiting for monolog 2.0 final to be out (because we might still have BC breaks I wouldn't recommend relying on master branch at the moment).

.travis.yml
```diff
@@ -1,10 +1,14 @@
...
-sudo: false
+sudo: required
+
+before_install:
+  - sudo ./tests/install-librdkafka.sh v0.9.2 $HOME
 
 cache:
   directories:
     - $HOME/.composer/cache
+    - $HOME/librdkafka
```
stof: is this cache actually useful ?

stof:
@Seldaek what is the state of Monolog 2 ? Will it be released soon ? If no, it would be better to add this class in the 1.x branch.

stof:
hmm, I agree with you about shipping it externally instead.

kozlice:
Fair enough, Kafka is a specific thing. I'll create a separate package for this handler. Thanks for review, I'll need to fix those anyway

Seldaek:
Great I'll close this then, feel free to add it to the wiki once it's published :)

PR #908, opened 2016-12-30
Add missed return phpdoc tag.

Aliance:

### DIFF
--- a/src/Monolog/Handler/RavenHandler.php
+++ b/src/Monolog/Handler/RavenHandler.php
@@ -221,7 +221,7 @@ class RavenHandler extends AbstractProcessingHandler
 
     /**
      * @param string $value
-     * @return $this
+     * @return self
      */
     public function setRelease($value)
     {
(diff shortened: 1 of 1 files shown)


Seldaek:
Please use `@return self` if anything, and send the pull request against the 1.x branch, because in master branch it targets php7 and should just be `function setRelease($release): self`.

Aliance:
May I edit this pr (branch can be edited, you need just to reopen it) & make another for php7 branch?

Seldaek:
You can open a new PR from a new branch or the same branch.. up to you as long as it goes to the right branch here :) I can reopen if you like but it doesn't matter much.

Aliance:
Do not like reopening because of 'broken' pr ID :rofl: Nevermind, do I make the right changes?

Aliance:
What about adding `.idea` directory to the git ignore?

Aliance:
@Seldaek any response?

staabm:
> What about adding .idea directory to the git ignore?

usually you should have that in your global gitconfig instead of each project

PR #911, opened 2017-01-07
Replacing the add* methods in the docs and comments with the PSR-3 compliant methods

StevenWInfo:
Noticed some of the docs had code using `addInfo` which I had also noticed was removed in the recent upgrade, so I did some grepping and replaced mentions of it with `info` in some comments and documentation. I did this with the other add* methods as well.


Seldaek:
Thanks

PR #914, opened 2017-01-12
Logmatic fix

gpolaert:
This PR add a field marker to the Logmatic.io handler.
Logmatic.io is designing an integration for PHP/Monolog (https://logmatic.io/integrations/#integration-php)

### DIFF
--- a/src/Monolog/Handler/LogmaticHandler.php
+++ b/src/Monolog/Handler/LogmaticHandler.php
@@ -23,17 +23,17 @@ class LogmaticHandler extends SocketHandler
     /**
      * @var string
      */
-    protected $logToken;
+    private $logToken;
 
     /**
      * @var string
      */
-    protected $hostname;
+    private $hostname;
 
     /**
      * @var string
      */
-    protected $appname;
+    private $appname;
 
     /**
      * @param string     $token    Log token supplied by Logmatic.


gpolaert:
Hello all, @Seldaek 
Do I need to open any issue or something? Or the PR is sufficient?

Seldaek:
It is sufficient I am just very busy atm and not much time for following up on opensource :/

src/Monolog/Formatter/LogmaticFormatter.php
```diff
@@ -11,6 +11,10 @@
 
 namespace Monolog\Formatter;
 
+const APP_FIELD_NAME = "appname";
+const HOST_FIELD_NAME = "hostname";
+const MARKER_FIELD_NAME = "@marker";
```
Seldaek: Please define these as class constants if anything, although I don't know that changing existing code for the sake of changing it brings much value. But anyway they can't be just defined as regular constants as it might clash with other stuff.

src/Monolog/Formatter/LogmaticFormatter.php
```diff
@@ -1,4 +1,4 @@
-<?php declare(strict_types=1);
+<?php declare(strict_types = 1);
```
Seldaek: Please don't change this :)

src/Monolog/Formatter/LogmaticFormatter.php
```diff
@@ -19,6 +23,11 @@
 class LogmaticFormatter extends JsonFormatter
 {
     /**
+     * @param array $marker Used by Logmatic.io to identify the technology
+     */
+    protected $marker = ["sourcecode", "php"];
```
Seldaek: I think there should at least be a setter so this can be changed no?

stof:
and this should be done in the 1.x branch of Monolog, to be available in stable releases soon

src/Monolog/Formatter/LogmaticFormatter.php
stof: and if it cannot be changed du to Logmatic requirements, it should not be a property at all

src/Monolog/Formatter/LogmaticFormatter.php
gpolaert: Sure, my IDE config ... :)

src/Monolog/Formatter/LogmaticFormatter.php
gpolaert: I will change it for a class constant.

gpolaert:
Hello @stof, @Seldaek,
Thanks a lot for the review. I will push today the updated version.

Seldaek:
Thanks

PR #915, opened 2017-01-12
Logmatic handler

gpolaert:
Add the Logmatic.io handler, extram from the master branch.

### DIFF
--- a/src/Monolog/Formatter/LogmaticFormatter.php
+++ b/src/Monolog/Formatter/LogmaticFormatter.php
@@ -21,14 +21,14 @@ class LogmaticFormatter extends JsonFormatter
 {
 
     /**
-     * @param string
+     * @param string The hostname reported to Logmatic.io
      */
-    protected $hostname = '';
+    private $hostname = '';
 
     /**
-     * @param string
+     * @param string The application reported to Logmatic.io
      */
-    protected $appname = '';
+    private $appname = '';
 
     /**
      * Set hostname
--- a/src/Monolog/Handler/LogmaticHandler.php
+++ b/src/Monolog/Handler/LogmaticHandler.php
@@ -22,17 +22,17 @@ class LogmaticHandler extends SocketHandler
     /**
      * @var string
      */
-    protected $logToken;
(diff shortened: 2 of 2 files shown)


src/Monolog/Formatter/LogmaticFormatter.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * @param string
+     */
+    protected $hostname = '';
+
+    /**
+     * @param string
+     */
+    protected $appname = '';
```
stof: I would make these private to simplify the handling of backward compatibility

src/Monolog/Handler/LogmaticHandler.php
```diff
@@ -0,0 +1,87 @@
...
+
+    /**
+     * @var string
+     */
+    protected $hostname;
+
+    /**
+     * @var string
+     */
+    protected $appname;
```
stof: same here

tests/Monolog/Handler/LogmaticHandlerTest.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    private function createHandler()
+    {
+        $useSSL = extension_loaded('openssl');
+        $args = array('testToken', 'testHostname', 'testAppname', $useSSL, Logger::DEBUG, true);
+        $this->res = fopen('php://memory', 'a');
+        $this->handler = $this->getMockBuilder('Monolog\Handler\LogmaticHandler')
+            ->setConstructorArgs($args)
+            ->setMethods(['fsockopen', 'streamSetTimeout', 'closeSocket'])
```
stof: you still have a short array syntax here

src/Monolog/Handler/LogmaticHandler.php
gpolaert: Same behavior the hostname prop?

Seldaek:
Sorry but rather not add more handlers to 1.x at this point.. It's in master and I'll try to wrap up 2.0 so it becomes usable in a stable version :)

PR #916, opened 2017-01-18
Splunk Formatter

dsposito:
This adds a new formatter that is compatible with Splunk's parser. It ensures that a "timestamp" key exists at the beginning of the logged JSON.

### DIFF
--- a/src/Monolog/Formatter/SplunkFormatter.php
+++ b/src/Monolog/Formatter/SplunkFormatter.php
@@ -1,0 +1,42 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Formatter;
+
+use DateTimeInterface;
+
+/**
+ * Encodes message information into JSON in a format compatible with Splunk.
+ *
+ * @author Daniel Sposito <daniel.g.sposito@gmail.com>
+ */
+class SplunkFormatter extends JsonFormatter
+{
+    /**
+     * Prepends the timestamp parameter for indexing by Splunk.
+     *
+     * @see http://dev.splunk.com/view/logging-best-practices/SP-CAAADP6
+     * @see \Monolog\Formatter\JsonFormatter::format()
+     */
+    public function format(array $record): string
+    {
+        if (isset($record["datetime"]) && ($record["datetime"] instanceof DateTimeInterface)) {
+            $record = array_merge(
+                array("timestamp" => $record["datetime"]->format("Y-m-d H:i:s.u T")),
+                $record
+            );
+
+            unset($record["datetime"]);
+        }
+
+        return parent::format($record);
+    }
+}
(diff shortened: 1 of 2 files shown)


mpdude:
Alternatively, you could just initialize a LineFormatter with the date format you provided and Splunk should happily parse the _time from the line beginning?

dsposito:
@mpdude Thanks for the suggestion. That was our original solution (LineFormatter with custom datetime format string) but the extra meta data wasn't getting parsed correctly by Splunk. For example, Splunk would render a concatenated string of `message` and `extra` data as:

```[2016-11-11 23:30:04 PST] lumen.INFO: Retrieved Clean List From FTP Server {"count":3,"files":[".","..","Thumbs.db"]}```

The LineFormatter may have worked for Splunk in the past but it seems JSON is its preferred data structure now.

Let me know if you have any other suggestions to try.

mpdude:
Not sure what you're trying to do: Prefix the log line with a timestamp Splunk recognizes, or define a log format that can easily be parsed by Splunk (for example, every line being a single JSON string)?

I am basically using the default Monolog LineFormatter format with some extra markup that allows me to parse out the `context` and `extra` data as JSON strings. Then, in Splunk at run-time, use the `spath` function to parse out additional values (see https://answers.splunk.com/answers/117121/extract-json-data-within-the-logs.html for an example).

dsposito:
@mpdude We're just trying to define a log format that can easily be parsed by Splunk. This new SplunkFormatter includes miliseconds in the timestamp to ensure it's unique and forces `timestamp` to be the first key in the JSON data per Splunk's [recommendations](http://dev.splunk.com/view/logging-best-practices/SP-CAAADP6).

We could use LineFormatter and customize the Splunk parsing configuration via `spath` but this new formatter allows Splunk parsing/formatting to work out of the box. :-)

Here are before/after examples of what the parsed logs look like in Splunk:

**LineFormatter**
splunk-before

**SplunkFormatter**
splunk-after

mpdude:
Do you need to configure Splunk to interpret each logfile line as JSON data?

dsposito:
@mpdude Nope, that works out of the box with Splunk.

dsposito:
@mpdude Are you cool with merging and tagging a release for this? We've had success using this formatter during the past week+. We'd like to start using it in our other applications and it'd be great to use core monolog rather than our forked version.

mpdude:
@dsposito Sorry, I'm just a bystander and Splunk user, you'd need to wait for feedback from the repo owners/leads.

dsposito:
@Seldaek Can we get a 👍 from you?

Seldaek:
@dsposito would you mind rebasing this commit on the 1.x branch? Because master is for 2.0 and that's not ready to release, so if you want it in a release I need it targetted at 1.x :) Otherwise the code looks fine by me (except the php7 stuff that has to be removed for 1.x branch). Maybe just one thing would be to skip the `if (isset())` condition, as the datetime really is always there.

jlmaners:
Any update on this getting pulled in or moving forward?

Seldaek:
Closing this as it doesn't seem to have a ton of interest. Feel free to publish it as a third party package and add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

PR #926, opened 2017-02-17
Improves code coverage

nawarian:
This will improve code coverage on `Monolog\ErrorHandler` class.

I left `ErrorHandler::register` not fully covered on purpose, since I got no easy way to test handlers methods.

Hope this helps. :)


Seldaek:
Thanks

PR #929, opened 2017-02-21
Php5.3 autoload errors

fabriceparallel:
See this [PHP5.3 bug](https://bugs.php.net/bug.php?id=47987)

When an error is triggered on compile-time the autoloader is not called which create a "`LogLevel` class not found" error.

I make the ErrorHandler extend LogLevel so ErrorHandler will already have all the log level it's looking for, and the autoloader won't be called.

### DIFF
--- a/src/Monolog/ErrorHandler.php
+++ b/src/Monolog/ErrorHandler.php
@@ -58,6 +58,9 @@ class ErrorHandler
      */
     public static function register(LoggerInterface $logger, $errorLevelMap = array(), $exceptionLevel = null, $fatalLevel = null)
     {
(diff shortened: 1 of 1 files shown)


stof:
A better solution IMO would be to force loading the LogLevel class when registering the ErrorHandler (using ``class_exists('...\LogLevel');`` with a comment explaining the reason), to force loading the class earlier.
Extending LogLevel does not make sense. the ErrorHandler is not a LogLevel.

fabricekabongo:
I'm still testing your advice. Please don't review it.

fabricekabongo:
@stof it works perfectly for us. Thanks. See if you like it.

Seldaek:
Thanks

PR #930, opened 2017-02-21
Allow to set custom event_id [LEGACY]

tstrijdhorst:
Raven_Client allows for the event_id to be custom if it's in the data. So I made a PR that adds this to the extra parameters. I made the PR for 0.1.x since our company is not yet using PHP7. I also made a PHP7 version since it's rather simple:

https://github.com/Seldaek/monolog/pull/931

This is the code that allows it:
https://github.com/getsentry/sentry-php/blob/master/lib/Raven/Client.php#L785-L787


Seldaek:
Thanks

PR #931, opened 2017-02-21
Allow to set custom event_id [PHP7]

tstrijdhorst:
Raven_Client allows for the event_id to be custom if it's in the data. So I made a PR that adds this to the extra parameters.

For the 1.x legacy version see: https://github.com/Seldaek/monolog/pull/930

This is the code that allows it:
https://github.com/getsentry/sentry-php/blob/master/lib/Raven/Client.php#L785-L787


Seldaek:
Closing as the 1.x PR will be merged into master

PR #934, opened 2017-02-27
Micro-optimisation: delayed unnecessary `static::getLevelName`

zerkms:
I am micro-optimising one local project and have noticed that the

```php
$levelName = static::getLevelName($level);
```

inside the `Logger::addRecord()` can be moved slightly later so that it was not invoked when we don't need it (when there are no handlers for a given error level).


zerkms:
The failed check - is for php-nightly and is not relevant to this change :-S

Seldaek:
Thanks

PR #936, opened 2017-02-27
Replaced current/next-style loop with foreach

zerkms:
In this PR I optimised the hot path for the cases when there are no handlers to handle a message (due to a level).

In my environment the performance gain is about x2-3.

The cost of this change is slightly more complicated message handling code: it is more tangled now (so I'm open for suggestions on how to improve it). And it is likely to be slightly less performant, but in case when a message is handled - the handlers would be the slowest part, so the penalty for the lines I added should be negligible (it is my speculations, it needs a proof).

If necessary and if the dev team agrees this optimisation worth it, I may provide a small set of performance tests (but I'm not a performance engineer and don't do it professionally, so they likely to be not ideal).

So, what would be your thoughts?

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -283,14 +283,11 @@ class Logger implements LoggerInterface
 
         // check if any handler will handle this message so we can return early and save cycles
         $handlerKey = null;
-        reset($this->handlers);
-        while ($handler = current($this->handlers)) {
+        foreach ($this->handlers as $key => $handler) {
             if ($handler->isHandling(['level' => $level])) {
-                $handlerKey = key($this->handlers);
+                $handlerKey = $key;
                 break;
             }
-
-            next($this->handlers);
         }
 
         if (null === $handlerKey) {
@@ -311,7 +308,17 @@ class Logger implements LoggerInterface
             $record = call_user_func($processor, $record);
         }
 
+        reset($this->handlers);
+        $skip = true;
         while ($handler = current($this->handlers)) {
+            if ($skip) {
+                if ($key !== key($this->handlers)) {
+                    next($this->handlers);
+                    continue;
+                }
+                $skip = false;
+            }
+
             if (true === $handler->handle($record)) {
                 break;
             }
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -308,6 +308,7 @@ class Logger implements LoggerInterface
             $record = call_user_func($processor, $record);
         }
 
+        // advance the array pointer to the first handler that will handle this record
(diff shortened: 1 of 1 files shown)


zerkms:
The failed check - is for php-nightly and is not relevant to this change :-S

src/Monolog/Logger.php
```diff
@@ -311,6 +308,11 @@ public function addRecord(int $level, string $message, array $context = []): boo
             $record = call_user_func($processor, $record);
         }
 
+        reset($this->handlers);
+        while ($key !== key($this->handlers)) {
```
stloyd: Shouldn't this be `$handlerKey`?

src/Monolog/Logger.php
```diff
@@ -311,6 +308,11 @@ public function addRecord(int $level, string $message, array $context = []): boo
...
+            next($this->handlers);
```
staabm: as you dont do anything in this look, can't you just use `end` ?
http://php.net/manual/en/function.end.php

src/Monolog/Logger.php
zerkms: @stloyd it actually does not matter - they are guaranteed to be the same value at this point

src/Monolog/Logger.php
zerkms: @staabm I do "something" - I loop until we reach the point where the `foreach` above completed its execution. Effectively - we are looking for the last processed key.

src/Monolog/Logger.php
```diff
@@ -311,6 +308,11 @@ public function addRecord(int $level, string $message, array $context = []): boo
...
+        }
+
         while ($handler = current($this->handlers)) {
```
staabm: will this loop also benefit from a transformation into `foreach`?

zerkms: @staabm it would, if it was an easy way to rewind to the place we stopped in the first loop. My first implementation was `foreach`-based, but it looked uglier. Additionally, the minor overhead of `next/current` would be amortized by generally heavy handlers.

src/Monolog/Logger.php
Seldaek: It doesn't matter but for clarity I think it'd be better ot use handlerKey still.

Seldaek:
I find it surprising but I measured and indeed foreach seems much faster than a while loop that does essentially the same as foreach does internally.. 

As for the second while loop, it probably matters a lot less indeed, but if you find a reasonable way to optimize it away feel free.

src/Monolog/Logger.php
zerkms: Yep, changed it

Issue #933, opened 2017-02-26
GELF publisher can throw exceptions which should be handleable


derFunk:
Hi there,

I think https://github.com/Seldaek/monolog/issues/702 is still valid.

At least if a TCP/UDP GELF transport has ben configured, but the server is not reachable for any reason, `gelf-php`'s `StreamSocketClient->buildSocket()`([See here](https://github.com/bzikarsky/gelf-php/blob/bdf3c93d62d8e2896879c7a9e3f0df48e50fd436/src/Gelf/Transport/StreamSocketClient.php#L137)) would throw an exception which is not handled by `GelfHandler`.
(I didn't check if other gelf publishers would also show this behaviour.)

This has also been discussed in the `gelf-php` repository already, but the sentiment there was that the component calling their `publish` methods should wrap it in a `try/catch`: https://github.com/bzikarsky/gelf-php/issues/67.

The unfortunate thing here is that the app and also running unit/integration tests would crash if the logging server is simply not available, because `gelf-php`'s `StreamSocketClient` tries to resolve the host in [here](https://github.com/bzikarsky/gelf-php/blob/bdf3c93d62d8e2896879c7a9e3f0df48e50fd436/src/Gelf/Transport/StreamSocketClient.php#L146).

Example stacktrace:
```
RuntimeException: Failed to create socket-client for udp://myloghostname.com:1515: php_network_getaddresses: getaddrinfo failed: Try again (0)

/app/vendor/graylog2/gelf-php/src/Gelf/Transport/StreamSocketClient.php:156
/app/vendor/graylog2/gelf-php/src/Gelf/Transport/StreamSocketClient.php:183
/app/vendor/graylog2/gelf-php/src/Gelf/Transport/StreamSocketClient.php:202
/app/vendor/graylog2/gelf-php/src/Gelf/Transport/UdpTransport.php:89
/app/vendor/graylog2/gelf-php/src/Gelf/Publisher.php:76
/app/vendor/monolog/monolog/src/Monolog/Handler/GelfHandler.php:63
/app/vendor/monolog/monolog/src/Monolog/Handler/AbstractProcessingHandler.php:37
/app/vendor/monolog/monolog/src/Monolog/Logger.php:337
/app/vendor/monolog/monolog/src/Monolog/Logger.php:532
```

Can we make handling possible exceptions in `GelfHandler->write()` or on a higher level `AbstractProcessingHandler->handle()` configurable?

Any ideas?

Seldaek:
[...]

PR #939, opened 2017-03-06
re-throws exception thrown by execute function 

Riser101:
We've observed that the runtime exception thrown by the function execute() in util.php does not propagate to created instances of logger or pushHandler. 

Because of this php threads break when curl requests to loggly hosts fail after max retries.
This issue is similar to #933


Riser101:
The failed check - is for php-nightly and I it's not relevant to the above change.

Seldaek:
I closed #933 with [an explanation](https://github.com/Seldaek/monolog/issues/933#issuecomment-285957643). I don't really understand what this PR is doing so please explain in more details what you are fixing?

PR #940, opened 2017-03-10
Handle DateTime objects in formatted messages

pierswarmers:
Adds support for DateTime objects being passed and formatted in PSRLogMessages.

Seems like it might be a good addition. Totally understand if this change pushes this processor beyond it's intended scope.

Cheers.

### DIFF
--- a/src/Monolog/Processor/PsrLogMessageProcessor.php
+++ b/src/Monolog/Processor/PsrLogMessageProcessor.php
@@ -22,7 +22,7 @@ class PsrLogMessageProcessor
 {
     const SIMPLE_DATE = "Y-m-d\TH:i:sP";
 
-    protected $dateFormat;
+    private $dateFormat;
 
     /**
      * @param string $dateFormat The format of the timestamp: one supported by DateTime::format
--- a/src/Monolog/Processor/PsrLogMessageProcessor.php
+++ b/src/Monolog/Processor/PsrLogMessageProcessor.php
@@ -20,6 +20,18 @@ namespace Monolog\Processor;
  */
 class PsrLogMessageProcessor
 {
+    const SIMPLE_DATE = "Y-m-d\TH:i:sP";
+
+    protected $dateFormat;
+
+    /**
+     * @param string $dateFormat The format of the timestamp: one supported by DateTime::format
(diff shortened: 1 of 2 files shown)


src/Monolog/Processor/PsrLogMessageProcessor.php
```diff
@@ -33,7 +45,9 @@ public function __invoke(array $record): array
         $replacements = [];
         foreach ($record['context'] as $key => $val) {
             if (is_null($val) || is_scalar($val) || (is_object($val) && method_exists($val, "__toString"))) {
-                $replacements['{'.$key.'}'] = $val;
+                $replacements['{' . $key . '}'] = $val;
+            } elseif ($val instanceof \DateTime) {
```
stof: should check for DateTimeInterface too, to support DateTimeImmutable

src/Monolog/Processor/PsrLogMessageProcessor.php
```diff
@@ -33,7 +45,9 @@ public function __invoke(array $record): array
         $replacements = [];
         foreach ($record['context'] as $key => $val) {
             if (is_null($val) || is_scalar($val) || (is_object($val) && method_exists($val, "__toString"))) {
-                $replacements['{'.$key.'}'] = $val;
+                $replacements['{' . $key . '}'] = $val;
```
stof: why are you changing coding standards ?

src/Monolog/Processor/PsrLogMessageProcessor.php
```diff
@@ -20,6 +20,18 @@
  */
 class PsrLogMessageProcessor
 {
+    const SIMPLE_DATE = "Y-m-d\TH:i:sP";
+
+    protected $dateFormat;
```
stof: should be private IMO

Seldaek:
Seems fine to me but yes as master target php7 this should be checking for DateTimeInterface. Please fix the other comments from @stof too, then it should be mergeable :)

pierswarmers:
Thanks for the feedback! Changes have been made :)

Seldaek:
:+1: Thanks

PR #947, opened 2017-03-17
Update Logger.php

adjenks:
Update Logger class to accept a callback function to handle exceptions in log Handlers.
Logs to error_log by default in order to capture failures in Handlers
Prevents the entire program from failing and losing all logs if a single log throws an exception.


Seldaek:
Sorry but nope, use WhatFailureGroupHandler to wrap your handlers if you want to ignore exceptions.

Issue #852, opened 2016-09-14
RollbarHandler: Critical message reduced to error when exception included


AlexStansfield:
When there is an exception in the record context the RollbarHandler uses `RollbarNotifier:: report_exception` to log the error.

The problem is this doesn't pass on the level of the message to rollbar (there is no level argument on `report_exception`). The RollbarNotifier then sets it at a default level of `error`. 

This means no matter what level you log the exception at it will always show up as level `error` in rollbar.

Symfony's own ExceptionListener logs exceptions as `critical` when they are not instances of `HttpExceptionInterface` or have a status code great than or equal to 500.

A quick solution (I think) to this would be to add:

```
$payload['level'] = strtolower($record['level_name']);
```

prior to calling the `report_exception` method ([line 70](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/RollbarHandler.php#L70))

Seldaek:
So I looked into the RollbarHandler and found out that it was kinda suboptimal, so I did some tweaking in there and you should now get the level and some other stuff in the context data, no matter if logging an exception or a regular message. Let me know if it helps.

PR #948, opened 2017-03-18
Add level parameter to payload when reporting Rollbar exceptions (#852)

knugen:
As mentioned by @AlexStansfield in #852, the level is not properly reflected in Rollbar when reporting exceptions. The issue is closed but it still didn't work for us.

So while digging into the Rollbar library I arrived at the same conclusion that Alex did; that the level has to be included in the payload data to work.

I have verified that this fix works with manual tests. I also took the opportunity to introduce a basic test case for the RollbarHandler as there wasn't one before.


Seldaek:
Thanks, I'll backport the fix to 1.x as well.

PR #949, opened 2017-03-21
Replace deprecated setExpectedException in PHPUnit tests.

chrisforrence:
The `setExpectedException` method is deprecated in PHPUnit 5.2+ and removed in PHPUnit 6. This pull request simply replaces calls to `setExpectedException` with its replacement `expectException` and `expectExceptionMessage` (as needed).


stof:
You should detect whether ``expectException`` exists, and fallback to `setExpectedException` otherwise, to keep compat with PHPUnit 4.8 to run tests on PHP 5.3 (and you should do it in 1.x, as it is still the main active branch for now)

chrisforrence:
Just to clarify, are you suggesting that I check that so that the tests can essentially be copied from 1.x to master without needing to swap out `expectException` and `setExpectedException`?

Seldaek:
Thanks, think it's fine for now to only do this in master, it keeps the tests cleaner, and it's unlikely these tests will change a lot in 1.x..

Issue #954, opened 2017-03-27
E_USER_ERROR should not be in ErrorHandler::$fatalErrors


masterada:
Please remove E_USER_ERROR from the list of fatal errors as it's not a fatal error.
Pull request: https://github.com/Seldaek/monolog/pull/956

sfatula:
Since when? Run this simple php program:

```
<?php
trigger_error("Test", E_USER_ERROR);
echo "Do I get here\n";
```

Do you see an error, or, the echo? On all php versions I have, I see a fatal error. E_USER_ERROR has always been a fatal error. I know I saw a php page where it did not list it as fatal, but it is. You will break a lot of code if you change this!

masterada:
That's not completely true:

```
<?php
set_error_handler('handleError');
function handleError($errno, $errstr, $errfile, $errline)
{
	return;
}

trigger_error("Test", E_USER_ERROR);
echo "Do I get here\n";
```
This will print Do I get here.


```
<?php
set_error_handler('handleError');
function handleError($errno, $errstr, $errfile, $errline)
{
	return false;
}

trigger_error("Test", E_USER_ERROR);
echo "Do I get here\n";
```
This will throw a fatal error.

So unless you return false from the error handler, E_USER_ERROR will not be fatal. 

The problem with current monolog is if you set $callPrevious false, you will never see E_USER_ERRORs, because they won't be logged by the error handler or the fatal handler.

sfatula:
Your example is speaking of error handlers. I am speaking of the simple 3 line file, which you have to agree *is* fatal in php, right? Without any added error handler, execution does stop. I was more picking on the statement that E_USER_ERROR is not fatal. Perhaps semantics.

Seldaek:
Yeah if you have a custom error handler that says it handled things to php, then php won't fatal, but that's also true of E_ERROR AFAIK, so sorry but I don't think this is right. If your error handler changes default semantics, then you should configure monolog differently or override it as well so it matches your project's semantics.

PR #956, opened 2017-03-28
Removed E_USER_ERROR from fatal errors, because it does not trigger a fatal

masterada:
Removed E_USER_ERROR from fatal errors, because it does not trigger a fatal


Seldaek:
Closing as per https://github.com/Seldaek/monolog/issues/954#issuecomment-309304142

PR #958, opened 2017-03-28
Allow usage of elasticsearch client

martijncalker:
Proposal for using both `runflin/elastica` and `elasticsearch/elasticsearch` clients in the ElasticSearchHandler. 

Not very sure if this is a nice way, looking forward for some feedback :)

### DIFF
--- a/src/Monolog/Handler/ElasticSearchHandler.php
+++ b/src/Monolog/Handler/ElasticSearchHandler.php
@@ -11,10 +11,11 @@
 
 namespace Monolog\Handler;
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks but I think it'd make more sense to add a separate handler for the elasticsearch lib than having ifs everywhere for typechecking. Problem is the name is kinda conflicting.. so maybe best to publish it as a standalone package instead and list it at https://github.com/seldaek/monolog/wiki/Third-Party-Packages

Issue #913, opened 2017-01-10
LogstashFormatter record[level_name] is is mapped to message[level], record[level] is not mapped at all


pupsschluepfer:
Shouldn't a mapping from monolog-record to logstash-message be as accurate as possible?

```php
// ...
if (isset($record['level_name'])) {
    $message['level_name'] = $record['level_name'];
}
if (isset($record['level'])) {
    $message['level'] = $record['level'];
}
// ...
```

At the moment the content of `$record['level_name']` is just written to `$message['level']`. IMHO this might be a little misleading because most loggin-systems consider `level` an interger and `level_name` a human-readable string.

Seldaek:
This was done when logstash V1 was added, before we used the level alone for V0.. I guess the problem is that while a lot of systems treat levels as integers, the exact numbers used vary. Names are at least non ambiguous. And well changing this at this point would break things for existing users that built tooling around these messages, so not sure it's worth it..

pupsschluepfer:
Concerning existing solutions, you're right of course.
Maybe it would be an option to just map the (int)$record[level] to some extra field in $message so that this value is at least present in the message, too.

Seldaek:
Sure that'd be an option.

pupsschluepfer:
How can I add a pull-request?

Seldaek:
https://help.github.com/articles/creating-a-pull-request/

PR #960, opened 2017-03-28
map loglevel (int-value) to log-message, as well

pupsschluepfer:
as mentioned in #913 just add the integer-value of level to an additional field in the log-message.


Seldaek:
Thanks, merged but I renamed the field to monolog_level to emphasize that it is monolog specific numbers in there

pupsschluepfer:
Alright, thank you very much!

PR #965, opened 2017-04-01
Fix build

Sam-Burns:
Fixes the Monolog build, which is currently having two problems:
* A [bug in PHP 7.1.3](https://bugs.php.net/bug.php?id=74258) means microseconds in DateTimes don't work
* The `php nightly` build on Travis is failing ATOW, because it is seg faulting while running Composer.


Seldaek:
Cheers

PR #966, opened 2017-04-04
Support HeadlessChrome

redthor:
We've started using Chromium 57 with the `--headless` flag during integration tests. This means we no longer need Xvfb to be installed.

However Chromium changes the Agent string to:
`'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome Safari/537.36'`
this doesn't match ChomeLogger's REGEX so our int tests won't get the Chrome logger data.

Can we have this? We are still on PHP 5.6 so I haven't been able to test against `master`.

Thanks.

### DIFF
--- a/src/Monolog/Handler/ChromePHPHandler.php
+++ b/src/Monolog/Handler/ChromePHPHandler.php
@@ -36,7 +36,7 @@ class ChromePHPHandler extends AbstractProcessingHandler
     /**
      * Regular expression to detect supported browsers (matches any Chrome, or Firefox 43+)
      */
-    const USER_AGENT_REGEX = '{\b(?:Chrome\/\d+(?:\.\d+)*|HeadlessChrome|Firefox\/(?:4[3-9]|[5-9]\d|\d{3,})(?:\.\d)*)\b}';
+    const USER_AGENT_REGEX = '{\b(?:Chrome/\d+(?:\.\d+)*|HeadlessChrome|Firefox/(?:4[3-9]|[5-9]\d|\d{3,})(?:\.\d)*)\b}';
 
     protected static $initialized = false;
 
--- a/src/Monolog/Handler/ChromePHPHandler.php
+++ b/src/Monolog/Handler/ChromePHPHandler.php
@@ -36,7 +36,7 @@ class ChromePHPHandler extends AbstractProcessingHandler
     /**
      * Regular expression to detect supported browsers (matches any Chrome, or Firefox 43+)
      */
-    const USER_AGENT_REGEX = '{\b(?:Chrome/\d+(?:\.\d+)*|Firefox/(?:4[3-9]|[5-9]\d|\d{3,})(?:\.\d)*)\b}';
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/ChromePHPHandler.php
```diff
@@ -36,7 +36,7 @@ class ChromePHPHandler extends AbstractProcessingHandler
     /**
      * Regular expression to detect supported browsers (matches any Chrome, or Firefox 43+)
      */
-    const USER_AGENT_REGEX = '{\b(?:Chrome/\d+(?:\.\d+)*|Firefox/(?:4[3-9]|[5-9]\d|\d{3,})(?:\.\d)*)\b}';
+    const USER_AGENT_REGEX = '{\b(?:Chrome\/\d+(?:\.\d+)*|HeadlessChrome|Firefox\/(?:4[3-9]|[5-9]\d|\d{3,})(?:\.\d)*)\b}';
```
stof: there is no need to escape slashes here. They are not special chars in the regex (as they are not the delimiter

tests/Monolog/Handler/ChromePHPHandlerTest.php
```diff
@@ -47,6 +54,16 @@ public function testHeaders()
         $this->assertEquals($expected, $handler->getHeaders());
     }
 
+    public function agentsProvider()
+    {
+        return [
```
stof: the 1.x branch supports PHP 5.3, so you cannot use the short array syntax

tests/Monolog/Handler/ChromePHPHandlerTest.php
```diff
@@ -25,8 +25,15 @@ protected function setUp()
         $_SERVER['HTTP_USER_AGENT'] = 'Monolog Test; Chrome/1.0';
     }
 
-    public function testHeaders()
+    /**
+     * testHeaders
```
stof: this short description is totally useless. Please remove it.

tests/Monolog/Handler/ChromePHPHandlerTest.php
```diff
@@ -47,6 +54,16 @@ public function testHeaders()
         $this->assertEquals($expected, $handler->getHeaders());
     }
 
+    public function agentsProvider()
```
stof: I suggest using a static method, which avoids PHPUnit to instantiate the TestCase when running the provider during the loading of the testsuite

stof:
it is weird that the headless chrome does not display its version at all in the user agent though.

redthor:
Thanks for the comments. Yes, the agent string isn't great :/

redthor:
Do you think we can have this? The are two symfony changes we've made. The first one will fix our problem because the regex is overridden by the monolog bridge. But, the second change (which includes increasing the monolog version) will *re-instroduce* the issue if it is merged, and this one stays open because the second change will inherit the regex from this library instead of overriding it.

redthor:
Hi @stof , the symfony monolog bridge request to inherit the regex has been merged into v3.30-BETA1

Can we have this merged so the issue does not simply re-emerge when we get v3.3 ?

ref: https://github.com/symfony/symfony/pull/22603

Thanks

Seldaek:
Thanks

PR #970, opened 2017-04-12
improve docblocks and AWS DynamoDbHandler

irfanevrens:


Seldaek:
Thanks I carried over some of the changes.. a lot of them don't make sense IMO, adding docblocks to repeat what type declarations have is useless, and dropping Aws 2.x support is not really needed.

PR #972, opened 2017-04-12
Better PSR-3 message formatting

lstrojny:
Does two things:

 - Include milliseconds in the default date format
 - Show actual content of the array

### DIFF
--- a/src/Monolog/Processor/PsrLogMessageProcessor.php
+++ b/src/Monolog/Processor/PsrLogMessageProcessor.php
@@ -20,7 +20,7 @@ namespace Monolog\Processor;
  */
 class PsrLogMessageProcessor
 {
-    const SIMPLE_DATE = "Y-m-d\TH:i:sP";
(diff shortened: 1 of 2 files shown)


src/Monolog/Processor/PsrLogMessageProcessor.php
```diff
@@ -51,7 +51,7 @@ public function __invoke(array $record): array
             } elseif (is_object($val)) {
                 $replacements['{'.$key.'}'] = '[object '.get_class($val).']';
             } else {
-                $replacements['{'.$key.'}'] = '['.gettype($val).']';
+                $replacements['{'.$key.'}'] = 'array'.@json_encode($val);
```
Seldaek: We really don't know that it is an array there? Could be a resource, or maybe other stuff I can't think of right now because it's late :p

Seldaek: Feel free to add an elseif is_array() above and do this, but this last one should remain the same IMO

PR #976, opened 2017-04-26
LogstashFormatter: format context and extra fields as JSON objects

### LINKED ISSUE #975: LogstashFormatter: format context and extra fields as JSON objects
The current way of prefixing `context` and `extra` fields is just weird. Monolog is the first library I've seen doing something like that.

It should just send `context` and `extra` objects (associative arrays).

So instead of:

```
"context.username": "foo"
"context.ip_address": "127.0.0.1"
```
it should be:
```
"context": {
  "username": "foo",
  "ip_address": "127.0.0.1"
}
```

elnur:
Fixes #975.

I'll fix some other problems with this formatter after this PR gets merged.

### DIFF
--- a/src/Monolog/Formatter/LogstashFormatter.php
+++ b/src/Monolog/Formatter/LogstashFormatter.php
@@ -87,9 +87,7 @@ class LogstashFormatter extends NormalizerFormatter
             $message['type'] = $this->applicationName;
         }
         if (!empty($record['extra'])) {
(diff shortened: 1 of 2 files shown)


elnur:
It's a BC break, BTW. Hence 2.0.

elnur:
@Seldaek, ping, man. 😉

Seldaek:
Thanks

PR #979, opened 2017-05-03
Update JsonFormatter PHPDoc.

Aliance:
I decide to add missing param for constructor PHPDoc. But did not sure because of missing PHPDoc in master.


moderndeveloperllc:
@Seldaek Do you like to put something like `(Optional) Default to true` into the description of the parameter for things like this?
```
@param bool $appendNewline (Optional) Default to true
```

Seldaek:
@moderndeveloperllc no I generally don't like repeating what's in the code as it's redundant at best and misleading at worst (when you change the code and forget the docs)

@Aliance thanks

PR #986, opened 2017-05-05
PHP Versions less than 5.3.6 throws undefined constant DEBUG_BACKTRAC…

brantmessenger:
…E_IGNORE_ARGS

Notice: Use of undefined constant DEBUG_BACKTRACE_IGNORE_ARGS - assumed 'DEBUG_BACKTRACE_IGNORE_ARGS' in /vendor/monolog/monolog/src/Monolog/Processor/IntrospectionProcessor.php on line 58

### DIFF
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -60,10 +60,7 @@ class IntrospectionProcessor
         * As of 5.3.6, DEBUG_BACKTRACE_IGNORE_ARGS option was added.
         * Any version less than 5.3.6 must use the DEBUG_BACKTRACE_IGNORE_ARGS constant value '2'.
         */
-        if (!defined('DEBUG_BACKTRACE_IGNORE_ARGS')) {
-            define('DEBUG_BACKTRACE_IGNORE_ARGS', 2);   
-        }
-        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        $trace = debug_backtrace((PHP_VERSION_ID < 50306) ? 2 : DEBUG_BACKTRACE_IGNORE_ARGS);
 
         // skip first since it's always the current method
         array_shift($trace);
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -55,7 +55,7 @@ class IntrospectionProcessor
             return $record;
         }
 
-        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        $trace = debug_backtrace((version_compare(PHP_VERSION, '5.3.6') < 0) ? 2 : DEBUG_BACKTRACE_IGNORE_ARGS);
 
         // skip first since it's always the current method
         array_shift($trace);
--- a/src/Monolog/Processor/IntrospectionProcessor.php
+++ b/src/Monolog/Processor/IntrospectionProcessor.php
@@ -55,7 +55,15 @@ class IntrospectionProcessor
             return $record;
         }
 
-        $trace = debug_backtrace((version_compare(PHP_VERSION, '5.3.6') < 0) ? 2 : DEBUG_BACKTRACE_IGNORE_ARGS);
+        /*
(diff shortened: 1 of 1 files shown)


src/Monolog/Processor/IntrospectionProcessor.php
```diff
@@ -55,7 +55,7 @@ public function __invoke(array $record)
             return $record;
         }
 
-        $trace = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS);
+        $trace = debug_backtrace((version_compare(PHP_VERSION, '5.3.6') < 0) ? 2 : DEBUG_BACKTRACE_IGNORE_ARGS);
```
xabbuh: you can use the `PHP_VERSION_ID` constant here to avoid the function call: `\PHP_VERSION_ID >= 50306 ? DEBUG_BACKTRACE_IGNORE_ARGS : 0`

brantmessenger: I'm ok with using the (int) PHP_VERSION_ID.

`(PHP_VERSION_ID >= 50306) ? DEBUG_BACKTRACE_IGNORE_ARGS : 2`

src/Monolog/Processor/IntrospectionProcessor.php
```diff
@@ -55,6 +55,14 @@ public function __invoke(array $record)
             return $record;
         }
 
+        /*
+        * http://php.net/manual/en/function.debug-backtrace.php
+        * As of 5.3.6, DEBUG_BACKTRACE_IGNORE_ARGS option was added.
+        * Any version less than 5.3.6 must use the DEBUG_BACKTRACE_IGNORE_ARGS constant value '2'.
+        */
+        if (!defined('DEBUG_BACKTRACE_IGNORE_ARGS')) {
+            define('DEBUG_BACKTRACE_IGNORE_ARGS', 2);
```
xabbuh: I would not define internal constants. Doing so may break other libraries which use this constant for version/feature detection.

brantmessenger: A little history... We are forced to use version PHP 5.3.3. Because of this we set in the composer.json file the platform like so...
`    "config": {
        "platform": {
            "php": "5.3.3"
        }
    },`
...which resulted in monolog "version": "1.22.1" getting pullled in. Monolog states that it requires...
`            "require": {
                "php": ">=5.3.0",
                "psr/log": "~1.0"
            },`
...but DEBUG_BACKTRACE_IGNORE_ARGS is not defined until PHP 5.3.6.

Since this really is a PHP version issue and PHP version less than 5.2.0 would create an undefined constant error for PHP_VERSION_ID. I thought trying to simplify the process and only address the issue at hand which is an "undefined constant".

DEBUG_BACKTRACE_IGNORE_ARGS is not defined prior to PHP 5.3.6. So, DEBUG_BACKTRACE_IGNORE_ARGS would not be defined for any other libraries prior to PHP 5.3.6.

For PHP versions greater than or equal to 5.3.6 the constant "DEBUG_BACKTRACE_IGNORE_ARGS" is defined and would not impact other libraries that use the DEBUG_BACKTRACE_IGNORE_ARGS constant.

In addition, DEBUG_BACKTRACE_IGNORE_ARGS has a value of (int) 2 when it is defined either by PHP or through this code.

brantmessenger: This bug was introduced...
https://github.com/Seldaek/monolog/commit/cf2faeb3b73144666bb0e819b72b5bae977a9978

xabbuh: I am not arguing against fixing compat for PHP < 5.3.6. I just would not define the constant here as other libraries may use the presence of this constant to conditionally execute some code. That's why I said I would check for the value of the `PHP_VERSION_ID` constant instead.

brantmessenger: Ok... I'll go with round 3. ;)

Seldaek:
Thanks

brantmessenger:
Glad I could help! Unfortunately, I'm forced to use older versions of PHP which result in coming across issues that generally don't present themselves to others. Hopefully this helps someone else in the same situation.

PR #993, opened 2017-05-26
Update SwiftMailerHandler.php

### LINKED ISSUE #992: Swift_Mime_SimpleMessage class "setDate" method requires DateTimeInterface
Trying to use the SwiftMailerHandler results in an error...
It's because the Swift Message class type checks the input to set date which is currently a timestamp.
Instead of a timestamp, it needs to be:
```php
->setDate(new DateTimeImmutable())
```

bo-brainlabs:
fix #992

### DIFF
--- a/src/Monolog/Handler/SwiftMailerHandler.php
+++ b/src/Monolog/Handler/SwiftMailerHandler.php
@@ -79,7 +79,7 @@ class SwiftMailerHandler extends MailHandler
         }
 
         $message->setBody($content, $mime);
-        $message->setDate(time());
+        $message->setDate(new DateTimeImmutable());
 
         return $message;
     }


src/Monolog/Handler/SwiftMailerHandler.php
```diff
@@ -79,7 +79,7 @@ protected function buildMessage(string $content, array $records): Swift_Message
         }
 
         $message->setBody($content, $mime);
-        $message->setDate(time());
+        $message->setDate(new DateTimeImmutable());
```
moderndeveloperllc: This needs a slash beforehand to import from the core namespace. If you run the tests before you commit, you will see:
```
Error: Class 'Monolog\Handler\DateTimeImmutable' not found
```

stof:
this now supports Swiftmailer 6.x only. It would be better to write code supporting both 5.x and 6.x, as there is no composer requirement enforcing usage of Swiftmailer 6.x (and I don't think it is a good idea to add a conflict rule, as it would forbid people using Swiftmailer 6.x from using an uptodate version of Monolog).

Also, this should be fixed in Monolog 1.x branch actually

bo-brainlabs:
Ah thanks, I will check it out!

PR #995, opened 2017-05-31
Minor code cleanup on Handler/ErrorLogHandler.php

wshafer:
Just some cleanup for readability.

### DIFF
--- a/src/Monolog/Handler/ErrorLogHandler.php
+++ b/src/Monolog/Handler/ErrorLogHandler.php
@@ -71,13 +71,14 @@ class ErrorLogHandler extends AbstractProcessingHandler
      */
     protected function write(array $record)
     {
-        if ($this->expandNewlines) {
-            $lines = preg_split('{[\r\n]+}', (string) $record['formatted']);
-            foreach ($lines as $line) {
-                error_log($line, $this->messageType);
-            }
-        } else {
+        if (!$this->expandNewlines) {
             error_log((string) $record['formatted'], $this->messageType);
+            return;
+        } 
+        
+        $lines = preg_split('{[\r\n]+}', (string) $record['formatted']);
+        foreach ($lines as $line) {
+            error_log($line, $this->messageType);
         }
     }
 }


src/Monolog/Handler/ErrorLogHandler.php
```diff
@@ -71,13 +71,14 @@ protected function getDefaultFormatter(): FormatterInterface
      */
     protected function write(array $record)
     {
-        if ($this->expandNewlines) {
-            $lines = preg_split('{[\r\n]+}', (string) $record['formatted']);
-            foreach ($lines as $line) {
-                error_log($line, $this->messageType);
-            }
-        } else {
+        if (!$this->expandNewlines) {
```
HonzaMac: Why do we need to use Negative condition instead of truthy?

wshafer: This makes the code a bit easier to read and removes the else (which is not needed at all here).

The other way would put the whole function in an additional un-needed code block.

wshafer: Here's how it would look the other way:

```php
protected function write(array $record)
    {
        if ($this->expandNewlines) {
            $lines = preg_split('{[\r\n]+}', (string) $record['formatted']);
            foreach ($lines as $line) {. // Nested foreach
               error_log($line, $this->messageType);
            }
            return;
        } 
        error_log((string) $record['formatted'], $this->messageType);
    }
```

You'll notice all the code is nested inside an if block.  In addition we now have a nested foreach block also nested.

HonzaMac: I'm still not getting ,why falsy condition is better, when you still needs to handle both cases, truthy and falsy.

wshafer: It's all about readability.  Logically they are the same.

In the same way you see no value in the negititive, I see no way in this section of code how truths is better.  In this case we have blocks within blocks within blocks.   Not only does it make it hard to read it's also harder to follow.   If you exit out early you don't need to worry about such things. 

If the bulk of code fell inside the negitive block of code, I'd have submitted a PR in the reverse. 

But this isn't my project, just thought I'd pass this along while I was in there.

Seldaek:
Thanks

PR #997, opened 2017-05-31
Added a new ResettableInterface and implemented it where possible.

lyrixx:
Added a new ResetableInterface and implemented it where possible.

When one use Monolog in a long process like an AMQP worker with a
`FingersCrossedHandler` or `BufferHandler` there is a drawback: as soon as there
is an AMQP message that generate a log >= error (for example), all next AMQP
messages will output logs, even if theses messages don't generate log where
level >= error.

In the same context there is a drawback for processor that add an UUID to the
logs. The UUID should change for each AMQP messages.

---

This patch address this issue with a new interface: `ResettableInterface` interface.
Side note: `reset()`, `flush()`, `clear()`,  are already used in Monolog. So
basically, one can use the `reset()` on the `Logger` and on some
`Handler`s / `Processor`s.

It's especially useful for

* the `FingersCrossedHandler`: it `close()` the buffer, then it `clear()` the buffer.
* the `BufferHandler`: it `flush()` the buf

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -25,7 +25,7 @@ use Exception;
  *
  * @author Jordi Boggiano <j.boggiano@seld.be>
  */
-class Logger implements LoggerInterface
+class Logger implements LoggerInterface, ResettableInterface
 {
     /**
      * Detailed debug information
@@ -352,6 +352,21 @@ class Logger implements LoggerInterface
         }
 
         return true;
+    }
+
+    public function reset()
+    {
+        foreach ($this->handlers as $handler) {
+            if ($handler instanceof ResettableInterface) {
+                $handler->reset();
+            }
+        }
+
+        foreach ($this->processors as $processor) {
+            if ($processor instanceof ResettableInterface) {
+                $processor->reset();
+            }
+        }
     }
 
     /**
--- a/src/Monolog/Processor/UidProcessor.php
+++ b/src/Monolog/Processor/UidProcessor.php
@@ -11,12 +11,14 @@
 
 namespace Monolog\Processor;
 
+use Monolog\ResettableInterface;
+
 /**
  * Adds a unique identifier into records
  *
  * @author Simon Mönch <sm@webfactory.de>
  */
-class UidProcessor
+class UidProcessor implements ResettableInterface
 {
     private $uid;
 
@@ -26,7 +28,8 @@ class UidProcessor
             throw new \InvalidArgumentException('The uid length must be an integer between 1 and 32');
         }
 
-        $this->uid = substr(hash('md5', uniqid('', true)), 0, $length);
+
+        $this->uid = $this->generateUid($length);
     }
 
     public function __invoke(array $record)
@@ -42,5 +45,15 @@ class UidProcessor
     public function getUid()
     {
         return $this->uid;
+    }
+
+    public function reset()
+    {
+        $this->uid = $this->generateUid(strlen($this->uid));
+    }
+
+    private function generateUid($length)
+    {
+        return substr(hash('md5', uniqid('', true)), 0, $length);
     }
 }
--- a/src/Monolog/ResettableInterface.php
+++ b/src/Monolog/ResettableInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog;
+
+/**
+ * Handler or Processor implementing this interface will be reset when Logger::reset() is called.
+ *
+ * Resetting an Handler or a Processor usually means cleaning all buffers or
+ * resetting in its internal state.
+ *
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ */
+interface ResettableInterface
+{
+    public function reset();
+}
--- a/tests/Monolog/Handler/FirePHPHandlerTest.php
+++ b/tests/Monolog/Handler/FirePHPHandlerTest.php
@@ -21,7 +21,7 @@ class FirePHPHandlerTest extends TestCase
 {
     public function setUp()
     {
-        TestFirePHPHandler::reset();
+        TestFirePHPHandler::resetStatic();
         $_SERVER['HTTP_USER_AGENT'] = 'Monolog Test; FirePHP/1.0';
     }
 
@@ -77,7 +77,7 @@ class TestFirePHPHandler extends FirePHPHandler
 {
     protected $headers = array();
 
-    public static function reset()
+    public static function resetStatic()
     {
         self::$initialized = false;
         self::$sendHeaders = true;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   * Fixed table row styling issues in HtmlFormatter
   * Fixed RavenHandler dropping the message when logging exception
   * Fixed WhatFailureGroupHandler skipping processors when using handleBatch
+  * Added a `ResettableInterface` in order to reset/reset/clear/flush handlers and processors
+    and implement it where possible
 
 ### 1.23.0 (2017-06-19)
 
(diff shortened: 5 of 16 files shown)


lyrixx:
I forgot to say that we are using this "feature" for years at SensioLabs. But in a very hacky way (with reflection). IMHO, it's better to expose a clean API. That's why I created this PR.

theofidry:
Fortunately in our case such logging is done only when an error occurs, so we can afford to kill the process to circumvent that issue. A clean API to do it would avoid this "hack" or having to rely on reflection 👍

lyrixx:
This PR needs a rebase, but before finishing it I would like to be sure If it will/could be merged.

More over Symfony get a new component: Messenger. So (I hope so) AMQP (or other queue system) will become more standard. So people will expects a nice logging system.

ping @fabpot @nicolas-grekas @sroze 

Next step: wire this restart method after each message processed in the Messenger Component

src/Monolog/Processor/UidProcessor.php
```diff
@@ -43,4 +45,9 @@ public function getUid(): string
     {
         return $this->uid;
     }
+
+    public function restart(int $length = 7)
```
nicolas-grekas: I'd suggest to store the length as a property instead of adding in here

nicolas-grekas:
> Next step: wire this restart method after each message processed in the Messenger Component

and on container reset

stof:
Why is it in the 2.0 milestone rather than adding this new feature in 1.x (2.0 is kind of stale, and Symfony will have a hard time moving to it due to the non-continuous migration path for handler authors currently)

src/Monolog/Logger.php
```diff
@@ -325,6 +325,21 @@ public function addRecord(int $level, string $message, array $context = []): boo
         return true;
     }
 
+    public function restart($a = null)
```
stof: why having an argument here ?

src/Monolog/Processor/UidProcessor.php
stof: agreed. None of the APIs are calling it with a value anyway, and we want it to be called with the original config.

src/Monolog/Processor/UidProcessor.php
```diff
@@ -43,4 +45,9 @@ public function getUid(): string
...
+    {
+        $this->__construct($length);
```
stof: I would rather move the uid generation to a private method and call it, rather than reusing `__construct` (which will be necessary anyway when applying the suggestions). I don't want to push people to think that calling `__construct` as a method is a supported usage of the library.

lyrixx:
> Why is it in the 2.0 milestone

I would prefer 1.X yes.

stof:
Well, I think the 2.0 milestone is because this was submitted to master.

lyrixx:
Thanks. I have addressed your comments and now this PR targets the 1.X branch.

src/Monolog/RestartableInterface.php
```diff
@@ -0,0 +1,25 @@
+<?php declare(strict_types=1);
```
sroze: Is that OK even if Monolog's 1.x's target is PHP 5.3 ?

lyrixx: oups ;)

lyrixx:
There is a failure on HHVM, but for another test. So this PR is ready.

fabpot:
I would recommend to drop HHVM support as almost all project have done by now (even on 1.x).

nicolas-grekas:
s/RestartableInterface::restart()/ResettableInteraface::reset()/?
then later one add the Symfony interface on the classes here (or as a child of the Symfony interface?)

lyrixx:
Hmm, `reset` is already used in [monolog](https://github.com/Seldaek/monolog/blob/07b6c1c4171a030787f2bc961272b1887032304e/src/Monolog/Handler/FingersCrossedHandler.php#L150) with a different meanings.

IIRC, this is why I choose `restart` instead of `reset`

nicolas-grekas:
a bridge/proxy would do it also

Seldaek:
Hey, this is something I wanted to support in 2.0, as well as having a Logger::close() to be able to close all handlers (as 2.0 has close in the HandlerInterface).

I am not against supporting this in 1.x though, but I wonder if we shouldn't call this `reset` rather than `restart`. The only name conflict is in the FingersCrossedHandler but really the only reason it has a reset() is to do exactly the same thing as this PR intends, so I don't think it's a problem at all, it just generalizes the concept to the rest of Monolog.

lyrixx:
I updated the PR to rename `ResartableInterface` to `ResetableInterface`

nicolas-grekas:
s/ResetableInterface/ResettableInterface

lyrixx:
@nicolas-grekas Thanks ; done

src/Monolog/ResettableInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog;
+
+/**
+ * Handler or Processor implementing this interface will be reseted when Logger::reset is called.
+ *
+ * Reseting an Handler or a Processor usually means cleaning all buffers or
+ * reseting in its internal state.
```
nicolas-grekas: double "t": resetted, resetting

tests/Monolog/Handler/FirePHPHandlerTest.php
```diff
@@ -77,7 +77,7 @@ class TestFirePHPHandler extends FirePHPHandler
 {
     protected $headers = array();
 
-    public static function reset()
+    public static function resetInternalState()
```
nicolas-grekas: resetStatic?

src/Monolog/ResettableInterface.php
```diff
@@ -0,0 +1,25 @@
...
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog;
+
+/**
+ * Handler or Processor implementing this interface will be reseted when Logger::reset is called.
```
nicolas-grekas: resetted ;)
Logger::reset()

lyrixx: I used reset instead of resetted

nicolas-grekas:
What do you think of https://github.com/symfony/symfony/pull/27093#discussion_r201032119?

image

Seldaek:
Thanks!

Seldaek:
Hey guys so just one more thing before I release this in 1.x.. The more I look at it especially in the master branch for 2.0 where I made close() part of the HandlerInterface, it seems to me like close() and reset() do essentially the same thing.. Or at least they should. Do you also see it like that or not? If so I am thinking to just remove close() in 2.0 and make HandlerInterface implement ResettableInterface. I think it'd simplify things, and it resolves what I had in mind for https://github.com/Seldaek/monolog/issues/197 to add Logger::close() to close all handlers as Logger::reset() was part of this PR.

nicolas-grekas:
If we want to identify a difference, I could see this one: reset is meant to put a service back in a reusable state. This does *not* necessarily imply closing connections to remote backends. Actually, when using the same kernel in a loop, having connections already open for next requests should be a significant part of why these apps are faster (ie no `PDO::__construct()` overhead).
Not sure how it applies here, so that's just for thoughts :)

Seldaek:
That makes sense yes, although I don't know how many handlers would actually benefit from this distinction. But then I need to change the implementation a bit, because right now reset() calls close(), so things are even more confusing. 

It seems to me like the main use case for reset() is fingers crossed and buffer handlers to make sure they're cleared in between two requests/jobs in long running apps.

The use case for close() is to do some cleanup at the very end, but also right now that's when we flush buffers and stuff like that, so I guess in the buffer case you'd want to flush on reset instead of close. And then close() would become mostly useless, except sometimes it calls reset() to flush, and in a few rare cases it closes some resource. Makes sense? Mostly thinking out loud here..

nicolas-grekas:
Makes sense to me yes :)

sroze:
Makes sense. Close becomes useless as it’s basically end of the process so
`reset` is enough :)

Seldaek:
OK see https://github.com/Seldaek/monolog/commit/e07c948042045c9acff160648db45e40f75b1c58 for changes/clarifications. Happy to get any feedback on this.

Seldaek:
https://github.com/Seldaek/monolog/commit/b80352368cd79c978a805a76b724618ccb4ba465 is the core of the changes for 1.x but I messed up the commit msg :s

PR #998, opened 2017-05-31
Fixed test of RotatingFileHandler when the current day is 31.

lyrixx:
This did not work because the current way to generate date is weak.
So now we choose the first day of each month. Like that there are no
more issue.

To better understand the issue:

```
php > dump(date('Y-m', gmmktime(0, 0, 0, 5 - 1, 31, 2017)));
"2017-05"
php > dump(date('Y-m', gmmktime(0, 0, 0, 5 - 2, 31, 2017)));
"2017-03"
php > dump(date('Y-m', gmmktime(0, 0, 0, 5 - 3, 31, 2017)));
"2017-03"
php > dump(date('Y-m', gmmktime(0, 0, 0, 5 - 4, 31, 2017)));
"2017-01"
```

### DIFF
--- a/tests/Monolog/Handler/RotatingFileHandlerTest.php
+++ b/tests/Monolog/Handler/RotatingFileHandlerTest.php
@@ -27,7 +27,7 @@ class RotatingFileHandlerTest extends TestCase
      */
     public $lastError;
 
-    public function setUp()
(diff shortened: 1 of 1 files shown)


lyrixx:
see #999

PR #1000, opened 2017-05-31
Remove extra line-break at end of line in LineFormatter()

ksandell:
Remove extra line-break at end of line in format(), so that when sent to ErrorLogHandler() and extra line-break does not appear between logged lines. This behavior is seen on both Windows and Linux (RedHat) systems with PHP v7.x.

Before change output in php error log is:
```txt
[31-May-2017 17:11:55 UTC] [NOTICE] test1 

[31-May-2017 17:11:55 UTC] [NOTICE] test2 {"a":"val"}

```
after change the output is:
```txt
[31-May-2017 17:11:55 UTC] [NOTICE] test1 
[31-May-2017 17:11:55 UTC] [NOTICE] test2 {"a":"val"}

### DIFF
--- a/src/Monolog/Formatter/LineFormatter.php
+++ b/src/Monolog/Formatter/LineFormatter.php
@@ -106,7 +106,7 @@ class LineFormatter extends NormalizerFormatter
             $output = preg_replace('/%(?:extra|context)\..+?%/', '', $output);
         }
 
-        return $output;
+        return rtrim($output,"\r\n");
     }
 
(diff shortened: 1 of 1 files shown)


Seldaek:
The new line is part of the format you configure.. if you don't want one, then pass in a format without newline. The default LineFormatter::SIMPLE_FORMAT contains  a newline but removing them with rtrim is not useful, it's a big BC break and will cause problems for many users.

PR #1001, opened 2017-05-31
Fix SwiftMailerHandler with Swiftmailer 6.0+

### LINKED ISSUE #992: Swift_Mime_SimpleMessage class "setDate" method requires DateTimeInterface

Pierstoval:
After #993, I think this workaround may be the best option for now.

For Swiftmailer <6, `Swift::VERSION` is equal to `@SWIFT_VERSION_NUMBER@` [as we can see here](https://github.com/swiftmailer/swiftmailer/blob/v5.4.8/lib/classes/Swift.php), so `version_compare()` will return false anyway.

This seems to have been fixed in Swiftmailer 6, [as we can see here](https://github.com/swiftmailer/swiftmailer/blob/v6.0.1/lib/classes/Swift.php), so this workaround should work.

This fixes #992 by the way, which I'm experiencing on one of my apps.

What do you think?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "doctrine/couchdb": "~1.0@dev",
         "aws/aws-sdk-php": "^2.4.9 || ^3.0",
         "php-amqplib/php-amqplib": "~2.4",
-        "swiftmailer/swiftmailer": "~5.3",
+        "swiftmailer/swiftmailer": "^5.3|^6.0",
         "php-console/php-console": "^3.1.3",
         "jakub-onderka/php-parallel-lint": "^0.9",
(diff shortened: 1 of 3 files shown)


Pierstoval:
Failing  tests don't seem to come from my PR 😕

Seldaek:
Merged as d84af8db3ca6d4432d95b1abb9f0542f1a17f4cb - Thanks

PR #1009, opened 2017-06-26
add handler and formatter for https://logz.io/

Chrico:
Added new handler and formatter for https://logz.io/

### DIFF
--- a/src/Monolog/Formatter/LogzIoFormatter.php
+++ b/src/Monolog/Formatter/LogzIoFormatter.php
@@ -1,0 +1,50 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Formatter;
+
+/**
+ * Encodes message information into JSON in a format compatible with Logz.io.
+ *
+ * @author Christian Brückner <chris@chrico.info>
+ */
+class LogzIoFormatter extends JsonFormatter
+{
+
+    /**
+     * yyyy-MM-dd'T'HH:mm:ss.SSSZ
+     */
+    const DATETIME_FORMAT = "c";
+
+    /**
+     * Overrides the default batch mode to new lines for compatibility with the Logz.io bulk API.
+     */
+    public function __construct(int $batchMode = self::BATCH_MODE_NEWLINES, bool $appendNewline = true)
+    {
+        parent::__construct($batchMode, $appendNewline);
+    }
+
+    /**
+     * Appends the '@timestamp' parameter for Logz.io.
+     *
+     * @see https://support.logz.io/hc/en-us/articles/210206885-How-can-I-get-Logz-io-to-read-the-timestamp-within-a-JSON-log-
+     * @see \Monolog\Formatter\JsonFormatter::format()
+     */
+    public function format(array $record): string
+    {
+        if (isset($record["datetime"]) && ($record["datetime"] instanceof \DateTimeInterface)) {
+            $record["@timestamp"] = $record["datetime"]->format(self::DATETIME_FORMAT);
+            unset($record["datetime"]);
+        }
+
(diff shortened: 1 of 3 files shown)


src/Monolog/Formatter/LogzIoFormatter.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @author Christian Brückner <chris@chrico.info>
+ */
+class LogzIoFormatter extends JsonFormatter
+{
+
+    /**
+     * yyyy-MM-dd'T'HH:mm:ss.SSSZ
+     */
+    const DATETIME_FORMAT = "c";
```
martijncalker: Why don't you include microseconds (i.g. `Y-m-d\TH:i:s.uP`)? Personally I prefer (auto-)ordering my log-trace based on time (especially in the ELK-stack). 

```php
$dt = new DateTime('now');

echo $dt->format('c');
echo $dt->format('Y-m-d\TH:i:s.uP');

----
2017-06-29T11:55:49+00:00
2017-06-29T11:55:49.763333+00:00
```

Chrico: Howdy @martijncalker !

Thanks for your feed. I was just using the format they're suggesting here: https://support.logz.io/hc/en-us/articles/210206885-How-can-I-get-Logz-io-to-read-the-timestamp-within-a-JSON-log- .

Shall i change the pull request to fit into your format? :-)

martijncalker: If logz.io use `format('c') `, you should use that I guess. I was a bit surprised that they don't support/use microseconds :)

Chrico: Howdy!

As far as i know, the current used time library for date parsing in Logstash and ElasticSearch are only supporting millisecond precision (http://www.joda.org/joda-time/). So there is no support for microseconds.
But as they mentioned in the article above, it's possible to add other date formats as well. Since this is a customization, it shouldn't be handled by the default Handler/Formatter in Monolog.

The [integration in Python](https://github.com/logzio/logzio-python-handler/blob/master/logzio/handler.py#L34) also uses `now.microsecond / 1000` (= milliseconds). Just the NodeJS-integration above has a additional field `@timestamp_nano` which is not offical supported by default.

Chrico:
Howdy @martijncalker . Is there anything a can do here? The pull request passes the travis tests and is implemented as described in documentation on logz.io. What's the timeline for merging new Handlers? :-)

Xymanek:
Hello, we need to use in our project. Is there a reason it isn't merged?

Rowern:
Hi, any news on this PR?

Rowern:
Would it be possible to add the hostname to the data sent?
The LogstashFormatter already has this feature and it would help us to search for logs from a specific host.

yakobe:
@Seldaek would it be possible to get a decision / opinion here so we can decide what to do? If it's not wanted in the standard library, maybe we can create a separate repo of monolog handlers? 

I see there are handlers for others such as Rollbar and NewRelic so it seems like it might be nice to have it in the standard.

Chrico:
@Rowern
> Would it be possible to add the hostname to the data sent?

This is not a standard field of logz.io, so i would suggest to add it via [WebProcessor](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Processor/WebProcessor.php).

Chrico:
@yakobe here we go: https://github.com/inpsyde/logzio-monolog have fun!

yakobe:
Thank you!

Seldaek:
Thanks but maintaining all handlers inside this repo is kinda unsustainable for me so please rather use the third party package instead of adding it in Monolog itself. You're welcome to add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages as well for discovery.

Chrico:
Thanks @Seldaek , that's all i wanted: feedback. ;-) I've added the package https://github.com/inpsyde/logzio-monolog to the "handlers" section. 👍

Seldaek:
Yeah sorry for the delay on responding I kinda took a long break from dealing with issues here. Trying to get back on track.

Issue #946, opened 2017-03-17
WhatFailureGroupHandler not handling bubbling.


adjenks:
Bubbling doesn't appear to be working, I tried changing the order of handlers in the array passed to the WhatFailureGroupHandler, but it doesn't seem to affect it, I also tried setting bubble to true and false. No combination of settings seem to prevent bubbling. I also tried setting bubbling on the actual WhatFailureGroupHandler in addition to the handlers it contains. How do I stop bubbling with this handler?

Seldaek:
As far as I can see in the code it does the right thing, so not sure what you mean here. If you set bubble to false in the constructor, and any of its handlers is handling a given message, it should return true in handle() which stops the bubbling chain.

I'll need more details if you want help on this one I think.

adjenks:
If you set bubble to false on sub-handler in the WhatFailureGroupHandler array, they do not trigger bubbling to stop.
This is because WhatFailureGroupHandler simply iterates over the handers and doesn't check their bubble status, it simply calls all of them like so:
```
        foreach ($this->handlers as $handler) {
            try {
                $handler->handle($record);
            } catch (\Throwable $e) {
                // What failure?
            }
        }
```

After discovering this, I was going to modify WhatFailureGroupHandler to do something like this:

```
        foreach ($this->handlers as $handler) {
            try {
                if(true == $handler->handle($record)){
                    break;
                }
            } catch (\Throwable $e) {
                // What failure?
            }
        }
```

Similar to how bubbling is controlled in Logger. However I realized that I would rather just be able to let Logger not die, since logging is important and WhatFailureGroupHandler seemed redundant, which is why I made that fork and pull request.

So as you can see above, WhatFailureGroupHandler simply calls all of its handlers, and ignores bubbling.
It is true that the WhatFailureGroupHandler does obey bubbling itself, if it's own bubble is set to false, it will stop any logging if there are other handlers in the log registry, however, internally it just ignores bubbling.

adjenks:
Here's an example:
```
<?php

use Monolog\Logger;
use Monolog\Handler\StreamHandler;
use Monolog\Handler\WhatFailureGroupHandler;
require 'vendor/autoload.php';


// create a log channel
$log = new Logger('name');
$filehandler1 = new StreamHandler($_SERVER["DOCUMENT_ROOT"].'/', Logger::DEBUG,true); //fails without file name, should continue on after failure
[...]

Issue #500, opened 2015-01-28
Logging should be idempotent and not affect application state


paracycle:
I have just ran into a problem with Monolog and file permissions. While I realize that I do need to solve the file permission problem on its own; the behaviour of Monolog in this situation really surprised me.

I would expect a logging library to work in an idempotent manner and not affect the state of the application as a whole. However, Monolog (`StreamHandler` more specifically) intentionally chooses to throw an `UnexpectedValueException` if the output stream can't be opened. This, I find, is not acceptable in a logging framework. If the stream is not writable then all I'd be willing to lose is a few logs, not my whole application state.

Such behaviour _could_ make sense in a debug setting where one would like to be alerted of such problems and raising an exception could be made into a global or per handler flags that could be turned **on** (ie OFF by default).

If the author(s) of this library agree with me, I'd be willing to send a PR to this effect.

Seldaek:
I don't really agree we should silently suppress such errors. It's a clear misconfiguration and if we don't let you know the day you will need your logs they won't be there.

If you really want to you can easily create a `Monolog\Logger` wrapper that implements the `Psr\Log\LoggerInterface` and wraps the monolog call in a try/catch to suppress all errors or convert them to softer errors/php error_log.

paracycle:
It actually makes me sad to see this reply. I *do* understand that not being able to log is a problem and might lead to lost logs down the line. However, do you really believe that the right way to notify the application is to actually crash it by throwing an error. That might lead to lost application data which IMHO is much much more important than any logs.

I realize that I can wrap logging and others have done so, eg #426, but if throwing an exception was an option that people could turn on if they need/want it, wouldn't it make Monolog a better library? After all, throwing/catching exceptions is more wasteful than not throwing them in the first place.

I realize that at the end of the day, this is your codebase and the choice is yours but I really would like you to reconsider this point. Thank you.

fprochazka:
I would be really pissed if I found out after weeks my application hasn't been logging. So :+1: to @Seldaek

chriseskow:
:-1: for this request. In some situations, logging may be a critical requirement of the application (e.g. for auditing), and even if it isn't, failing silently by default will cause confusion by hiding real problems.

That said, perhaps the exceptions should be better documented (e.g. in docblocks) and/or converted to custom exception classes.

Seldaek:
[...]

PR #1012, opened 2017-06-27
Custom exception handler (#500)

vincepare:
Handlers, processors and formatters may rise exceptions while logging a record, and if the user does not wrap each Monolog calls with a try/catch, this will crash the entire app (#500). Using a WhatFailureGroupHandler seems to have [some downsides](https://github.com/Seldaek/monolog/issues/946).

So this pull request aims to keep Monolog's default behavior unchanged while allowing users to define their own default behavior through a callback, and for example turn Monolog exceptions to PHP Warnings, or change the HTTP response code.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -134,6 +134,11 @@ class Logger implements LoggerInterface
     protected $microsecondTimestamps = true;
 
     /**
+     * @var callable
+     */
+    protected $exceptionHandler;
+
+    /**
      * @param string             $name       The logging channel
      * @param HandlerInterface[] $handlers   Optional stack of handlers, the first one in the array is called first, etc.
      * @param callable[]         $processors Optional array of processors
@@ -329,16 +334,20 @@ class Logger implements LoggerInterface
             'extra' => array(),
         );
 
-        foreach ($this->processors as $processor) {
-            $record = call_user_func($processor, $record);
-        }
-
-        while ($handler = current($this->handlers)) {
-            if (true === $handler->handle($record)) {
-                break;
+        try {
+            foreach ($this->processors as $processor) {
+                $record = call_user_func($processor, $record);
             }
 
-            next($this->handlers);
+            while ($handler = current($this->handlers)) {
+                if (true === $handler->handle($record)) {
+                    break;
+                }
+
+                next($this->handlers);
+            }
+        } catch (\Exception $ex) {
+            $this->handleException($ex, $record);
         }
 
         return true;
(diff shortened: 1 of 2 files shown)


adjenks:
I did this on our site:


```
use Monolog\Handler\StreamHandler;

/**
 * Handles logging with grace. (Proceeds to next handler on Exceptions)
 *
 * Was created because the standard logger does not have the ability to fall 
 * back to another handler when one dies.
 */
class GracefulLogger extends \Monolog\Logger
{

	/**
     * Handler for throwables thrown by handlers
     *
     * To handle errors thrown by handlers, letting the next handler continue if one fails.
	 * Set to call error_log by default so you can debug any log handlers that fail
     *
     * @var callable
     */
	protected $handlerExceptionHandler;

    /**
     * @param string             $name       The logging channel
     * @param HandlerInterface[] $handlers   Optional stack of handlers, the first one in the array is called first, etc.
     * @param callable[]         $processors Optional array of processors
	 * @param \Closure           $handlerExceptionHandler The function to pass exceptions to when a handler fails. Defaults to error_log();
     */
    public function __construct($name, array $handlers = array(), array $processors = array(), \Closure $handlerExceptionHandler = null)
    {
        parent::__construct($name, $handlers, $processors);
		$this->handlerExceptionHandler = is_callable($handlerExceptionHandler) ? $handlerExceptionHandler : function($e){error_log($e);};
    }


    /**
     * Adds a log record.
     *
     * @param  int     $level   The logging level
     * @param  string  $message The log message
     * @param  array   $context The log context
     * @return Boolean Whether the record has been processed
     */
    public function addRecord($level, $message, array $context = array())
    {
        if (!$this->handlers) {
            $this->pushHandler(new StreamHandler('php://stderr', static::DEBUG));
        }

        $levelName = static::getLevelName($level);

        // check if any handler will handle this message so we can return early and save cycles
        $handlerKey = null;
        reset($this->handlers);
            if ($handler->isHandling(array('level' => $level))) {
                $handlerKey = key($this->handlers);
                break;
            }

            next($this->handlers);
        }

        if (null === $handlerKey) {
            return false;
        }

        if (!static::$timezone) {
            static::$timezone = new \DateTimeZone(date_default_timezone_get() ?: 'UTC');
        }

        // php7.1+ always has microseconds enabled, so we do not need this hack
        if ($this->microsecondTimestamps && PHP_VERSION_ID < 70100) {
            $ts = \DateTime::createFromFormat('U.u', sprintf('%.6F', microtime(true)), static::$timezone);
        } else {
            $ts = new \DateTime(null, static::$timezone);
        }
        $ts->setTimezone(static::$timezone);

        $record = array(
            'message' => (string) $message,
            'context' => $context,
            'level' => $level,
            'level_name' => $levelName,
            'channel' => $this->name,
            'datetime' => $ts,
            'extra' => array(),
        );

        }

		$stopBubble = false;
		
			try{
				$stopBubble = $handler->handle($record);
			}catch(\Throwable $e){ //php7
				($this->handlerExceptionHandler)($e); //brackets to explicitly indicate it is a function
			}catch(\Exception $e){ //php<7
				($this->handlerExceptionHandler)($e);
			}
			
            if ($stopBubble) {
                break;
            }

            next($this->handlers);
        }

        return true;
    }

}

```

adjenks:
Oh, my mistake, I didn't look at the code until now. Looks good, kinda like what I did except it looks more thorough.

adjenks:
I like that you made a setExceptionHandler() function instead of making the constructor take even more arguments like I did. You didn't want to call it setHandlerExceptionHandler()? Haha. Since it handles the exceptions of the handlers.

Seldaek:
Thanks, this is a good compromise I guess to allow people to handle errors in their own way.

PR #1017, opened 2017-07-10
Fixed message not being passed to client when capturing an exception

bendubuisson:
The raven client expects the message to be at the first level of the `$data` array when passing an exception.
see https://github.com/getsentry/sentry-php/blob/master/lib/Raven/Client.php#L795


Seldaek:
Merged as 22b92c7 in 1.x & master branches, thanks!

PR #1018, opened 2017-07-11
Add context to fluentd formatter

tnucera:
This pull request aims to add the record context, which is missing, to fluentd formatter.


Seldaek:
Thanks, merged in 1.x and master as 678693e

PR #1020, opened 2017-07-11
Update docs for GelfHandler

wshafer:
Updated docs to reflect the dependency of package graylog2/gelf-php in order to work


Seldaek:
Thanks

PR #1023, opened 2017-07-13
fix-for-slack-record-encoding

dis1:
When message appears in slack channel, there is a problem with cyrillic symbols. They are encoded in unicode and not readable. This fix allow to avoid such problem.


Seldaek:
Thanks

PR #1024, opened 2017-07-17
Check if monolog runs in web context

TiMESPLiNTER:
I faced the problem that if I use the FirePHP and ChromePHP handlers in a non-web context especially the firephp handler collected a lot of data (it stores every record in an internal array). So if you have a long process (for example a migration task) and the this process logs a lot of data you end up with a huge amount of memory in the end.

This PR checks if monolog/PHP runs in web context and if not doesn't do anything in case of the ChromePHP and FirePHP handlers.

**I will fix the tests once I got feedback for the idea and if it will be accepted.**

### DIFF
--- a/src/Monolog/Handler/AbstractBrowserHandler.php
+++ b/src/Monolog/Handler/AbstractBrowserHandler.php
@@ -1,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace Monolog\Handler;
+
+abstract class AbstractBrowserHandler extends AbstractProcessingHandler
+{
+
+    /**
+     * Checks if PHP's serving a web request
+     * @return bool
+     */
+    public function isWebRequest(): bool
+    {
+        return 'cli' !== php_sapi_name();
+    }
+}
--- a/src/Monolog/Handler/ChromePHPHandler.php
+++ b/src/Monolog/Handler/ChromePHPHandler.php
@@ -22,8 +22,11 @@ use Monolog\Logger;
  *
  * @author Christophe Coevoet <stof@notk.org>
  */
-class ChromePHPHandler extends AbstractBrowserHandler
+class ChromePHPHandler extends AbstractProcessingHandler
 {
+
+    use WebRequestRecognizerTrait;
+
     /**
      * Version of the extension
      */
--- a/src/Monolog/Handler/ChromePHPHandler.php
+++ b/src/Monolog/Handler/ChromePHPHandler.php
@@ -22,7 +22,7 @@ use Monolog\Logger;
  *
  * @author Christophe Coevoet <stof@notk.org>
  */
-class ChromePHPHandler extends AbstractProcessingHandler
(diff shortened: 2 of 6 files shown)


src/Monolog/Handler/AbstractBrowserHandler.php
```diff
@@ -0,0 +1,16 @@
+<?php declare(strict_types=1);
+
+namespace Monolog\Handler;
+
+abstract class AbstractBrowserHandler extends AbstractProcessingHandler
```
Seldaek: This could be a trait instead, that way it doesn't force the inheritance chain to include AbstractProcessingHandler in case another handler could make use of this method.

Seldaek:
Makes sense, thanks. If you can get the tests working again that'd be great, and also turn that abstract class into a trait.

src/Monolog/Handler/AbstractBrowserHandler.php
TiMESPLiNTER: Done 👍

TiMESPLiNTER:
@Seldaek what's the best way to test this now? As I make use of the builtin PHP function `php_sapi_name` I somehow need to emulate it returns not `cli`. There is a PHP function mock lib for PHPUnit but I don't want to introduce a new dependency just because of that. Should I create mocked Chrome/FirePHP handlers in the tests an mock the trait function of them?

Seldaek:
Yes I think mocking the function is the best way here.

TiMESPLiNTER:
@Seldaek done ✅

src/Monolog/Handler/ChromePHPHandler.php
```diff
@@ -75,6 +78,10 @@ public function __construct($level = Logger::DEBUG, $bubble = true)
      */
     public function handleBatch(array $records)
     {
+        if (false === $this->isWebRequest()) {
```
stof: We tend to use `if (!$this->isWebRequest())` in Monolog

src/Monolog/Handler/WebRequestRecognizerTrait.php
```diff
@@ -0,0 +1,16 @@
...
+trait WebRequestRecognizerTrait
+{
+
+    /**
+     * Checks if PHP's serving a web request
+     * @return bool
+     */
+    protected function isWebRequest(): bool
+    {
+        return 'cli' !== php_sapi_name();
```
stof: use `'cli' !== \PHP_SAPI && 'phpdbg' !== \PHP_SAPI` instead. The `\PHP_SAPI` will be resolved at compile time (when fully qualifying it), which will allow OPCache to resolve the full condition. `php_sapi_name` does not benefit from this AFAICT (especially when not fully qualified).

And the `phpdbg` SAPI is not a web one either.

TiMESPLiNTER:
@stof thanks for your feedback. Done ✅

Seldaek:
Thanks!

PR #1026, opened 2017-07-18
Add Hostname Processor

PurpleBooth:
I have found myself reimplementing this processor a number of times for
different clients. This processor is useful when you're running a number
of instances of an application across multiple hosts. This allows you to
narrow incidents to a specific host.

### DIFF
--- a/src/Monolog/Processor/HostnameProcessor.php
+++ b/src/Monolog/Processor/HostnameProcessor.php
@@ -1,0 +1,27 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Processor;
+
+/**
+ * Adds value of gethostname into the log extra
+ *
+ * @author Billie Thompson
+ */
+class HostnameProcessor
+{
+    public function __invoke(array $record): array
+    {
+        $record['extra']['hostname'] = gethostname();
+
+        return $record;
+    }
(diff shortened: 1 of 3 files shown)


HonzaMac:
We also have this hostname processor, it would be awesome if this could be part of Monolog for next release 👍

src/Monolog/Processor/HostnameProcessor.php
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Adds value of gethostname into the log extra
+ *
+ * @author Billie Thompson
+ */
+class HostnameProcessor
+{
+    public function __invoke(array $record): array
+    {
+        $record['extra']['hostname'] = gethostname();
```
staabm: Would it make sense to cache it and retrieve the hostname only once per request?

PurpleBooth: I've added in caching, which could improve performance, the disadvantage now is that in long running processes, if the hostname changes while the process is running, we won't reflect that in the logs.

PurpleBooth:
One thing that's not addressed in this is that gethostname can return false on failure. Is that an acceptable thing to have in the hostname field on failure?

src/Monolog/Processor/HostnameProcessor.php
HonzaMac: Is it possible to reset these values for long running processes somehow?

HonzaMac:
In our codebase HostnameHandler injects resolved name via the constructor, as this does not change over time and we know it when app starts. We have long running php script.

Seldaek:
Thanks, however closing in favor of #1040 which is kinda the same but with caching and false handling.

HonzaMac:
No worries, thanks!

PR #1027, opened 2017-07-19
StreamHandler: correct permissions for directory tree

roomcays:
When creating directory tree for StreamHandler make sure that directories will get `0777` permissions by using `umask()` function just before and just after `mkdir()` call. Without this some (most?) Apache servers will propagate user mask of `0022` to the PHP process resulting with directories having permissions of `0755` even if specified otherwise in second argument of `mkdir()` function.


Seldaek:
We do want the umask to be applied, the goal isn't to set everything to be 0777.

PR #1030, opened 2017-07-24
Check maximum allowed size of sqs message

HonzaMac:
SQS messages have upper limit of their body. 
It's better to log something instead of skip or fail this logg. 

I would prefer to log new error message, when this happen, via logger instance, but it's not accessible there.
i.e.: `$this->logger->error('Too long message for SQS handler [' . $headOfMessage . '...]')` 
Does anybody know, how to achive this and keep constructor minimal?


Seldaek:
Thanks

PR #1033, opened 2017-07-26
Queue interop handler.

makasim:
This handler allows to send logs to MQ using any [queue-interop](https://github.com/queue-interop/queue-interop) compatible transports (there 10th of them). 

Here's an example, the code could be found [here](https://github.com/php-enqueue/enqueue-sandbox/tree/master/monolog):

```php
<?php

use Monolog\Handler\QueueInteropHandler;
use Monolog\Logger;

require_once __DIR__.'/vendor/autoload.php';

$context = (new \Enqueue\Fs\FsConnectionFactory('file://'.__DIR__.'/queue'))->createContext();

// create a log channel
$log = new Logger('name');
$log->pushHandler(new QueueInteropHandler($context));

// add records to the log
$log->warning('Foo');
$log->error('Bar');
```

the consumer may look like this: 

```php
<?php

use Enqueue\Consumption\QueueConsumer;
use Interop\Queue\PsrMessage;
use Interop\Queue\PsrProcessor;

require_once __DIR__.'/vendor/autoload.php';

$context = (new \Enqueue\Fs\FsConnectionFactory('file://'.__DIR__.'/queue'))->createContext();

$consumer = new QueueConsumer($context);
$consumer->bind('log', function(PsrMessage $message) {
    echo $message->getBody().PHP_EOL;

    return PsrProcessor::ACK;
});

$consumer->consume();

```

TODO: 

* [ ] tests
* [ ] docs

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "php-console/php-console": "^3.1.3",
         "jakub-onderka/php-parallel-lint": "^0.9",
         "predis/predis": "^1.1",
-        "phpspec/prophecy": "^1.6.1"
+        "phpspec/prophecy": "^1.6.1",
+        "queue-interop/queue-interop": "^0.5"
     },
     "suggest": {
         "graylog2/gelf-php": "Allow sending log messages to a GrayLog2 server",
@@ -41,7 +42,8 @@
         "mongodb/mongodb": "Allow sending log messages to a MongoDB server (via library)",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
         "rollbar/rollbar": "Allow sending log messages to Rollbar",
-        "php-console/php-console": "Allow sending log messages to Google Chrome"
+        "php-console/php-console": "Allow sending log messages to Google Chrome",
+        "queue-interop/queue-interop": "Allow sending log messages to any MQ compatible with queue-interop"
     },
     "autoload": {
         "psr-4": {"Monolog\\": "src/Monolog"}
--- a/src/Monolog/Handler/QueueInteropHandler.php
+++ b/src/Monolog/Handler/QueueInteropHandler.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Interop\Queue\PsrContext;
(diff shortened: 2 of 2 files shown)


makasim:
@Seldaek are you positive on this? If so, I am going to add some tests and docs

Seldaek:
Thanks but for new handlers with third party dependencies I think it's better to publish them as separate packages with a hard require on the dependency, and then add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

makasim:
@Seldaek  It is not a dependency on an implementation, it is same as requiring `psr/log` package. It ships only MQ interfaces, you still have to add an implementation. 

IMO it would be good to have it here, and deprecate\remove other handlers such as AmqpHandler.

makasim:
A standalone package with one php file which provides only a contract. Kind of overkill

makasim:
Isn't the purpose of any contract in it? You can provide a feature without the need to couple on implementation.

Seldaek:
@makasim My point is, when you use the handler you need an additional dependency/requirement, so it makes more sense to have it externally so that if you want to use the handler you require it and it enforces the presence of its dependency instead of having to do runtime checks.

makasim:
The code was released as a standalone package (https://github.com/php-enqueue/monolog-queue-handler).

makasim:
The package has matured and gained some popularity. Supports [many transports](https://github.com/php-enqueue/enqueue-dev).

screenshot 2018-10-23 17 42 18

screenshot 2018-10-23 17 43 57

@Seldaek would you reconsider your choice? What could I do to make it happen?

Seldaek:
I don't really see a major benefit to having it in monolog itself.. And I'd rather not deprecate the AmqpHandler which right now works with the raw amqp handler class and not an abstraction that would then be forced on to people. I doubt this is going to change in the future.

Seldaek:
WTF man, no need to call me ignorant or to accuse me of double standards. I am not willing to maintain all handlers possible in the world, that's all. I have to make a choice at some point. I am sorry you disagree with that choice but that's all there is here.

makasim:
No personal accusation. That's how I see it from my perspective.

Chrico:
@makasim you're free to add your Handlers to the wiki of Monolog [1] - which already contains a ton of Handlers, i guess nobody wants to see a ton of handlers in Monolog.

I would even suggest, that Monolog does not provide any third party handlers anymore in future. Instead it should split them into own repositories or the offical company behind (e.G. Logmatic, HipChat, ...) should host them in their own account.

[1] https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

makasim:
I've removed my messages.

PR #1034, opened 2017-07-26
[WIP] Amqp interop handler.

makasim:
based on https://github.com/Seldaek/monolog/pull/1033, and should be merged after that one only. 

It is better than [AmqpHandler](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/AmqpHandler.php) in several ways

* monolog is not coupled to amqp implementation but only interfaces.
* monolog can work with any amqp-interop compatible trnasport. There are amqp-interop transports that work on top of phpamqplib as well as php-amqp extension.
* it can declare exchange

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "php-console/php-console": "^3.1.3",
         "jakub-onderka/php-parallel-lint": "^0.9",
         "predis/predis": "^1.1",
-        "phpspec/prophecy": "^1.6.1"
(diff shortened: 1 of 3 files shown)


Seldaek:
Same as the other PR, please publish as standalone package.

makasim:
A standalone package with one php file which provides only a contract. Kind of overkill

makasim:
The code was released as a standalone package (https://github.com/php-enqueue/monolog-queue-handler).

PR #1037, opened 2017-08-09
s/Boolean/boolean/

weirdan:
Having `@return` specify `Boolean` (with initial capital 'B') causes some static analyzers think it's a class (`Monolog\Handler\Boolean`). Also, it's not allowed according to phpdocumentor's BNF[1] (it only accepts all-lowercase version).

[1] https://www.phpdoc.org/docs/latest/references/phpdoc/types.html


Seldaek:
Fixed by #1077 but thanks

PR #1038, opened 2017-08-10
Remove misleading reference to RFC 5424 standard

EnoahNetzach:
Log levels (both the severity and the code) used here are not [RFC 5424](https://www.rfc-editor.org/rfc/rfc5424.txt) compliant.

I find it would be better off removing the misleading comment than changing log levels for end-users.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -87,8 +87,6 @@ class Logger implements LoggerInterface
     const API = 2;
 
     /**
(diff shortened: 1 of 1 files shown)


alexjfisher:
FWIW there's still https://github.com/Seldaek/monolog/blob/8471abd5ef19fb0d62828d8b807b2d08aea57e6e/src/Monolog/Level.php#L19

Seldaek:
I think line 20 explains well that the values are not the RFC ones, only the levels as identifiers

PR #1040, opened 2017-08-14
[feature] add host processor

mahone3297:


HonzaMac:
What is the difference from #1026 ?

mahone3297:
@HonzaMac I don't have a look at the PRs before implmenting it. I think it's the same. 😭

HonzaMac:
Your is caching `gethostname()` which is ok, but still does not handle all return types from that function

mahone3297:
@HonzaMac >but still does not handle all return types from that function
What's the meaning? I can't understand, can you explain?

HonzaMac:
It could [return FALSE ](http://php.net/manual/en/function.gethostname.php)
So you will have string or boolean(false) in extra arguments. 
i.e. elastic search will not like that and omit this message at all.

mahone3297:
@HonzaMac 
* function __invoke will return array
* $record['extra']['host'] will have default value ''
so, I don't think return bool(false) is a good option.

HonzaMac:
The default value will be immediately overwritten by value from the constructor and that could be string or false.

I.e. in `$record['extra']['host'] = 'localhost';` or `$record['extra']['host'] = false;`

staabm:
IMO those very simple processor classes should be defined in-line and dont need to be bundled with monolog. re-using them is not always possible (e.g. long running app, vs. web-sapi based app etc).

I would propose adding a example to the docs like

```php
$hostProcessor = new class {
    private static $host;
    
    public function __construct()
    {
        self::$host = gethostname();
    }
    
    public function __invoke(array $record): array
    {
        $record['extra']['host'] = self::$host;
 
        return $record;
    }
};
```

mahone3297:
@HonzaMac 
* host field, I think it's a string tyle, so I think `$record['extra']['host'] = ‘’;` is better than  `$record['extra']['host'] = false;`
* `$record['extra']['host'] = 'localhost';`
 maybe 'localhost' is not the correct value, so i think `''` is better than `'localhost'`, at least we can know get host failed

@staabm I don't think so. you can say it's very simple, but, I think many people will re-use it, so put it in monolog will be better, many people can be benefit from it.
ps: you say defined in-line, I am writing the code below in my project, it's more simple
```
$logger->pushProcessor(function (array $record) {
    $record['extra']['host'] = $this->host;
    return $record;
});
```

staabm:
> ps: you say defined in-line, I am writing the code below in my project, it's more simple

event better!

HonzaMac:
@mahone3297 
But your code does not do what you are saying ;)
add something like this: 
```
$hostname = gethostname();
if ($hostname === FALSE) {
  self::$host = '';
}
```

mahone3297:
@staabm
because 
```
$logger->pushProcessor(new HostProcessor());
```
is more simple than 
```
$logger->pushProcessor(function (array $record) {
    $record['extra']['host'] = $this->host;
    return $record;
});
```

src/Monolog/Processor/HostProcessor.php
```diff
@@ -0,0 +1,34 @@
...
+namespace Monolog\Processor;
+
+use Monolog\Logger;
+
+/**
+ * Injects value of gethostname in all records
+ */
+class HostProcessor
+{
+    private static $host = '';
```
HonzaMac: Default values is not necessary

src/Monolog/Processor/HostProcessor.php
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+namespace Monolog\Processor;
+
+use Monolog\Logger;
+
+/**
+ * Injects value of gethostname in all records
+ */
+class HostProcessor
```
HonzaMac: `HostnameProcessor` sounds more aqurate.

mahone3297: ok, I will change.

src/Monolog/Processor/HostProcessor.php
mahone3297: ok

Seldaek:
Thanks

PR #1041, opened 2017-08-16
Check for directory before throwing 'no existing directory' exception

joshmcrae:
In `StreamHandler::createDir()` it's possible for a directory to be created (by another request for example) between checking if the directory exists and actually attempting to create the directory. As a result it's possible that an `UnexpectedValueException` will be thrown even if the directory exists, since `mkdir` will error in this situation.

This PR simply adds an additional test for the absence of the directory before throwing the exception.


Seldaek:
Thanks, merged as c0eeca7 in 1.x & master

PR #1042, opened 2017-08-16
Sync RollbarHandler with the latest changes rollbar/rollbar package

ArturMoczulski:
We received a request to update Monolog's `RollbarHandler` to be in sync with the latest version of our SDK ( https://github.com/rollbar/rollbar-php ). The github issue can be found here: https://github.com/rollbar/rollbar-php/issues/249

This PR makes the `RollbarHandler` up to date with `rollbar-php` v1.0+.

Please, review and merge to master if it looks good to go.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "php-console/php-console": "^3.1.3",
         "jakub-onderka/php-parallel-lint": "^0.9",
         "predis/predis": "^1.1",
-        "phpspec/prophecy": "^1.6.1"
+        "phpspec/prophecy": "^1.6.1",
+        "rollbar/rollbar": "^1.3"
     },
     "suggest": {
         "graylog2/gelf-php": "Allow sending log messages to a GrayLog2 server",
--- a/composer.json
+++ b/composer.json
@@ -41,8 +41,7 @@
         "mongodb/mongodb": "Allow sending log messages to a MongoDB server (via library)",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
         "rollbar/rollbar": "Allow sending log messages to Rollbar",
-        "php-console/php-console": "Allow sending log messages to Google Chrome",
-        "rollbar/rollbar": "^1.3"
+        "php-console/php-console": "Allow sending log messages to Google Chrome"
     },
     "autoload": {
         "psr-4": {"Monolog\\": "src/Monolog"}
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,7 @@
     ],
     "require": {
         "php": "^7.0",
-        "psr/log": "^1.0.1",
-        "rollbar/rollbar": "^1.3"
+        "psr/log": "^1.0.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7",
@@ -42,7 +41,8 @@
         "mongodb/mongodb": "Allow sending log messages to a MongoDB server (via library)",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
         "rollbar/rollbar": "Allow sending log messages to Rollbar",
-        "php-console/php-console": "Allow sending log messages to Google Chrome"
+        "php-console/php-console": "Allow sending log messages to Google Chrome",
+        "rollbar/rollbar": "^1.3"
     },
     "autoload": {
         "psr-4": {"Monolog\\": "src/Monolog"}
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,7 @@
     ],
     "require": {
         "php": "^7.0",
-        "psr/log": "^1.0.1",
-        "rollbar/rollbar": "^1.3"
+        "psr/log": "^1.0.1"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7",
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,8 @@
     ],
     "require": {
         "php": "^7.0",
-        "psr/log": "^1.0.1"
+        "psr/log": "^1.0.1",
+        "rollbar/rollbar": "^1.3"
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7",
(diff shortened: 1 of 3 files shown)


arendjantetteroo:
@ArturMoczulski can you check the test failure in https://travis-ci.org/Seldaek/monolog/jobs/265012937 . I'm interested in this but doubt will be merged with a failing test.

ArturMoczulski:
@arendjantetteroo Just fixed the issue. This is ready to be reviewed and merged.

joostelders:
Can someone please merge this request?

ArturMoczulski:
Can any of the maintainers merge this?

ArturMoczulski:
We have two open community issues because of this PR still not being merged:
https://github.com/rollbar/rollbar-php/issues/275
https://github.com/rollbar/rollbar-php/issues/249

Can you look into this?

composer.json
```diff
@@ -14,7 +14,8 @@
     ],
     "require": {
         "php": "^7.0",
-        "psr/log": "^1.0.1"
+        "psr/log": "^1.0.1",
+        "rollbar/rollbar": "^1.3"
```
tPl0ch: This should probably go into `suggest` section of `composer.json` since not all Monolog users need the `rollbar/rollbar` dependency.

ArturMoczulski:
Rollbar has been moved to the suggested packages.

tPl0ch:
@ArturMoczulski you should additionally add it to `require-dev`, then you can run the tests unconditionally within the travis CI.

ArturMoczulski:
Fixed dependency conflict when building with `deps=low` (`composer update --prefer-dist --prefer-lowest --prefer-stable`).

ArturMoczulski:
@Seldaek can you review and merge?

composer.json
```diff
@@ -28,7 +28,8 @@
         "php-console/php-console": "^3.1.3",
         "jakub-onderka/php-parallel-lint": "^0.9",
         "predis/predis": "^1.1",
-        "phpspec/prophecy": "^1.6.1"
+        "phpspec/prophecy": "^1.6.1",
+        "rollbar/rollbar": "^1.3"
```
cordoval: i am not sure this should be only 2 digits, what if there is stuff that goes on 1.4 or beyond?

ArturMoczulski: I don't think I understand? We do want to allow non-breaking updates so I'm allowing everything up to `2.0.0`. Am I missing something?

cordoval:
can you rename toLog? does not sound like a good name to me

also do you have to use mb_strtolower on first argument of log? i just saw my custom logger like:

```
<?php

namespace Ghs\Application;

use Monolog\Handler\AbstractHandler;
use Monolog\Logger;
use Psr\Log\LoggerInterface;
use Rollbar\RollbarLogger;

class RollbarHandler extends AbstractHandler
{
    /** @var RollbarLogger */
    private $logger;

    public function __construct(LoggerInterface $logger, $level = Logger::DEBUG, $bubble = true)
    {

        $this->logger = $logger;
    }

    public function handle(array $record)
    {
        if (!$this->isHandling($record)) {
            return false;
        }

        $context = $record['context'];
        if (isset($context['exception']) && $context['exception'] instanceof \Exception) {
            $record['message'] = $context['exception'];
        }

        $this->logger->log(mb_strtolower($record['level_name']), $record['message'], $record['context']);

        return false === $this->bubble;
    }
}
```
and was just curious

composer.json
cordoval: maybe 1.5.x will have something that is somewhat breaking, or additional things not taken into account here. Hinting that this leverages those additional things is too presumptuous i guess.

ArturMoczulski: We follow the semantic versioning standard so we do not introduce any backward incompatible changes in `1.*`. If there is anything backward incompatible that we need to introduce, we bump the major to `2.*.*`. I believe `^1.3` constraint supports that the best. What do you think?

cordoval: @ArturMoczulski oh i see you are from rollbar itself, pardon my presumptuousness 👴

cordoval: 👍

ArturMoczulski: No worries. I appreciate your input ❤️️

ArturMoczulski:
@cordoval that's a great suggestion, but I'd rather stick to the convention we are already using in our other repos, like here: https://github.com/rollbar/rollbar-php/blob/master/src/RollbarLogger.php#L53-L74

Also, I don't see the need to use `mb_strtolower()`. Am I missing something?

ArturMoczulski:
Can you review and merge this anytime soon?

sburesi:
Any new on this ? Without this we are stuck with an outdated version of Rollbar
And Rollbar did the effort to propose a fix on monolog for this…

cordoval:
@sburesi @Seldaek is a bit busy probably but i am sure this is imminent so i expect it to be merged soon

theUniC:
Any updates on this?

cordoval:
i am sorry if i come to be annoying @Seldaek, can you tell us something please?

thanks

pzhivitsa:
Probably the support of this library is dead. Otherwise, I don't understand why there is no any feedback from @Seldaek

cordoval:
he is busy doing some terraform stuff https://github.com/terraform-aws-modules/terraform-aws-vpc/issues/77
The infrastructure for composer packagist and those sites is a heavy burden i bet. It locks him up for community work. But again let's hope he has a bit of time for this.

Let's keep up the patience guys.

jasonbouffard:
As much as I'd like to see Monolog support this first class, there is no reason why Rollbar cannot include this Monolog Handler in https://github.com/rollbar/rollbar-php.

ArturMoczulski:
@jasonbouffard yes, we have been waiting for this a long time now. I think we will just include `MonologHandler` as part of the `rollbar/rollbar-php` repo until this gets merged.

ArturMoczulski:
We included this handler in our repo: https://github.com/rollbar/rollbar-php/pull/325 until this PR gets merged in.

Find more information on how to use it here: https://github.com/rollbar/rollbar-php#using-monolog

Seldaek:
Thanks - for users of monolog 1.x I suggest using https://github.com/rollbar/rollbar-php-examples/tree/master/monolog - for monolog 2 users (which I wouldn't recommend using until we tag at least an alpha release), feel free to use the new version of the RollbarHandler.

PR #1044, opened 2017-08-17
Don't stop processing other handlers if one fails

kristijorgji:
If one handler throws exception while processing the record the other ones won't process the record because of the exception.

I suggest you surround that part with a try catch, so all handlers will try to logg properly the record.
That way will be more fault tolerant the logging.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -315,10 +315,12 @@ class Logger implements LoggerInterface
         }
 
         while ($handler = current($this->handlers)) {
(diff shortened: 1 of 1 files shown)


kristijorgji:
I checked WhatGroupHandler and decided to close this request and open later request in which each handler can decide itself it's own fault tolerance.

Something like 

 try {
                if (true === $handler->handle($record)) {
                    break;
                }
            } catch (\Exception $e) {
                if ($handler->isFaultTolerant()) {
                    throw $e;
                }
            }

PR #1045, opened 2017-08-17
If context has exactly 1000 items, do not truncate

itscaro:
When there is 1000 items, the normalizer truncates the 1000th item and add the message of normlization abort. It would be better not to truncate when there is exactly 1000 items. When there are more than 1000 items, log 1000 items + the message of abort.


Seldaek:
Merged as ada5d30 in 1.x & master branches, thanks!

PR #1047, opened 2017-08-24
Get the last logged message

banstola:
For testing it is useful to get what was the last thing that the logger logged.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -134,6 +134,12 @@ class Logger implements LoggerInterface
      * @var DateTimeZone
      */
     protected $timezone;
+    
+    /**
+     * @var array
+     */
+    
+    protected $logMessageStack = [];
 
     /**
      * @param string             $name       The logging channel, a simple descriptive name that is attached to all log records
@@ -315,6 +321,7 @@ class Logger implements LoggerInterface
         }
 
         while ($handler = current($this->handlers)) {
+            $this->logMessageStack [] = $message;
             if (true === $handler->handle($record)) {
                 break;
             }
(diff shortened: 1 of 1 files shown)


src/Monolog/Logger.php
```diff
@@ -315,6 +321,7 @@ public function addRecord(int $level, string $message, array $context = []): boo
         }
 
         while ($handler = current($this->handlers)) {
+            $this->logMessageStack [] = $message;
```
martijncalker: This builds up an array with all processed log-messages. Within small scripts or web-requests, that wouldn't be a big problem, but with long running applications it'll be a big problem.

Try it yourself and add the method
```php
    public function counting()
    {
        return count($this->logMessageStack);
    }
```

and run the script:
```php
<?php

require __DIR__.'/vendor/autoload.php';

$logger = new \Monolog\Logger(
    'test',
    [new \Monolog\Handler\StreamHandler('php://stdout')],
    [new \Monolog\Processor\MemoryUsageProcessor(), new \Monolog\Processor\MemoryPeakUsageProcessor()]
);


for ($i = 0; $i < 1000000; ++$i)
{
    $logger->info(
        'Logging your message',
        [
            'i' => $i,
            'count' => $logger->counting(),
        ]
    );
}
```


```sh
❯ php test.php 
[2017-08-24 15:33:26] test.INFO: Logging your message {"i":0,"count":0} {"memory_usage":"2 MB","memory_peak_usage":"2 MB"}
...
[2017-08-24 15:37:52] test.INFO: Logging your message {"i":113462,"count":113462} {"memory_usage":"40 MB","memory_peak_usage":"40 MB"}
```

Is it not possible for you to mock a handler?

banstola: Thanks for your suggestion. I will write a test. Maybe the better idea would be to just get the last log without making it an array.

HonzaMac:
Sound interesting, but it's just for testing. Could you write it as a separate handler?

PR #1049, opened 2017-08-24
[ PHPDOC ] Add phpdoc for function return

banstola:


Seldaek:
Thanks but it's not needed to add type hints when the function declaration already has a return type IMO.

banstola:
True that.

Issue #1039, opened 2017-08-11
Unable to extend SlackWebhookHandler 


JayChandler:
I need to extend the slack record before it gets posted via curl. My idea was to subclass SlackWebhookHandler and overwrite the `write` method but unfortunately the properties from the base class are **private**

The only thing i have is 
```
public function getSlackRecord()
    {
        return $this->slackRecord;
    }
```

but then there is this!

```
 /**
     * Slack Webhook token
     * @var string
     */
    private $webhookUrl;
```

To get this working i would have to copy the whole class into a new file and do my moifications. Thats not very nice.

It would be very nice if the class would have a callback that is called before the JSON serialization happens to have some kind of extensibility.

```
protected function write(array $record)
    {
        $postData = $this->slackRecord->getSlackData($record);
        // apply callback
        $postData = $this->extendRecord(  $postData );
        $postString = json_encode($postData);

        $ch = curl_init();
        $options = array(
            CURLOPT_URL => $this->webhookUrl,
            CURLOPT_POST => true,
            CURLOPT_RETURNTRANSFER => true,
            CURLOPT_HTTPHEADER => array('Content-type: application/json'),
            CURLOPT_POSTFIELDS => $postString
        );
        if (defined('CURLOPT_SAFE_UPLOAD')) {
            $options[CURLOPT_SAFE_UPLOAD] = true;
        }

        curl_setopt_array($ch, $options);

        Curl\Util::execute($ch);
    }
```

Seldaek:
Fixed by #1088

PR #1051, opened 2017-08-28
Change private to protected

rap2hpoutre:
See: https://github.com/Seldaek/monolog/issues/1039

Also, in similar handler classes (see: [mandrill](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/MandrillHandler.php) or [newrelic](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/NewRelicHandler.php)), all properties are `protected` (and not `private`)


Seldaek:
Fixed by #1088

PR #1052, opened 2017-09-10
Include InsightOps handler

gmsantos:
[Rapid7 bought Logentries](https://www.rapid7.com/about/press-releases/rapid7-acquires-logentries/) and decided to [rebrand it to InsightOps](https://www.rapid7.com/info/logentries-insightops/).

InsightOps works the same way that LogEntries, except that a new endpoint (either on Europe or US) is used.

Since Monolog supports LogEntries, I think its good to support InsightOps as well.

### DIFF
--- a/doc/02-handlers-formatters-processors.md
+++ b/doc/02-handlers-formatters-processors.md
@@ -55,6 +55,7 @@
 - _RollbarHandler_: Logs records to a [Rollbar](https://rollbar.com/) account.
 - _SyslogUdpHandler_: Logs records to a remote [Syslogd](http://www.rsyslog.com/) server.
 - _LogEntriesHandler_: Logs records to a [LogEntries](http://logentries.com/) account.
+- _InsightOpsHandler_: Logs records to a [InsightOps](https://www.rapid7.com/products/insightops/) account.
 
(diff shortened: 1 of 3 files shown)


Seldaek:
Thanks, I guess this means the LogEntries handler should be deprecated/removed then?

gmsantos:
Hi @Seldaek. Logentries signup page redirects to InsightOps signup, but they didn't fully migrate all the older accounts (still using logentries at work and didn't recevied any migration/service shutdown notice).

I'm okay to deprecate LogentriesHandler for 1.x and remove it on 2.x. 

In the worst scenario I could maintain LogentriesHandler for 2.x like I did provisionally for [Insightops handler](https://github.com/gmsantos/monolog-insightops-handler)

Seldaek:
Ok that's fine we can just keep the old one I guess. It's fairly low maintenance. Thanks for the details.

stof:
should we actually bring the insightops handler into the core, or keep it as a third-party repo @Seldaek ?

Seldaek:
I think I'll merge it in as we had Logentries in and this provides continuity plus it's very little code.

PR #1055, opened 2017-09-19
Fix Comparison operator in RotatingFileHandler

chonaso:
This fixes the handling of comparing $this->nextRotation and $record['datetime'] whether Log-file must be rotated.
At present, Log-file can't be rotated in the case of just 'YYYY-MM-DD 00:00:00.000000'.


Seldaek:
Thanks

PR #1057, opened 2017-10-03
Change LogstashFormatter prefix properties to keys

CarsonF:
It doesn't really make sense to let half of the key be configurable and the other half hardcoded.
Previously, the prefixes made sense because it wasn't a nested array so each key was dynamic.


Seldaek:
Makes sense I think but that'll have to be for 2.0 as it's a breaking change.

PR #1058, opened 2017-10-03
Fixing file rotation when similar files are present

albankora:
When file rotate is active and there are files with similar names like dev-2017-08-18.log and dev-doctrine-2017-08-18.log and dev-requests-2017-08-18.log the current Glob pattern will match all of them and not the one we want. The problem was that if we had set it up for max files to be 2 and file dev-2017-08-18.log was not present It will have not been created. Changing the pattern from * to [0-9][0-9][0-9][0-9]* will make the matching more targeted to the file we want.

### DIFF
--- a/src/Monolog/Handler/RotatingFileHandler.php
+++ b/src/Monolog/Handler/RotatingFileHandler.php
@@ -166,7 +166,7 @@ class RotatingFileHandler extends StreamHandler
         $fileInfo = pathinfo($this->filename);
         $glob = str_replace(
             ['{filename}', '{date}'],
-            [$fileInfo['filename'], '*'],
+            [$fileInfo['filename'], '[0-9][0-9][0-9][0-9]*'],
             $fileInfo['dirname'] . '/' . $this->filenameFormat
         );
         if (!empty($fileInfo['extension'])) {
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/RotatingFileHandler.php
```diff
@@ -166,7 +166,7 @@ protected function getGlobPattern()
         $fileInfo = pathinfo($this->filename);
         $glob = str_replace(
             ['{filename}', '{date}'],
-            [$fileInfo['filename'], '*'],
+            [$fileInfo['filename'], '[0-9][0-9][0-9][0-9]*'],
```
jtyr: This glob patter should as explicit as possible to cover the used file format. The current pattern covers only the year but not the month and day. This should be changed to something like this:

```php
            [$fileInfo['filename'], '20[0-9][0-9]-[01][0-9]-[0-3][0-9]*'],
```

jtyr:
@Seldaek Please could you review this PR?

albankora:
@Seldaek please could you review?

Seldaek:
Merged as 5f87836 in 1.x & master branches, thanks!

PR #1062, opened 2017-10-15
Rename Boolean in phpDocs to bool

ErikBooij:
### Why?
It might seem a little pedantic, but because the declaration of boolean values in doctypes wasn't in accordance with standards (according to [phpdoc.org](https://docs.phpdoc.org/guides/types.html) and the proposed [PSR-5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#appendix-a-types)), it sometimes gives issues in my IDE and it always gives problems when running for example [vimeo/psalm](https://github.com/vimeo/psalm) over a project using Monolog, because `Boolean` in the doctype is (rightfully) interpreted as class `Monolog[\Namespace]\Boolean`.

### DIFF
--- a/src/Monolog/Handler/AbstractHandler.php
+++ b/src/Monolog/Handler/AbstractHandler.php
@@ -24,8 +24,8 @@ abstract class AbstractHandler extends Handler
     protected $bubble = true;
 
     /**
-     * @param int     $level  The minimum logging level at which this handler will be triggered
-     * @param Boolean $bubble Whether the messages that are handled can bubble up the stack or not
+     * @param int  $level  The minimum logging level at which this handler will be triggered
+     * @param bool $bubble Whether the messages that are handled can bubble up the stack or not
      */
     public function __construct($level = Logger::DEBUG, $bubble = true)
     {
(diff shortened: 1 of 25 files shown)


ErikJson:
Hah, I was just about to create a PR for this myself, addressing this issue that was created soon after this PR: #1063 

I'll leave this one for you instead then :) However note that you need to adjust the spacing for adjacent lines in some places where "Boolean" was previously the longest word (see the very first file in the diff), and you should also update `@return` annotations.

ErikBooij:
@ErikJson Good point, will fix as soon as I find the time, probably tonight. Thanks.

ErikBooij:
@Seldaek If you have time, could you take a look? Apparently there a few more people waiting for this and it's only docblocks. Would be appreciated.

Seldaek:
I'm not against this but it should be done against the 1.x branch so it can be released soon. Master branch is for 2.0 and that'll still take a while I'm afraid.

ErikBooij:
Good to know, I wasn't aware. Created a new one: https://github.com/Seldaek/monolog/pull/1077

Want to keep this PR for 2.0, or will it be rebased against 1.x?

PR #1064, opened 2017-10-19
Add native RELP handler to reliably log to Rsyslog.

derekmarcotte:


stof:
I suggest submitting this to the 1.x branch if you want to have it in a release anytime soon.

src/Monolog/Handler/RelpHandler.php
```diff
@@ -0,0 +1,318 @@
...
+        $res = $this->_relpReadResponse();
+        if ($res == "0") {
+            $this->_isOpen = false;
+        }
+
+        $this->_disconnectSocket();
+        return $res;
+    }
+
+    public function syslog($message) {
```
stof: why would it be public ?

PR #1065, opened 2017-10-20
Add native RELP handler to reliably log to Rsyslog.

derekmarcotte:
Resubmitting #1064 to 1.x branch.

### DIFF
--- a/src/Monolog/Handler/RelpHandler.php
+++ b/src/Monolog/Handler/RelpHandler.php
@@ -1,0 +1,318 @@
+<?php
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Logger;
+use Monolog\Handler\AbstractSyslogHandler;
+use Monolog\Formatter\LineFormatter;
+
+/**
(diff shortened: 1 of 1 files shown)


Seldaek:
Do you have any details as to how this is more reliable/different than the [SyslogUdpHandler](https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/SyslogUdpHandler.php) for example? I'm rather against adding new handlers unless they are really needed, and this one is quite a bit of code, so I'd like to understand the use case/tech better.

derekmarcotte:
Sure!  Glad you asked.  

RELP is a reliable syslog protocol.  Messages sent over UDP are not guaranteed to be delivered.  Same with TCP.   RELP adds a transactional component whereby the server acknowledges receipt.  Rsyslog is the primary server that supports this protocol, but there are others.

Our use case is receiving `JSON Lines` over HTTP and injecting them into our reliable logging stream.  We don't want to lose these.

We have multiple RELP servers, and can retry, and load balance between them, and so have a very high guarantee that we have not lost messages.

More details about RELP here:
https://en.wikipedia.org/wiki/Reliable_Event_Logging_Protocol
https://www.rsyslog.com/doc/relp.html

Seldaek:
Thanks for the details. Due to the fact it's a fairly large amount of code though I'd rather have you submit this as a standalone package on packagist.org so I don't have to maintain it in here. I'm trying to avoid adding new handlers unless they are really general purpose. You can also add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages to help discoverability.

PR #1067, opened 2017-10-23
Fix #1066: Use correct docblock castings.

### LINKED ISSUE #1066: Wrong DocBlock annotation in ErrorHandler::register causes static analyses to fail
The signature of the method `ErrorHandler::register()` is this (`v1.23.0`):

```php
    /**
     * Registers a new ErrorHandler for a given Logger
     *
     * By default it will handle errors, exceptions and fatal errors
     *
     * @param  LoggerInterface $logger
     * @param  array|false     $errorLevelMap  an array of E_* constant to LogLevel::* constant mapping, or false to disable error handling
     * @param  int|false       $exceptionLevel a LogLevel::* constant, or false to disable exception handling
     * @param  int|false       $fatalLevel     a LogLevel::* constant, or false to disable fatal error handling
     * @return ErrorHandler
     */
    public static function register(LoggerInterface $logger, $errorLevelMap = array(), $exceptionLevel = null, $fatalLevel = null)
```

It accepts a `LogLevel::*` constant or `null` or `false`.

This causes problems as:

1. `LogLevel::*` constants contain `string`s and not `int`egers;
2. It is possible to pass `null` to use the default `LogLevel::*` constant (`LogLevel::ALERT` for `$fatalLevel` and `LogLevel::ERROR` for `$exceptionLevel` that internally is `$this->uncaughtExceptionLevel`).
3. `int` are never used and so it is impossible to fix by ourselves the issue.

**To fix this: Fix the docblock to cast to `null|string|false`.**

INFO: To statically analyze my code I'm using `Phan` (https://github.com/phan/phan).

Aerendir:
This fixes the castings of param `$fataLevel` to be `string|null|false` as it should be.

### DIFF
--- a/src/Monolog/ErrorHandler.php
+++ b/src/Monolog/ErrorHandler.php
@@ -49,9 +49,9 @@ class ErrorHandler
      *
      * By default it will handle errors, exceptions and fatal errors
      *
-     * @param  LoggerInterface    $logger
-     * @param  array|false        $errorLevelMap     an array of E_* constant to LogLevel::* constant mapping, or false to disable error handling
-     * @param  array|false        $exceptionLevelMap an array of class name to LogLevel::* constant mapping, or false to disable exception handling
+     * @param  LoggerInterface   $logger
+     * @param  array|false       $errorLevelMap     an array of E_* constant to LogLevel::* constant mapping, or false to disable error handling
+     * @param  array|false       $exceptionLevelMap an array of class name to LogLevel::* constant mapping, or false to disable exception handling
      * @param  string|null|false $fatalLevel        a LogLevel::* constant, null to use the default LogLevel::ALERT or false to disable fatal error handling
      * @return ErrorHandler
      */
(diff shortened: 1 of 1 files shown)


src/Monolog/ErrorHandler.php
```diff
@@ -49,10 +49,10 @@ public function __construct(LoggerInterface $logger)
...
      * By default it will handle errors, exceptions and fatal errors
      *
-     * @param  LoggerInterface $logger
-     * @param  array|false     $errorLevelMap     an array of E_* constant to LogLevel::* constant mapping, or false to disable error handling
-     * @param  array|false     $exceptionLevelMap an array of class name to LogLevel::* constant mapping, or false to disable exception handling
-     * @param  int|false       $fatalLevel        a LogLevel::* constant, or false to disable fatal error handling
+     * @param  LoggerInterface    $logger
+     * @param  array|false        $errorLevelMap     an array of E_* constant to LogLevel::* constant mapping, or false to disable error handling
+     * @param  array|false        $exceptionLevelMap an array of class name to LogLevel::* constant mapping, or false to disable exception handling
+     * @param  string|null||false $fatalLevel        a LogLevel::* constant, null to use the default LogLevel::ALERT or false to disable fatal error handling
```
tractorcow: Does the double `||` have significance?

Aerendir: It was a typo. Fixed now.

tractorcow:
I trust it's ok for me to give reviews of PRs on this repo. :)

Seldaek:
Thanks

PR #1071, opened 2017-10-31
Fix typo

kalifg:
## Description
All other documentation and code suggests that the replacement token is of the form `{var}` and not `%var%`


Seldaek:
thanks

PR #1075, opened 2017-11-02
FIX: Fixes #1074 By passing 4th param to captureMessage().

### LINKED ISSUE #1074: Unable to send stacktrace using RavenHandler
Using `Monolog\Handler\RavenHandler::write()` there is no option to send a stacktrace ala `Raven_Client::captureMessage()` (4th arg is not collected or passed).

Solution: Add an extra dimension to the `write()` method's `$record` param "trace" or "stack" and then pass `$record['some_param']` as the 4th arg to `$this->ravenClient->captureMessage($record['formatted'], array(), $options, $record['some_param']);`

phptek:
A 4th param is now passed to to `RavenHandler::captureMessage()`

### DIFF
--- a/src/Monolog/Handler/RavenHandler.php
+++ b/src/Monolog/Handler/RavenHandler.php
@@ -184,7 +184,13 @@ class RavenHandler extends AbstractProcessingHandler
             $options['extra']['message'] = $record['formatted'];
             $this->ravenClient->captureException($record['context']['exception'], $options);
         } else {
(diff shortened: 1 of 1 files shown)


phptek:
Bump

Seldaek:
The right way to capture the stack is to pass an exception instance in the log context's `exception` key (as defined in PSR-3), then Raven's captureException is used, which captures the stack.

PR #1076, opened 2017-11-02
Added SlackRecord parameter for markdown supppport in attachments

ryan-christie:
Was attempting to bold messages when using the Slack handler and attachments and it wasn't processing the message as markdown. Currently markdown support is set statically to only 'fields.' Added argument that excepts an array to allow user to allow markdown in 'text' and 'pretext' as allowed in the Slack documentation.

### DIFF
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -76,6 +76,12 @@ class SlackRecord
     private $excludeFields;
 
     /**
+     * List of fields to use markdown in when using an attachment
+     * @var [type]
+     */
+    private $markdownIn;
+
+    /**
      * @var FormatterInterface
      */
     private $formatter;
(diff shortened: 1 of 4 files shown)


src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -85,7 +91,7 @@ class SlackRecord
      */
     private $normalizerFormatter;
 
-    public function __construct($channel = null, $username = null, $useAttachment = true, $userIcon = null, $useShortAttachment = false, $includeContextAndExtra = false, array $excludeFields = array(), FormatterInterface $formatter = null)
+    public function __construct($channel = null, $username = null, $useAttachment = true, $userIcon = null, $useShortAttachment = false, $includeContextAndExtra = false, array $excludeFields = array(), array $markdownIn = array(), FormatterInterface $formatter = null)
```
Seldaek: If you want this in a 1.x release the argument need to be added last otherwise it's a BC break.

src/Monolog/Handler/SlackWebhookHandler.php
```diff
@@ -46,8 +46,9 @@ class SlackWebhookHandler extends AbstractProcessingHandler
      * @param int         $level                  The minimum logging level at which this handler will be triggered
      * @param bool        $bubble                 Whether the messages that are handled can bubble up the stack or not
      * @param array       $excludeFields          Dot separated list of fields to exclude from slack message. E.g. ['context.field1', 'extra.field2']
+     * @param array       $markdownIn             List of fields to use markdown in when using an attachment
```
Seldaek: Would be good to have docs here as to which fields are accepted, or maybe even validation in SlackRecord of which fields can be configured if this is a fixed list on slack's end.

Seldaek:
Closing this to clean up, feel free to re-open against latest master if there is still interest.

PR #1077, opened 2017-11-04
Rename Boolean in phpDocs to bool on 1.x

ErikBooij:
### Why?
It might seem a little pedantic, but because the declaration of boolean values in doctypes wasn't in accordance with standards (according to [phpdoc.org](https://docs.phpdoc.org/guides/types.html) and the proposed [PSR-5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#appendix-a-types)), it sometimes gives issues in my IDE and it always gives problems when running for example [vimeo/psalm](https://github.com/vimeo/psalm) over a project using Monolog, because `Boolean` in the doctype is (rightfully) interpreted as class `Monolog[\Namespace]\Boolean`.

### DIFF
--- a/src/Monolog/Handler/AbstractHandler.php
+++ b/src/Monolog/Handler/AbstractHandler.php
@@ -32,8 +32,8 @@ abstract class AbstractHandler implements HandlerInterface
     protected $processors = array();
 
     /**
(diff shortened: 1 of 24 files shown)


Seldaek:
Thanks

PR #1079, opened 2017-11-07
microtime uses UTC always

Steache:
- based on https://github.com/Seldaek/monolog/blob/master/src/Monolog/DateTimeImmutable.php#L34
- cannot use output of microtime, because it uses UTC and ignores actual timezone

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -316,8 +316,8 @@ class Logger implements LoggerInterface
 			$timestamp = microtime(true);
 
 			// apply offset of the timezone as microtime() is always UTC
-			if (self::$timezone && self::$timezone->getName() !== 'UTC') {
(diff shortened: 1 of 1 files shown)


src/Monolog/Logger.php
```diff
@@ -311,9 +311,16 @@ public function addRecord($level, $message, array $context = array())
             static::$timezone = new \DateTimeZone(date_default_timezone_get() ?: 'UTC');
         }
 
+		$timestamp = microtime(true);
```
stof: this code should be inside the `if ($this->microsecondTimestamps && PHP_VERSION_ID < 70100) {`, to avoid doing useless work on newer PHP versions

Steache: - @stof Can I kindly ask you for another check? :) 
- What should be the next step. Should it be merged to one commit or something like that? :)

PR #1083, opened 2017-11-17
Add an OverflowHandler

krisbuist:
This handler can be used to not let any messages from a certain lever through, unless they cross a configured threshold of number of messages in a level.
This is useful together with, for example, a MailHandler when some batch processing is done and you're only interested in
major failure and not a minor one.

Example usage:

```php
$handler = new NativeMailHandler(...);
$handler = new OverflowHandler($handler, Logger::WARNING => 10]);

$logger = new Logger('my-logger', [$handler]);

// Batch processing starts
foreach ($iterable as $row) {
  try {
    ...
  } catch (\Exception $ex) {
    $logger->warning($ex->getMessage());
  }
}
```

This code sample will only pass the log  messages to the mail handler when there are more than 10 warning messages are logged

### DIFF
--- a/doc/02-handlers-formatters-processors.md
+++ b/doc/02-handlers-formatters-processors.md
@@ -125,10 +125,10 @@
   has accessors to read out the information.
 - [_HandlerWrapper_](../src/Monolog/Handler/HandlerWrapper.php): A simple handler wrapper you can inherit from to create
  your own wrappers easily.
-- [_ThresholdHandler_](../src/Monolog/Handler/ThresholdHandler.php): This handler will buffer all the log messages it
-  receives, up until a configured threshold is reached, after it will pass all log messages to the wrapped handler.
-  Useful for applying in bath processing when you're only interested in significant failures instead of minor, single
-  erroneous events. 
+- [_OverflowHandler_](../src/Monolog/Handler/OverflowHandler.php): This handler will buffer all the log messages it
+  receives, up until a configured threshold of number of messages of a certain lever is reached, after it will pass all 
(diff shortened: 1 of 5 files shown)


Seldaek:
Thanks for the PR. However this sounds to me like it does exactly the same as the FingersCrossedHandler does? Am I mistaken or did you perhaps miss it due to the not so expected naming? :)

krisbuist:
The FingersCrossedHandler and the ThresholdHandler have a different purpose, but I agree they are somewhat related. Where the FingersCrossedHandler buffers lower levels until a certain _level_ is reached, the ThresholdHandler buffers a specific level until it reached a certain _number_ of logs of that level.

I apply this handler in certain batch programming tasks where I process – for example – 100K rows. But when just a percentage of the rows triggers log errors, I only want them to end up in the general log, but when the amount of log messages reaches a configured threshold I want my logs to end up in a more prominent channel.

Maybe the naming of the handler could be improved to avoid confusing with the FingersCrossedHandler, but I thought the current naming would cover the load.

Seldaek:
Ok I see, thanks for the explanation. I think the problem in the description at least is that "Threshold" sounds like it could apply to an error-level threshold while it is in fact a threshold on the amount of log records present. If anyone has a better idea for a name I'm all ears, otherwise it'll have to do I guess but we have to make sure the description/docs are very clear.

Seldaek:
Still looking for a better name here.. How about OverflowHandler? FloodgateHandler?

krisbuist:
Neat! I thought about it a lot and couldn't come up with something decent, but I think `OverflowHandler` covers it quite well. Thanks! I'll change the name soon

Seldaek:
Thanks

PR #1084, opened 2017-11-21
Clean up the Logger docs

xuanquynh:
I try to clean the docs whenever I read the source code.


xuanquynh:
It isn't necessary really.

PR #1088, opened 2017-11-30
Add getter for webhook url in SlackWebhookHandler class

MilesChou:
Add getter so that the subclass can easy to inherit SlackWebhookHandler class


Seldaek:
Thanks

JayChandler:
Thanks

PR #1090, opened 2017-12-05
Fix message placeholder example

sdelicata:
Use braces instead of percent char


Seldaek:
Fixed in e3ab4e680bf28c40798ba79bad057fa01e45be9d already but thanks anyway :)

Issue #1091, opened 2017-12-05
Two Objects in log line confuses cloud watch


acinader:
The default line logger emits two objects: channel and extra.  Amazon Cloud Watch doesn't allow one to target these objects in metrics if more than one object exists in the logline.

JSON targetting works great if there is only one object, but when a log line has two, the JSON targetting doesn't work at all.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/FilterAndPatternSyntax.html

My proposed solution is to make an option on the LineFormatter to combine extra and context objects.

Seldaek:
Closing as per https://github.com/Seldaek/monolog/pull/1092#issuecomment-395973783

PR #1092, opened 2017-12-05
Merge context & extra

acinader:
fixes: #1091

1.  Note that there are two commits, the first one adds some coverage and fixes some whitespace nits with `ignoreEmptyContextAndExtra`

2. The second commit introduces the `combineContextAndExtra` option and tests to cover the new option.

I chose to have context win if there is a collision between extra and context.  It makes most sense to me to allow an individual log event to override an extra.


Seldaek:
I think it'd be best to do this with a processor on the handler that merges extra and context, then you can make the LineFormatter only output %context% for example if that's all you are interested in. I'd rather not add more options for this class.

acinader:
@Seldaek makes total sense.  thanks for the tip to use a processor for this too!

PR #1098, opened 2017-12-21
Update README.md

ilyaplot:


ilyaplot:
Yii 2.1 already uses Psr logger.

PR #1099, opened 2017-12-22
Send the stacktrace in the message

thiagotalma:
Added the option to send the stacktrace in the message that is not an exception.


Seldaek:
Thanks but I am not sure this is the best idea to have a stacktrace context changing the mode of operation of the handler. The best is to log exception as  `['exception' => $e]` in the context info to get stack traces.

PR #1102, opened 2018-01-05
Configurable JsonFormatter maximum normalized depth and number of items

### LINKED ISSUE #1094: Can you set this value to be set
https://github.com/Seldaek/monolog/blob/7b992836275e09ed63c63fe33ca9993e515e6c5d/src/Monolog/Formatter/JsonFormatter.php#L134

1000 is too larger，can you set this value to be set

3Q

tomzx:
Allow JsonFormatter maximum normalized depth and maximum number of items per level to be configured in the constructor.

Fixes #1094.

### DIFF
--- a/src/Monolog/Formatter/JsonFormatter.php
+++ b/src/Monolog/Formatter/JsonFormatter.php
@@ -27,6 +27,8 @@ class JsonFormatter extends NormalizerFormatter
 
     protected $batchMode;
     protected $appendNewline;
+    protected $maxNormalizeDepth = 9;
+    protected $maxNormalizeItemCount = 1000;
 
     /**
      * @var bool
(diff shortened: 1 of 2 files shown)


Seldaek:
Could you add setters instead of new constructor arguments? There are too many of these already it gets super messy and this isn't something that's frequently enough tweaked that it needs to be in constructor IMO.

tomzx:
@Seldaek Done!

src/Monolog/Formatter/JsonFormatter.php
```diff
@@ -59,6 +61,38 @@ public function isAppendingNewlines(): bool
...
+     */
+    public function getMaxNormalizeItemCount(): int
+    {
+        return $this->maxNormalizeItemCount;
+    }
+
+    /**
+     * @param $maxNormalizeItemCount
+     */
+    public function setMaxNormalizeItemCount($maxNormalizeItemCount)
```
Seldaek: Why no int type hint here? Also please don't include @param annotations for parameters with type hints, they're redundant.

tomzx: Alright. Since the code isn't too consistent, I didn't know what to go with. I also see that this is 7.1 and not 7.0, so I'll add the void return type as well.

Seldaek:
Thanks

PR #1104, opened 2018-01-11
Optional possibility to filter log records on logger level

PavelLoparev:
### Motivation
It might be quite useful to have a possibility to filter log records on logger level, not only on handler level. For example, I want to log events from different parts of my application into the same location and I want to control log level of records for each defined logger as well. For now, it looks like:
```
$logger1 = new Logger('logger1');
$logger1->pushHandler(new StreamHandler('test.log', Logger::INFO)); // Logs records with level >= info.

$logger2 = new Logger('logger2');
$logger2->pushHandler(new StreamHandler('test.log', Logger::ALERT)); // Logs records with level >= alert.

...
```
Where `test.log` is the same target file *for all of the handlers*. Thus I have to create as many *same handlers* as loggers. It becomes a mess when you use pure `monolog` library and define all the loggers & handlers in `*.yml` file. I propose next approach (which doesn't break any previous approaches with handler level filtering):
```
$handler = new StreamHandler('test.log');

$logger1 = new Logger('logger1', [$handler], [...], 'info'); // Logs records with level >= info.
$logger2 = new Logger('logger2', [$handler], [...], 'alert'); // Logs records with level >= alert.
...
```
It's not necessary to pass log level directly to the constructor. It can be set up with `Logger::setLevel()` method.

### Benefit
We don't need to create many same handlers for different loggers, we can create one, share it with loggers and control log level on logger side.

### Backward compatibility
This patch doesn't break anything. You still can filter log records at handler level.

### Modified files
* `src/Monolog/Logger.ph`
* `tests/Monolog/LoggerTest.php`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ composer.lock
 .DS_Store
 .php_cs.cache
 .hg
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ composer.lock
 .DS_Store
 .php_cs.cache
 .hg
-.idea
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -721,11 +721,12 @@ class Logger implements LoggerInterface
      *
      * @param string $level
      */
-    public function setLevel($level) {
+    public function setLevel($level)
+    {
         $level = strtoupper($level);
 
-        if (!in_array($level, self::$levels, TRUE)) {
-            throw new InvalidArgumentException('Level "'.$level.'" is not defined, use one of: '.implode(', ', array_keys(array_flip(self::$levels))));
+        if (!in_array($level, self::$levels, true)) {
+            throw new InvalidArgumentException('Level "'.$level.'" is not defined, use one of: '.implode(', ', array_values(self::$levels)));
         }
 
         $this->level = array_search($level, self::$levels);
@@ -736,7 +737,8 @@ class Logger implements LoggerInterface
      *
      * @return int
      */
-    public function getLevel() {
+    public function getLevel()
+    {
         return $this->level;
     }
 }
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -725,7 +725,7 @@ class Logger implements LoggerInterface
     {
         $level = strtoupper($level);
 
-        if (!in_array($level, self::$levels, true)) {
+        if (!in_array($level, self::$levels)) {
             throw new InvalidArgumentException('Level "'.$level.'" is not defined, use one of: '.implode(', ', array_values(self::$levels)));
         }
 
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -15,6 +15,7 @@ use Monolog\Handler\HandlerInterface;
 use Monolog\Handler\StreamHandler;
 use Psr\Log\LoggerInterface;
 use Psr\Log\InvalidArgumentException;
+use Psr\Log\LogLevel;
 
 /**
  * Monolog log channel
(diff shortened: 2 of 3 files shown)


PavelLoparev:
Hi @Seldaek and @stof 
Could you please take a look at this PR?

.gitignore
```diff
@@ -5,3 +5,4 @@ composer.lock
 .DS_Store
 .php_cs.cache
 .hg
+.idea
```
BR0kEN-: I wouldn't add this change since it's unrelated.

src/Monolog/Logger.php
```diff
@@ -137,12 +160,15 @@ class Logger implements LoggerInterface
      * @param string             $name       The logging channel
      * @param HandlerInterface[] $handlers   Optional stack of handlers, the first one in the array is called first, etc.
      * @param callable[]         $processors Optional array of processors
+     * @param string             $level Minimal level to log
```
BR0kEN-: I think the first letter of comment to the parameter should be adjusted to the same level as others.

src/Monolog/Logger.php
```diff
@@ -137,12 +160,15 @@ class Logger implements LoggerInterface
...
      */
-    public function __construct($name, array $handlers = array(), array $processors = array())
+    public function __construct($name, array $handlers = array(), array $processors = array(), $level = 'DEBUG')
```
BR0kEN-: I would prefer having the `self::DEBUG` as the default value for the `$level` parameter with a further modification of the `setLevel` method.

src/Monolog/Logger.php
```diff
@@ -697,4 +728,28 @@ public static function setTimezone(\DateTimeZone $tz)
...
+
+    /**
+     * Set the minimal log level.
+     *
+     * Logger will log only records with log level >= $level.
+     *
+     * @param string $level
+     */
+    public function setLevel($level) {
+        if (!isset(static::$levelMapping[strtoupper($level)])) {
```
BR0kEN-: In my opinion the `$levelMapping` property is redundant and we can go with `in_array(strtoupper($level), static::$levels, TRUE)`.

BR0kEN-: The `strtoupper($level)` meets twice per method and it seems reasonable to store the result in a variable.

PavelLoparev: Fixed.

PavelLoparev: Fixed.

src/Monolog/Logger.php
PavelLoparev: It's not comfortable to use integer value in `*.yml` file when you define a logger. You have to always remember mapping `int -> string`. I believe it's better to use level names instead of the level key.

But I agree it's not ok to have hardcoded string in constructor parameter so I've just changed it to `LogLevel::DEBUG` constant from `Psr/Log` package which is a dependency for `monolog` library.

src/Monolog/Logger.php
PavelLoparev: Fixed

.gitignore
PavelLoparev: Fixed

PavelLoparev:
Thanks for the review @BR0kEN-

Seldaek:
Thanks for the PR. The problem is filtering logs based on a simple level is just one way to do it. We have a lot of different ways to filter https://github.com/Seldaek/monolog/blob/master/doc/02-handlers-formatters-processors.md#wrappers--special-handlers and so I don't think this is good enough, but I also don't wanna have a ton more complexity in the logger. Creating instances of filtering handlers is really cheap, so I don't see this brings much added value to be honest.

PR #1108, opened 2018-01-29
Fix mkdir concurrency creating

FallDi:
Issue:
Two PHP processes try to create directory concurrently.
Both process successfully pass  `if (null !== $dir && !is_dir($dir)) {` condition, but only one process successfully creates directory and second process will throw Exception, but directory is created.

`clearstatcache()` is not required due to PHP documentation http://php.net/manual/en/function.clearstatcache.php

> You should also note that PHP doesn't cache information about non-existent files. So, if you call file_exists() on a file that doesn't exist, it will return FALSE until you create the file.

Same for `is_dir()`

Please also check https://github.com/kalessil/phpinspectionsea/blob/master/docs/probable-bugs.md#mkdir-race-condition

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -169,7 +169,7 @@ class StreamHandler extends AbstractProcessingHandler
             set_error_handler([$this, 'customErrorHandler']);
             $status = mkdir($dir, 0777, true);
             restore_error_handler();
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, but the same fix was applied already in c0eeca76fdb1b5410e7af36bbfe5af265eeb5bf6 so closing this.

PR #1113, opened 2018-02-13
Updating Sentry Handler and Adding Breadcrumb Handler

Coffee2CodeNL:


Seldaek:
This seems to add BC breaks for no benefit so closing.

PR #1115, opened 2018-02-14
Update RotatingFileHandler.php

elwin1234:
Use own constant instead of hardcoded string for default dateFormat (const FILE_PER_DAY = 'Y-m-d';)


Seldaek:
Thanks

PR #1117, opened 2018-02-22
Register signal handlers

RGustBardon:


Seldaek:
Would love a review from someone else interested in the topic here. It's quite a lot of code.

helhum:
Code looks fine so far. Tests also pass on my system. 👍 

Before going into a deeper review a general question:

@RGustBardon Are there any benefits to add the code to `ErrorHandler` instead of adding a new class `SignalHandler`?

When reviewing this pr, I also stumbled upon https://github.com/Seldaek/signal-handler, which already provides to possibility to attach a psr logger to a signal.

Which brings me to an even more general question (targeted at @Seldaek), whether it would make sense to extract such functionality into a separate package (maybe with 2.0), or keep it here for pragmatic reasons.

RGustBardon:
@helhum:
> Are there any benefits to add the code to ErrorHandler instead of adding a new class SignalHandler?

No, I cannot see any. Moreover, it is confusing since POSIX signals are not (always related to) errors. The code has been refactored. Thank you for this helpful suggestion!

Seldaek:
Thanks!

PR #1121, opened 2018-03-02
ErrorLogHandler: use strict comparison

glensc:
otherwise `null` and `false` get accepted as valid types:

```php
$handler = new Handler\ErrorLogHandler(null, $level);
```

note: this is somewhat backward incompatible, as previously `null` and `false` and `""` would be converted to `0`, but now they are rejected.

also, maybe it would make sense to allow `null` to be "use default value `self::OPERATING_SYSTEM`"?


Seldaek:
Seems fine to me as is, no reason to give anything else than a valid value I guess. Thanks

PR #1122, opened 2018-03-02
Prevent to call newrelic functions when app and transaction names are…

dedalozzo:
… empty

The handler shouldn't set the app and transaction names when they are not been passed in the constructor or provided through the context. If the value returned by `getAppName()` or `getTransactionName()` is null or an empty string, the correspondent New Relic API shouldn't be called because the user might have called them in another part of the code. And anyway it doesn't make any sense call `newrelic_set_appname()` and `newrelic_name_transaction()` passing an empty string or null.

### DIFF
--- a/src/Monolog/Handler/NewRelicHandler.php
+++ b/src/Monolog/Handler/NewRelicHandler.php
@@ -167,7 +167,8 @@ class NewRelicHandler extends AbstractProcessingHandler
      */
     protected function setNewRelicAppName($appName)
     {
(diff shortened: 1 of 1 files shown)


Seldaek:
I don't quite see what you mean. https://github.com/Seldaek/monolog/blob/ce0a2b954b8cce09b68eb352833633c807b9099f/src/Monolog/Handler/NewRelicHandler.php#L79-L86 does not call the setters unless there is a value. Closing.

PR #1123, opened 2018-03-02
Newrelic handler with line formatter

thiagorb:
The New Relic Handler will crash when using the LineFormatter.

The current implementation tries to use `$record['formatted']` as an array, but it should be a string.

### DIFF
--- a/src/Monolog/Handler/NewRelicHandler.php
+++ b/src/Monolog/Handler/NewRelicHandler.php
@@ -80,21 +80,21 @@ class NewRelicHandler extends AbstractProcessingHandler
             $this->setNewRelicAppName($appName);
         }
 
+        $formattedRecord = is_array($record['formatted']) ? $record['formatted'] : $record;
         if ($transactionName = $this->getTransactionName($record['context'])) {
             $this->setNewRelicTransactionName($transactionName);
-            unset($record['context']['transaction_name']);
+            unset($formattedRecord['context']['transaction_name']);
         }
 
-        $message = isset($record['formatted']) ? $record['formatted'] : $record['message'];
         if (isset($record['context']['exception']) && $record['context']['exception'] instanceof \Exception) {
-            newrelic_notice_error($message, $record['context']['exception']);
-            unset($record['context']['exception']);
+            newrelic_notice_error($this->getMessage($formattedRecord), $record['context']['exception']);
+            unset($formattedRecord['context']['exception']);
         } else {
-            newrelic_notice_error($message);
+            newrelic_notice_error($this->getMessage($formattedRecord));
         }
 
-        if (isset($record['context']) && is_array($record['context'])) {
-            foreach ($record['context'] as $key => $parameter) {
+        if (isset($formattedRecord['context']) && is_array($formattedRecord['context'])) {
+            foreach ($formattedRecord['context'] as $key => $parameter) {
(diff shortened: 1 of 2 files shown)


moderndeveloperllc:
I don't think you want to change the handler to fit the output of one particular formatter. The default formatter is `NormalizerFormatter`. Not using the `formatted` key is basically negating having a formatter at all.

thiagorb:
Hey Mark, thanks for checking my PR. With your comment I saw that the 'normalisations' would be ignored with my changes.

I adjusted the implementation to work with the normalised data (in case it is present).

ralphschindler:
This is a legitimate issue, the NewRelicHandler needs to respect that a Formatter might return a string. 

@Seldaek / @stof care to weigh in or merge? :)

Seldaek:
I don't think the handler needs to be able to handle any formatter no.. Formatters and handlers have to work well together, that's just a fact. Maybe there is a valid case for using LineFormatter here but that's the first thing we should discuss IMO. What's the use case here?

ralphschindler:
I think the core issue here is that if you attempt to use a generalized `LineFormatter` in your application and then inject it into all of your loggers (you may have multiples), it is not until you try to log something to the `NewRelicHandler` that you see a PHP warning b/c it is attempting to run `unset()` on a string (thus you get another exception thrown in your exception handler).

> Formatters and handlers have to work well together, that's just a fact.

I would say from a purely code perspective, someone might have a reasonable expectation that anything implementing `LineFormatter` should generally "work" with a Hander it was injected in? If that is not a reasonable expectation, perhaps `NewRelicHandler` should override `setFormatter` and do a check to ensure `$formatter instanceof NormalizeFormatter`.

Personally, I think either decision is fine, I just think that if you're incorrectly setting something up, there should be an exception at setup time (when calling `setFormatter()`) rather than runtime (calling `write()`).

(for reference `NewRelicHandler` uses the `FormattableHandlerTrait` which generally allows any `FormatterInterface` to be injected: https://github.com/thiagorb/monolog/blob/829cd052bb4b0315c88c3606cd65a288985edc22/src/Monolog/Handler/FormattableHandlerTrait.php#L29-L37)

src/Monolog/Handler/NewRelicHandler.php
```diff
@@ -104,8 +105,8 @@ protected function write(array $record)
             }
         }
 
-        if (isset($record['formatted']['extra']) && is_array($record['formatted']['extra'])) {
-            foreach ($record['formatted']['extra'] as $key => $parameter) {
+        if (isset($formattedRecord['extra']) && is_array($formattedRecord['extra'])) {
+            foreach ($formattedRecord['extra'] as $key => $parameter) {
```
Seldaek: This won't actually work if you format to a string then formattedRecord['extra'] will not be normalized, it'll be the original data you passed to extra (or context above). That means if you pass in some data type that is not handled by the new relic code it'll error out. That's why the NormalizerFormatter is the default and the right choice IMO as it makes sure everything is normalized to common data-types.

Seldaek:
@ralphschindler I see the point, but the issue is it's not that easy. LineFormatter extends NormalizerFormatter so we can't simply type hint the latter. I don't wanna do an exact check for that class either because someone might want to use their own version of a normalizer formatter that works slightly differently. Documenting this better might be the best option here really.

Issue #909, opened 2017-01-05
New SlackWebookHandler and long context/extras problem


mirfilip:
Hi,
I just got around to test the [new cool SlackWebhookHandler](https://github.com/Seldaek/monolog/pull/894). It works nice and fixes problems I had before, there is however one annoyance I found. @an1zhegorodov I didn't want to continue discussion in your original PR though.

I use attachments and include context and extra data. Depending on the length of the context, it's either formatted well or cut:

### Looking good:
```php
$nestedArray = [];
for ($i = 0; $i < 27; $i++) {
    $nestedArray["foo $i"] = str_repeat($i, $i * 2);
}
$context = ['nested' => $nestedArray];
$this->logger->warning('Warning test', $context);
```

the whole nested array is displayed in code block (using "```").

### If I, however, make it bigger:
```php
$nestedArray = [];
for ($i = 0; $i < 30; $i++) {
    $nestedArray["foo $i"] = str_repeat($i, $i * 2);
}
$context = ['nested' => $nestedArray];
$this->logger->warning('Warning test', $context);
```
The context output is truncated and the whole code block is broken. Below, it's how it looks like at the beginning of the msg (I don't include full msg cause it's big):
![slack_monolog_handler_big_context](https://cloud.githubusercontent.com/assets/1812341/21681208/8d493788-d34e-11e6-8d2a-192a89bc27dd.png)

You can see at the beginning that there is no closing "```" of code block (probably because truncation). I don't have time to debug this now, it could be as well a limitation of Slack Webhook formatting.

What I'm trying to do:
I want to include a stacktrace (not based on exception) in a context so that it's prettified and displayed in code block on Slack. The `NormalizerFormatter` and "stringification" get the job done, but only for small contexts.

an1zhegorodov:
After debugging this, I can tell there is limitation somewhere around 2000 symbols for each attachment imposed by slack itself. Any ideas how to handle this?

acasademont:
AFAIK there are no limits when using the normal `message` field instead of attachments, which is the weirdest thing ever...

Seldaek:
Maybe you could truncate the output by hand to make sure the backticks is present at the end, or at least report it to slack so that they fix their truncation to not rip apart a code block..

mirfilip:
@Seldaek I hope by you, you mean me - the user of the library. Then - yes, I could, I already do that since I found out about the limitation.  

I didn't have time to look into it but I'll get around to reporting it someday ;)

[...]

PR #1127, opened 2018-03-18
Add a record with an object as context

gmponos:
I have added a test case where the record contains in it's context an object and in this case an Exception.

As we can see below the test fails because value for slack must always be a string but objects are not formatted to strings.

This results in slack displaying the field with no value inside of it.

If you like I can proceed with a fix on this PR but the only solution that I have is adding a Custom Formatter for slack. Would you accept this solution?

### DIFF
--- a/src/Monolog/Handler/Slack/SlackRecord.php
+++ b/src/Monolog/Handler/Slack/SlackRecord.php
@@ -137,7 +137,6 @@ class SlackRecord
                 $attachment['fields'][] = $this->generateAttachmentField('Level', $record['level_name']);
             }
 
-
             if ($this->includeContextAndExtra) {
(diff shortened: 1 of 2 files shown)


gmponos:
I've worked on a soluton for normalizing also objects on this [commit](056161130aed0b44291c0f56c37a17a7dce2bfe8).

Although this solution makes the exception still include the stack trace and does not solve the issue about the characters that it is mentioned #909

So I still believe that a slack normalizer would be nice.

Seldaek:
Thanks, I took the test in 68e39bd84f4ba24d9bc36bd44dfaa4dfd75c5c45 and applied a slightly shorter fix.

PR #1128, opened 2018-03-26
Improve table row output in HtmlFormatter

King2500:
* Makes table cells vertical-aligned top (otherwise you have to scroll a lot if you have large data in your extra fields for example)
* `width` HTML attribute should be used without "px" suffix
* Removes non existing `spacing` CSS attribute (and is obsolete, because table already has `spacing="1"`)


xelan:
Nice, thanks :+1: 
Another thing we noticed: Maybe the text color should be set wherever a background color is set. Otherwise, text might show as white on white if a dark theme is used in the mail client.

Seldaek:
Merged as 948bb4a in 1.x & master branches, thanks! 

@xelan I also set the text color to hopefully fix that dark theme support issue you mentioned. See 2893c2b

PR #1129, opened 2018-03-28
Allow setting stream chunk size in SocketHandler

klemenb:
This pull requests adds an option to the SocketHandler to set the desired stream chunk size via `stream_set_chunk_size`.

The main problem this pull request solves is logging larger messages via UDP using SocketHandler. The default chunk size set by PHP appears to be 8192 and any larger UDP messages get split into multiple packets and result in multiple log entries. When used with logstash (and similar processors) those messages are therefore not parsed correctly.

Since the UDP protocol allows for much larger packets (https://en.wikipedia.org/wiki/User_Datagram_Protocol#Packet_structure), this problem can be solved by simply using a larger chunk size.

### DIFF
--- a/src/Monolog/Handler/SocketHandler.php
+++ b/src/Monolog/Handler/SocketHandler.php
@@ -27,7 +27,7 @@ class SocketHandler extends AbstractProcessingHandler
     private $timeout = 0;
     private $writingTimeout = 10;
     private $lastSentBytes = null;
-    private $chunkSize = 8192;
+    private $chunkSize = null;
     private $persistent = false;
     private $errno;
     private $errstr;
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, I wonder why not just set it to a larger size though if it's overall a good thing? Is there a case where we don't want it to be long? 65,507 seems to be the max, I don't know if we should default to that or something more conservative but it sounds to me like having a smarter default beats one more option that people have to even know about to make use of.

klemenb:
Setting chunk size affects TCP streams as well and it could cause some unexpected issues in certain environments.

Also, one reservation that is more or less moot even at 8192 is that large UDP packets are even more unreliable - https://stackoverflow.com/a/35697810.

Honestly, I currently cannot think of a specific scenario where a larger default chunk size would cause issues, but my gut feeling says it is a bad idea to silently change this.

Seldaek:
Alright makes sense. One more thing though in that case, I'd rather have this default to null and only do the stream_set_chunk_size call in case a chunk size was provided, so if users don't provide anything the default is used whatever it is. In case the default value changes in PHP it's probably best to avoid overriding it with our own default.

Seldaek:
Thanks, I did the changes myself.

klemenb:
Great, thanks. Sorry I wasn't available in the past few days.

Seldaek:
No worries, just trying to move things forward.

PR #1132, opened 2018-03-30
Make Logger::setHandlers() consistent with Logger::__construct

### LINKED ISSUE #1131: Logger::__construct not consistent with Logger::setHandlers
When passing a handler array, there is a difference when this is done between passing the array to the constructor of `Logger`, compared to passing it to `setHandlers`. The constructor does support having a `null` value in the array, but `setHandlers` does not, due to the way that the parameters of `pushHandler` are typed.

CC007:
At the moment, the constructor supports having null in the handler array. This pull request makes it so that `setHandlers` also supports this.

Closes #1131

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -168,7 +168,7 @@ class Logger implements LoggerInterface
     /**
      * Pushes a handler on to the stack.
      */
-    public function pushHandler(HandlerInterface $handler): self
+    public function pushHandler(?HandlerInterface $handler): self
     {
         array_unshift($this->handlers, $handler);
 
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -167,8 +167,10 @@ class Logger implements LoggerInterface
 
     /**
      * Pushes a handler on to the stack.
+     *
+     * @param HandlerInterface|null $handler
      */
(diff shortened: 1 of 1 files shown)


gmponos:
A developer might want to initiate a Logger without a handler. 
Do various stuff with this and then set the handler. I can imagine use cases where you want to do that.
But why a developer would like to call `setHandlers` with null as argument? Can you explain your use case because I can't imagine any?

CC007:
To be specific, im not talking about setting handlers to null, but about setting handlers to an array that contains null. 

I found this out when our production environment had one of its handlers become null, as it was only used in the dev environment. Then while testing, I swapped out the handlers for a Testhandler temporarily and after the test tried to swap them back with setHandlers. 

So in effect, setHandlers didnt accept the array that I got through getHandlers.

CC007:
Also, can someone explain to me why the nightly build fails?

gmponos:
> So in effect, setHandlers didnt accept the array that I got through getHandlers.

Just to be clear the function name is `setHandler` hence it is supposed to accept a single handler and not an array of handlers.

Seldaek:
Closing as per https://github.com/Seldaek/monolog/issues/1131#issuecomment-395974662

PR #1133, opened 2018-04-01
Compatible to phpunit6

DQNEO:
* PHPUnit 5 is already out of support https://phpunit.de/
* When we require php7, we can use PHPUnit6.


Seldaek:
Thanks

PR #1134, opened 2018-04-04
Remove conditional php version check for debug_backtrace legacy support

lukewaite:
Given we will now only be supporting PHP 7.0 and greater, we can do away with this conditional which offered backwards compatible support to PHP versions less than 5.3.6.


Seldaek:
Thanks

Issue #901, opened 2016-12-11
Elastica vs Elasticsearch api?


rohdef:
First of thanks for this logger, I am very grateful for having such a useful logger available to me :)

I'm wondering why Monolog is not using the [Elasticsearch php API](https://www.elastic.co/guide/en/elasticsearch/client/php-api/current/index.html) for the Elastic logger? In stead of a 3rd party API that may or may not be up to date and may or may not stay so.

Are there deeper reasons for the choice that I'm not aware of? If not may I suggest using the official API (e.g., as that was the logical search for me was actually that) to integrate with Elasticsearch?

No matter what I would suggest updating the documentation for the handler, for instance referring to the API to use :)

Seldaek:
AFAIK Elastica has been the option of choice for a long time, and was started in 2010 while the official client only dates 2013.. That said at this point both might be equally good options. It probably would be possible to update the ElasticSearchHandler to accept either an Elastica client or one instance of the official one, if you are interested in using that one feel free to send a PR for that.

akalongman:
@Seldaek PR is ready :smile:

PR #1136, opened 2018-04-07
[2.0] Add support for official Elasticsearch client

akalongman:
This is a BC breaking change, so can be merged in future 2.0 version.

Changes:

- Renamed legacy `ElasticSearchHandler` to `ElasticaHandler` to avoid name collision
- Added new `ElasticSearchHandler` and `ElasticSearchFormatter`
- Added tests
- Updated upgrade guide

Related issue #901


akalongman:
Build is failing for nightly PHP: "php: symbol lookup error: /home/travis/.phpenv/versions/master/lib/php/extensions/no-debug-zts-20170718/mongodb.so: undefined symbol: ZEND_HASH_GET_APPLY_COUNT"

Looks like its Travis issue

Seldaek:
Please remove the .travis.yml changes, they don't belong in this PR.

src/Monolog/Handler/ElasticSearchHandler.php
```diff
@@ -46,20 +54,20 @@ class ElasticSearchHandler extends AbstractProcessingHandler
...
     {
         parent::__construct($level, $bubble);
         $this->client = $client;
         $this->options = array_merge(
             [
-                'index'          => 'monolog',      // Elastic index name
-                'type'           => 'record',       // Elastic document type
-                'ignore_error'   => false,          // Suppress Elastica exceptions
+                'index'        => 'monolog',      // Elastic index name
+                'type'         => 'record',       // Elastic document type
```
polyfractal: I'm not sure if there is bwc or legacy reasons for wanting the type name to be `record`, but if not I'd recommend using `_doc`.

Elasticsearch is in the process of removing types entirely.  Using `_doc` will allow the URLs to remain the same once types are removed, since `_doc` is being used as part of the URL in future versions (e.g. `POST /my_index/_doc/123` is the new URL in 7.0, with `_doc` being mandatory because it is part of the URL and not a type anymore).

So for future convenience, I'd go with `_doc` if possible.

More details here: https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html

src/Monolog/Handler/ElasticSearchHandler.php
```diff
@@ -112,17 +120,38 @@ public function handleBatch(array $records)
...
     /**
-     * Use Elasticsearch bulk API to send list of documents
-     * @param  array             $documents
+     * Use ElasticSearch bulk API to send list of documents
+     *
+     * @param  array $records
      * @throws \RuntimeException
      */
-    protected function bulkSend(array $documents)
+    protected function bulkSend(array $records)
```
polyfractal: I'm not sure if it matters, or if this is against Monolog philosophy or something, but I wanted to mention a potential concern.  It may be prudent to put some kind of limit on the bulk size.

It's possible to make an Elasticsearch cluster very unhappy if you send extremely large bulk requests.  So perhaps a soft limit for bulk size, or automatically send multiple bulk requests if the size is greater than a limit?  Unfortunately the issue is physical size, not number of documents, so it may not be practical to limit.  We recommend bulks in the range of ~50mb to give some context.

Also, it may be against the philosophy of Monolog (e.g. the library just logs whatever you send, and the user is responsible for making sure it's not abusive).  Which would make sense too.

polyfractal:
Left some comments about the Elasticsearch portions, but generally the ES bits LGTM!

Minor nit, I didn't bother commenting on them in the code since it doesn't really matter, but it's `Elasticsearch` not `ElasticSearch` or `Elastic Search`. :)

src/Monolog/Handler/ElasticSearchHandler.php
akalongman: I am not sure what @Seldaek thinks, but imho currently it is not necessary, because
> the library just logs whatever you send, and the user is responsible for making sure it's not abusive

akalongman:
@polyfractal thank you very much for the valuable review! I will update code considering your notes

tomsommer:
Any chance of getting this merged?

akalongman:
The question is for @Seldaek

Btw, today/tomorrow I will resolve current conflicts

makasim:
@Seldaek The PR introduces a dependency on 3rd party library. In another https://github.com/Seldaek/monolog/pull/1033 you said it is "no go". 

Could you please elaborate on this?

tomsommer:
Please merge this

tomsommer:
@akalongman Perhaps add `timeout` on `index()` to the possible options? and set a default "low" value?

PR #1140, opened 2018-04-11
Fix: Remove VersionEye badge

localheinz:
This PR

* [x] removes the VersionEye badge

💁‍♂️ For reference, see https://blog.versioneye.com/2017/10/19/versioneye-sunset-process/.


Seldaek:
Thanks

PR #1141, opened 2018-04-11
Fix: Boolean vs bool

localheinz:
This PR

* [x] uses `bool` instead of `Boolean`


stof:
I suggest doing this in the 1.x branch

localheinz:
@stof 

Give me a moment!

Seldaek:
Fixed by #1077

PR #1144, opened 2018-04-24
Adjust visibility of some properties to protected for extend slack handlers

MilesChou:
We must use another property to store the `private` properties at parent class.

> See an example: https://github.com/104corp/monolog-extensions/blob/master/src/Handlers/ProxyableSlackWebhookHandler.php#L27

It's not a good idea. Please use `protected` visibility so that we can extend slack handlers easily.

### DIFF
--- a/src/Monolog/Handler/SlackHandler.php
+++ b/src/Monolog/Handler/SlackHandler.php
@@ -27,13 +27,13 @@ class SlackHandler extends SocketHandler
      * Slack API token
      * @var string
      */
-    protected $token;
+    private $token;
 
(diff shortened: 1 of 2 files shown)


MilesChou:
I see. But we need extend the slack handler for proxy option / use Guzzle. Should I create another class (like the example) ? or change Monolog's source and submit a new PR?

MilesChou:
I forgot another choice: getter/setter.

I submit new commit for add the getter.

Seldaek:
Thanks

Issue #1145, opened 2018-05-02
New Relic handler doesn't correctly handle PHP7 Error instances


pauluswebster:
As per the New Relic (NR) documentation for [newrelic_notice_error()](https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newrelic_notice_error), current versions of their PHP Agent support passing instances of PHP7's Throwable to newrelic_notice_error() as well as the long standing support for instances of Exception.

However when monolog's NR handler is handling an error where 'exception' exists in context, it is still explicitly checking this is of class Exception, which will not match any PHP7 Error instances.

@see [Monolog/Handler/NewRelicHandler.php](https://github.com/Seldaek/monolog/blob/1.x/src/Monolog/Handler/NewRelicHandler.php#L87)

This then results in the fallback usage of newrelic_notice_error() where only the error message is passed to newrelic_notice_error(), which subsequently results in the NR agent compiling the error trace implicitly, including the call stack of the logging itself, rather than taking it explicitly from the Throwable's trace back to the origin of the error.

As per as similar change on the Raven Handler, this should be amended to support instances of Throwable in PHP >= 7. This change related to issue #759, and can be seen applied in PR #834.

I will submit a pull request for this shortly.

Seldaek:
Fixed by https://github.com/Seldaek/monolog/pull/1146

PR #1146, opened 2018-05-02
NewRelicHandler - handle Throwable in PHP7

pauluswebster:
Support for correct passing of Throwable to newrelic_notice_error() in PHP7 as outlined in #1145


Seldaek:
Thanks

PR #1148, opened 2018-05-13
Update LineFormatter.php replaceNewlines {

nickkuijpers:
Hello,

When enabling allowInlineLineBreaks and there is no { in the string, it returns a FALSE instead of a 0 so the function will not get triggered. I have filled in { in the first character of my string to fix this but i think it must be:

```
if (false === strpos($str, '{')) {
```
Instead of 
```
if (0 === strpos($str, '{')) {
```

or am i mistaken?

### DIFF
--- a/src/Monolog/Formatter/LineFormatter.php
+++ b/src/Monolog/Formatter/LineFormatter.php
@@ -157,7 +157,7 @@ class LineFormatter extends NormalizerFormatter
     protected function replaceNewlines(string $str): string
     {
         if ($this->allowInlineLineBreaks) {
-            if (0 === strpos($str, '{')) {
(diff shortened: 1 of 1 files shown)


Seldaek:
I believe this is by design so that we only do this for JSON lines (i.e. lines starting with `{`) as those are having \n and \r encoded as such.

I am not entirely sure though. I think this needs to be examined further as using false instead of 0 there would just enable it for lines that don't contain any json and that definitely won't help. I am not sure if we have a bug or not but I am fairly sure the fix isn't the right one.

nickkuijpers:
Thank you for responding.

I will use another approach untill.

PR #1153, opened 2018-05-21
Exceptionless handler implementation

israelito3000:
Hello, I recently work on a project which I had to integrate wordpress, exceptionless, I used to monolog and I  would like to contribute by adding this handler to submit logs to https://www.exceptionless.io (or any custom server where exceptionless is installed)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ dist: trusty
 php:
   - 7.0
   - 7.1
-  - nightly
 
 cache:
   directories:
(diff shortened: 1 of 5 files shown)


Seldaek:
Thanks, however maintaining all handlers inside this repo is kinda unsustainable for me so I will have to close this. Please instead publish such new handlers as individual packages on packagist.org and then add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages so it's easily discoverable.

Issue #1002, opened 2017-06-02
Let TestHandler::hasRecordThatContains methods also check the context (optional)


DanielSchwiperich:
I would find it helpful for unit tests if the hasRecordThatContains methods not only check the message but also the context. 
This would allow to have for example unit tests not only check that the correct log message is triggered but also that the correct data is written. 

instead of 
```php
    public function hasRecordThatContains($message, $level) {...}
```

```php
    public function hasRecordThatContains($message, $level, $testContext = false) {...}
```

Making it optional would also not break anything or make other stuff suddenly pass. 

I would try to provide a PR if the idea is generally accepted.

the-invisible-man:
I'm all for this, dealing with this myself so I can check that we're not logging items with sensitive information.

Seldaek:
Fixed by #1159

PR #1159, opened 2018-05-31
Make TestHandler::hasRecord assert context, not only message

p-blomberg:
The current implementation of hasRecord in the TestHandler only checks that the message is correct, not the context. This has also been reported in issue #1002 .

In this pull request, I'm changing the hasRecord method to also verify the context. However, I'm not sure if the hasRecordThatContains should also be modified, and if so, what that should actually mean.

I'm also adding two new tests to verify this behaviour.

### DIFF
--- a/src/Monolog/Handler/TestHandler.php
+++ b/src/Monolog/Handler/TestHandler.php
@@ -86,18 +86,16 @@ class TestHandler extends AbstractProcessingHandler
 
     public function hasRecord($record, $level)
     {
+        if (is_string($record)) {
+            $record = ['message' => $record];
+        }
+
         return $this->hasRecordThatPasses(function ($rec) use ($record) {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, would you mind rebasing the PR on the 1.x branch though?

p-blomberg:
> Thanks, would you mind rebasing the PR on the 1.x branch though?

Sure, I can do that. But what is the intention with the master branch? Will there be a Monolog 2 some day?

Seldaek:
Yes master is targetting 2.0, so if you want this available in 1.x I need it for the 1.x branch (which I then merge into master)

Seldaek:
Thanks, merged in 1.x and master as b381a973bcca4a2acb01dbe57e37bf5474fb7768

PR #1160, opened 2018-06-04
Use late static bindings in BrowserConsoleHandler

zaporylie:
Ref https://github.com/symfony/monolog-bundle/issues/152

TL;DR
BrowserConsoleHandler in a current state cannot be overridden because it is using `self::` instead of `static::` to access its members.

### DIFF
--- a/src/Monolog/Handler/BrowserConsoleHandler.php
+++ b/src/Monolog/Handler/BrowserConsoleHandler.php
@@ -43,11 +43,11 @@ class BrowserConsoleHandler extends AbstractProcessingHandler
     protected function write(array $record)
     {
         // Accumulate records
-        self::$records[] = $record;
+        static::$records[] = $record;
 
         // Register shutdown handler if not already done
-        if (!self::$initialized) {
-            self::$initialized = true;
+        if (!static::$initialized) {
+            static::$initialized = true;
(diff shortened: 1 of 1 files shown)


zaporylie:
Tests fails on:
```
1) Monolog\LoggerTest::testUseMicrosecondTimestamps with data set "with microseconds" (true, 'assertNotSame')
Failed asserting that two strings are not identical.
/home/travis/build/Seldaek/monolog/tests/Monolog/LoggerTest.php:537
2) Monolog\LoggerTest::testUseMicrosecondTimestamps with data set "without microseconds" (false, 'assertNotSame')
Failed asserting that two strings are not identical.
/home/travis/build/Seldaek/monolog/tests/Monolog/LoggerTest.php:537
FAILURES!
```
Seems to be unrelated?

zaporylie:
I can see that same errors are now part of 1.x HEAD. 

https://travis-ci.org/Seldaek/monolog/builds/388165231?utm_source=github_status&utm_medium=notification

Are you aware of that issue and, by any chance, know what's causing it?

Seldaek:
It's probably just a change in HHVM, don't worry about it I'm fairly positive it's unrelated to your change.

zaporylie:
Thank you @Seldaek

PR #1161, opened 2018-06-07
Improve the constructor of WebProcessor

gmponos:
Hello,

First of all note that since version 2.x is not tagged yet I believe that these changes can be made and they are BC.

Using these changes we ensure that we always pass arrays and not null making the code more strict.

### DIFF
--- a/src/Monolog/Processor/WebProcessor.php
+++ b/src/Monolog/Processor/WebProcessor.php
@@ -40,25 +40,27 @@ class WebProcessor
 
     /**
      * @param array|\ArrayAccess $serverData  Array or object w/ ArrayAccess that provides access to the $_SERVER data
-     * @param array|null         $extraFields Field names and the related key inside $serverData to be added. If not provided it defaults to: url, ip, http_method, server, referrer
+     * @param array              $extraFields Field names and the related key inside $serverData to be added. If not provided it defaults to: url, ip, http_method, server, referrer
      */
-    public function __construct($serverData = null, array $extraFields = null)
+    public function __construct($serverData = null, array $extraFields = [])
     {
-        if ($serverData === null || count($serverData) === 0) {
+        if (null === $serverData || 0 === count($serverData)) {
             $this->serverData = &$_SERVER;
-        } elseif (\is_array($serverData) || $serverData instanceof \ArrayAccess) {
+        } elseif (is_array($serverData) || $serverData instanceof \ArrayAccess) {
             $this->serverData = $serverData;
         } else {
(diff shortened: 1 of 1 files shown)


src/Monolog/Processor/WebProcessor.php
```diff
@@ -44,23 +44,21 @@ class WebProcessor
...
             throw new \UnexpectedValueException('$serverData must be an array or object implementing ArrayAccess.');
         }
 
-        if (null !== $extraFields) {
-            if (isset($extraFields[0])) {
-                foreach (array_keys($this->extraFields) as $fieldName) {
-                    if (!in_array($fieldName, $extraFields)) {
-                        unset($this->extraFields[$fieldName]);
-                    }
+        if (count($extraFields) === 0) {
```
stof: counting on null is not allowed on PHP 7.2

src/Monolog/Processor/WebProcessor.php
```diff
@@ -44,23 +44,21 @@ class WebProcessor
      */
     public function __construct($serverData = null, array $extraFields = null)
     {
-        if (null === $serverData) {
+        if ($serverData === null || count($serverData) === 0) {
```
stof: IMO, the previous signature is better:

- passing `null` means that Monolog will use $_SERVER
- passing an array or ArrayAccess object means it will be used

Now, you add a magic case where passing an empty array does not behave like passing a non-empty array

stof: This is broken when counting on an ArrayAccess object. It is perfectly valid to implement ArrayAccess without implementing Countable, and you even count here without doing any validation whether it is allowed.

stof:
oh, and that's a weird BC break for 2.x (WebProcessor is already released in 1.x)

Seldaek:
Not sure I see a clear benefit here sorry.

gmponos:
> Not sure I see a clear benefit here sorry.

If someone passes an empty array on one of the constructors arguments it will not be taken to account on these if: https://github.com/Seldaek/monolog/blob/master/src/Monolog/Processor/WebProcessor.php#L45

gmponos:
So if someone calls `new WebProcessor([], []);` will this work as expected?

xabbuh:
@gmponos Passing an empty array currently is the only way to not have any server data set. With the changes proposed here this behaviour would change and `$_SERVER` would be reused which means that one has to pass an object implementing `ArrayAccess` instead to preserve the old behaviour. I don't think that's really desired.

gmponos:
>  I don't think that's really desired.

OK then 👍

PR #1162, opened 2018-06-07
Test PHP 7.2 on travis-ci.

tomzx:


Seldaek:
Thanks but it should target the 1.x branch

Seldaek:
Actually never mind #1120 already does the same on the right branch.

PR #1164, opened 2018-06-08
Remove from slack handler on constructing the SlackRecord to pass the Formatte

gmponos:
During constructiong the SlackHandler a Formatter has not been set.

So it does not make sense passing the Formatter as an argument to SlackRecord. It is just confusing and makes people wonder where this has been set?


Seldaek:
Thanks

Issue #1105, opened 2018-01-17
Request: move to somewhere the hardcoded swiftmailer formatter


Fewald:
We've got an issue with default monolog formatters which leads us to custom formatters everywhere. Unfortunately, this one is hardcoded
https://github.com/Seldaek/monolog/blob/7b992836275e09ed63c63fe33ca9993e515e6c5d/src/Monolog/Handler/SwiftMailerHandler.php#L73

Will it be ok for you to have a getter for that "subject" formatter? If so, I can do PR

Seldaek:
Sure feel free to add a getter for it that takes the $message->getSubject() as argument or even the whole $message instance.

PR #1166, opened 2018-06-11
A getter for SwiftMailerHandler's subject formatter

Fewald:
https://github.com/Seldaek/monolog/issues/1105

This PR moves an instance creation of SwiftMailerHandler's subject formatter to a getter. 

Motivation: to get LineFormatter usage to be overwritable. 
Pretty minor change, still, can be useful.

I think there is no reason to cover this getter with a unit test but let me know if you'd like to.

P.S.
Why do I want to avoid LineFormatter usage? Rare cases, but for instance: to avoid calling this __toString() https://github.com/Seldaek/monolog/blob/master/src/Monolog/Formatter/NormalizerFormatter.php#L138 of this zf1 class https://github.com/zendframework/zf1/blob/master/library/Zend/Controller/Response/Abstract.php#L32 who thinks that __toString() method call is a reason to send headers
https://github.com/zendframework/zf1/blob/master/library/Zend/Controller/Response/Abstract.php#L768,%20L794

### DIFF
--- a/src/Monolog/Handler/SwiftMailerHandler.php
+++ b/src/Monolog/Handler/SwiftMailerHandler.php
@@ -12,6 +12,7 @@
 namespace Monolog\Handler;
 
 use Monolog\Logger;
+use Monolog\Formatter\FormatterInterface;
 use Monolog\Formatter\LineFormatter;
 use Swift_Message;
 use Swift;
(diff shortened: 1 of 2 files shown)


Fewald:
@Seldaek There was a failed unrelated test: I fixed him

Seldaek:
Thanks, merged as 6e1793e and c1b610a

PR #1169, opened 2018-06-18
Better PSR-3 message formatting

lstrojny:
Successor of #972 

Does two things:

 - Include milliseconds in the default date format
 - Show actual content of the array


Seldaek:
Thanks

PR #1170, opened 2018-06-18
Include stacktraces for "previous" excpetions in LineFormatter

### LINKED ISSUE #1086: LineFormatter does not include stacktraces for "previous" exceptions
I am using the `LineFormatter` to log exceptions with stack traces.

However, `previous` exceptions are only logged with their class name, file, code and message, but no stack trace included. In many cases I have seen the stack trace of the "root" exception would be the interesting information I'd need.

So, short question: Would you support/merge a (probably trivial) PR to include stacktraces for all `previous` exceptions as well?

Should this be changed in `LineFormatter` only, or are there any other/better places?

mpdude:
Fixes #1086.


Seldaek:
Thanks

PR #1179, opened 2018-07-20
Add support for browser and OS proper recognition

mleczakm:
By default, browser and OS are not properly recognized in sentry when logging from Monolog, even if proper http-agent data is provided. Only runtime options are visible:
image
With this simple fix, when we provide additional data ([using processor](https://symfony.com/doc/current/logging/processors.html)) under `contexts` index (same as in JSON generated from JS SDK) we have it :)
image


mleczakm:
Because of no reaction I've created own bundle providing this functionality (and a bit more) - https://github.com/mleczakm/monolog-sentry-bundle

PR #1180, opened 2018-07-20
Add support for browser and OS proper recognition

mleczakm:
Same as in #1179 :) but for 1.9 branch


mleczakm:
Because of no reaction I've created own bundle providing this functionality (and a bit more) - https://github.com/mleczakm/monolog-sentry-bundle

PR #1182, opened 2018-08-02
Cast slack value to string

barryvdh:
Make sure the value is a string. Eg. for an Exception object.


barryvdh:
Ehr, not really sure what is crashting the tests. But I think that's not my fault.

barryvdh:
I think this is fixed with the latest version anyways. Not sure where the error lies.

PR #1190, opened 2018-08-17
Fix displaying anonymous classes

nicolas-grekas:
Loggin e.g. `RuntimeException@anonymous` instead of `�class@anonymous�/.../src/Controller/HelloController.php0x7f8da350c15c`


nicolas-grekas:
See https://github.com/symfony/symfony/pull/28218 btw

Seldaek:
Patch looks good but would be cool to have this as a static getClassName method on an util class than having this copied all over the place though. If you have a chance to refactor this it'd be great.

nicolas-grekas:
Updated. Good enough?

PR #1195, opened 2018-09-24
Updated @see URL for GELF formatter.

mikeschinkel:
The @see URL for GELF was broken so I found the new URL and updated it in the PHPDoc.


Seldaek:
Thanks

PR #1204, opened 2018-10-12
Add ProcessorInterface, an optional one to allow labelling Monolog processors

nicolas-grekas:
Would replace https://github.com/symfony/symfony/pull/27801, as an interface better belongs to its original package than to a bridge.


Seldaek:
Thanks

PR #1210, opened 2018-10-28
[RavenHandler] Add environment and breadcrumbs support

ottaviano:
Add Environment and Breadcrumbs support
- https://docs.sentry.io/learn/environments/
- https://docs.sentry.io/learn/breadcrumbs/


Seldaek:
Thanks

mleczakm:
Environments were already supported using tags. Breadcrumbs aren't automatically cleaned after writing to Sentry, so it is possible to send it twice. Since we already have a lot BC breaks inside 2.0 version - stop writing batch logs int context, add them into breadcrumbs with proper info and reset them after sending, just like here https://github.com/mleczakm/monolog-sentry-bundle/blob/master/Handler/Raven.php#L170 .

Seldaek:
@mleczakm there is resetBreadcrumbs() to reset them when done.. I don't think monolog can make a decision here as to when those breadcrumbs don't apply anymore, but I am not super familiar with sentry to be honest.

As for the environment, I am not sure what you suggest we do.

PR #1214, opened 2018-11-02
[RavenHandler] Add support for contexts

rvalitov:
Originates from #1179, and problem discussion is available at [Sentry](https://github.com/getsentry/sentry/issues/8555).

Sentry provides [contexts](https://docs.sentry.io/clientdev/interfaces/contexts/) to add extra data to the log. But it's not supported in current release of RavenHandler. As a result, in Sentry we see empty info like this:

image

This PR fixes this, so that user can pass `contexts` field in the record, as a result we will see something like this:

image

Users can pass information to Sentry like this:

```php
$logger = new Monolog\Logger('YOUR_LOG_NAME');
$client = new Raven_Client('YOUR_SECRET_KEY');
$stream = new RavenHandler($client);
$log->addNotice('Some information', [
    'contexts' => [
        'os' => [
            'name' => 'Windows',
            'version' => '10'
        ]
    ]
]);
```

It's also possible to pass the `contexts` using [processors](https://github.com/Seldaek/monolog/blob/master/doc/02-handlers-formatters-processors.md#processors). Example:

```php
$log->pushProcessor(function ($record) {
    $record['contexts'] = [
        'os' => [
            'name' => 'Windows',
            'version' => '10'
        ]
    ];
    return $record;
});
```

### DIFF
--- a/src/Monolog/Handler/RavenHandler.php
+++ b/src/Monolog/Handler/RavenHandler.php
@@ -182,10 +182,12 @@ class RavenHandler extends AbstractProcessingHandler
             unset($record['contexts']);
         }
         if (!empty($record['extra']['contexts'])) {
-            if (!empty($options['contexts']))
+            if (!empty($options['contexts'])) {
                 $options['contexts'] = array_merge($options['contexts'], $record['extra']['contexts']);
-            else
+            }
+            else {
                 $options['contexts'] = $record['extra']['contexts'];
+            }
             unset($record['extra']['contexts']);
         }
 
--- a/src/Monolog/Handler/RavenHandler.php
+++ b/src/Monolog/Handler/RavenHandler.php
@@ -177,6 +177,18 @@ class RavenHandler extends AbstractProcessingHandler
             $options['extra']['extra'] = $record['extra'];
         }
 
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/RavenHandler.php
```diff
@@ -177,6 +177,18 @@ protected function write(array $record): void
             $options['extra']['extra'] = $record['extra'];
         }
 
+        if (!empty($record['contexts'])) {
+            $options['contexts'] = $record['contexts'];
+            unset($record['contexts']);
+        }
+        if (!empty($record['extra']['contexts'])) {
+            if (!empty($options['contexts']))
```
mleczakm: Missing brackets.

mleczakm:
Useful contribution (I could remove some part of my bundle ;) if it only were merged, thanks a lot) but - missing tests.

src/Monolog/Handler/RavenHandler.php
rvalitov: Thank you for, new commit pushed.

rvalitov:
I'll try to add the tests later...

PR #1215, opened 2018-11-05
Remove version of Symfony

samnela:


doc/01-usage.md
```diff
@@ -18,7 +18,7 @@ composer require monolog/monolog
 ```
 
 If you do not use Composer, you can grab the code from GitHub, and use any
-PSR-0 compatible autoloader (e.g. the [Symfony2 ClassLoader component](https://github.com/symfony/ClassLoader))
+PSR-0 compatible autoloader (e.g. the [Symfony ClassLoader component](https://github.com/symfony/ClassLoader))
```
xabbuh: I would remove the reference to the ClassLoader component entirely as it is no longer maintained.

Seldaek:
Thanks

PR #1217, opened 2018-11-06
Add newrelic_end_transcation() at the end of the write method

Kolesar:
Calling `newrelic_end_transaction()` is required for every transaction.

### DIFF
--- a/src/Monolog/Handler/NewRelicHandler.php
+++ b/src/Monolog/Handler/NewRelicHandler.php
@@ -117,6 +117,8 @@ class NewRelicHandler extends AbstractProcessingHandler
                 }
             }
         }
+
+        newrelic_end_transaction();
     }
 
     /**


Seldaek:
It doesn't seem to be required to call end transaction after every write operation that's done.

From reading https://docs.newrelic.com/docs/agents/php-agent/php-agent-api/newrelic_end_transaction it sounds to me like this should rather be done in the new reset() method implementing ResettableInterface. And it probably should call `newrelic_end_transaction();` then right after call `newrelic_start_transaction($this->transactionName ?: ini_get('newrelic.appname'));`. to start the next.

Kolesar:
Hi @Seldaek,

Thanks for the quick reply. I tried to implement as you suggest on my local/dev environment, but it does not work. It seems,`reset()` method will not be called anytime.

Seldaek:
No, you have to call Logger::reset() when your application ends a cycle (either a request, or a job). What's your use case anyway?

Kolesar:
Huh,

It will be hard in my case. I have different channels in my configuration, like:
```
        newrelic:
            type: newrelic
            app_name: "My Service Name"
            channels: ["app", "workers", "context", "php", "extensions"]
            formatter: MyProject\Monolog\Formatter\NewRelicFormatter
            level: debug
            verbosity_levels:
                VERBOSITY_NORMAL: DEBUG
```

And from another side `newrelic_end_transaction()` does not have any parameter to know which transaction we should close. I tried this in many different cases and it works (for now) only as you can see in my PR.

Any idea is welcome, of course :)

Seldaek:
I have no idea what you are up to though so hard to help. Please explain the use case.. Why do you need to call newrelic_end_transaction at all?

jderusse:
In NewRelic, a transactions wraps many metrics (errors, external calls, sql queries, cpu time, call stack, ...) a single log is one of those metrics. When an app log an Error (even critical), it does not mean the transaction is ended, the application could logs several error in the same call (for instance, a transaction could be a command that process a CSV file and logs error on each line), or the application could call external services (ie. call external payment service to cancel an order, ...), etc...

IMHO this is not the responsibility of the logger to start or ends a transactions. It should be done by the "request handler" of your application (ie. the Kernel or EventListener in a  Symfony application or a the "Processor" in a AMQP consumer)

Seldaek:
Ok thanks @jderusse that kinda matches my understanding, will close both these issues then.

PR #1223, opened 2018-11-13
Exclude files and folder from release archive

ankurk91:
Hi,

Let's not publish all those file.
`.gitattributes` can be used to exclude files from final archive that is downloaded by composer.

This file content is same as `master` branch now.

Thanks.


Seldaek:
Thanks

PR #1227, opened 2018-11-17
Reset test handler

gmponos:
Shouldn't the test handler be reset as well?

### DIFF
--- a/src/Monolog/Handler/TestHandler.php
+++ b/src/Monolog/Handler/TestHandler.php
@@ -79,6 +79,11 @@ class TestHandler extends AbstractProcessingHandler
         $this->recordsByLevel = array();
     }
 
+    public function reset()
+    {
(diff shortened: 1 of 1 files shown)


Seldaek:
Seems reasonable yes :) Thanks

Seldaek:
Seems to break tests though..

gmponos:
I can't seem to be able to fix the tests though.. I would like some help here.. if the test that fails make sense to you in the way it is write now..

Seldaek:
Fixed in https://github.com/Seldaek/monolog/commit/a335f6f1a564f43d2c63f2eeeca381e6c78f9006 - those tests rely on reset happening at the buffer/fingerscrossed handler but at that point they flush their buffer to their inner test handler so if that one gets reset then it just doesn't work.

PR #1228, opened 2018-11-17
Add visibility to constants

gmponos:
[PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md) Just for reference if it gets accepted: 

> Visibility MUST be declared on all constants if your project PHP minimum version supports constant visibilities (PHP 7.1 or later).

Personally I like it also this way better. Maybe you should revise some of them if they should be public but this would be a BC. Since you are moving to v2.0 it's a good chance.


Seldaek:
Thanks, adjusted visibilities in bb99e4c

Issue #1197, opened 2018-09-25
Monolog\TestCase class should be accessible for out of tree tests


yann-soubeyrand:
When writing a custom Monolog processor as part of a Symfony application, one cannot test this processor using the Monolog\TestCase base class. Indeed this class is part of the autoload-dev section of the composer.json file and this section is only used if the composer.json file is the root one (which is not the case when Monolog is used as part of a Symfony project).

Here is an example of such a test:
```
<?php
namespace Tests\Bridge\FrontBundle\Logger;

use Bridge\FrontBundle\Logger\RequestIdProcessor;
use Monolog\TestCase;
use Symfony\Component\HttpFoundation\Request;
use Symfony\Component\HttpFoundation\RequestStack;
class RequestIdProcessorTest extends TestCase
{
    private const REQUEST_ID = 'test-request-id';

    public function testValidateRequestIdProcessorWhenRequestIdHeaderIsPresent()
    {
        $request = new Request([], [], [], [], [], ["HTTP_X_REQUEST_ID" => self::REQUEST_ID]);
        $requestStack = new RequestStack();
        $requestStack->push($request);
        $requestIdProcessor = new RequestIdProcessor($requestStack);
        $record = $requestIdProcessor($this->getRecord());
        $this->assertArrayHasKey('request_id', $record['extra']);
        $this->assertTrue($record['extra']['request_id'] === self::REQUEST_ID);
    }
    public function testValidateRequestIdProcessorWhenRequestIdHeaderIsMissing()
    {
        $request = new Request();
        $requestStack = new RequestStack();
        $requestStack->push($request);
        $requestIdProcessor = new RequestIdProcessor($requestStack);
        $record = $requestIdProcessor($this->getRecord());
        $this->assertArrayHasKey('request_id', $record['extra']);
        $this->assertTrue($record['extra']['request_id'] === '-');
    }
}
```

Seldaek:
Monolog\Test\TestCase is available in src/ as of 2.0 (master branch only right now).

PR #1229, opened 2018-11-17
Added a test logger

gmponos:
Hello,

I have created a TestLogger. I hope that you feel that this fits inside your package. I know that Monolog has a `TestHandler` but I believe this is better for the following reasons:

- Simpler initialization of Logger inside tests. At the moment in order to perform my tests I need to initialize a `Logger` and then a `TestHandler`.. More properties that I have to maintain inside my code... etc etc.
- Faster tests since it does minimal work. It skips all the extra work that it is made by `Logger` class.
- Inside the TestHandler you have the `getRecords` function. Even if I retrieve all the records when I need to do simple asserts I can not.. Check this:

```
$this->assertSame(['level' => 'warning', 'message' => 'that', ['context' => []], $records[0]];
```

The record contains a lot of extra information that in my case they are not needed.

Even in your tests you had to [do unset](https://github.com/Seldaek/monolog/blob/master/tests/Monolog/Handler/TestHandlerTest.php#L53). So I don't need to have all these information like `formatted` etc etc.

All I need to test is that the API of the `LoggerInterface` class is called correctly.

### DIFF
--- a/src/Monolog/TestLogger.php
+++ b/src/Monolog/TestLogger.php
@@ -1,0 +1,158 @@
+<?php
+
+namespace Monolog;
+
+use Psr\Log\AbstractLogger;
+
+/**
+ * Used for testing purposes.
+ *
+ * It records all records and gives you access to them for verification.
+ *
+ * @method bool hasEmergency($record)
+ * @method bool hasAlert($record)
+ * @method bool hasCritical($record)
+ * @method bool hasError($record)
+ * @method bool hasWarning($record)
+ * @method bool hasNotice($record)
+ * @method bool hasInfo($record)
+ * @method bool hasDebug($record)
+ *
+ * @method bool hasEmergencyRecords()
+ * @method bool hasAlertRecords()
+ * @method bool hasCriticalRecords()
+ * @method bool hasErrorRecords()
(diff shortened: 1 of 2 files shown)


Seldaek:
Could you move this inside a Monolog\Test\ namespace (but still in src/)? If you also wanna help out by taking care of https://github.com/Seldaek/monolog/issues/1197 at the same time that'd be even better :)

Seldaek:
On the other hand, I kinda wonder if this doesn't belong more in the psr/log package as it doesn't have anything monolog specific really

gmponos:
O had the same dilemma also if I should commit it here or on the PSR. Checking PSR now again I just saw this namespace `Psr\Log\Test` so yes I believe it could fit in there. check the PR https://github.com/php-fig/log/pull/57

Thank you.

Seldaek:
Ok gonna close this here then.

gmponos:
Hi @Seldaek do you think that this could be added in here now that the TestLogger is removed from `psr/log` ?

stof:
@gmponos as there is still nothing specific to Monolog in that TestLogger (the monolog-specific way is to use the TestHandler of Monolog), you can maintain that in its own package.

PR #1239, opened 2018-12-09
Make CURL util final and internal

gmponos:
Hello,

Should these be final and internal to avoid BC changes in the future.

I am not sure if you would be willing to accept a change like this... 

Also I have one more suggestion to make.. how would you feel about removing the Curl Util and moving to a solution like HTTPlug now that PSR-18 is accepted?


Seldaek:
I'd rather not add a dependency for this as it makes usage more complex and the current solution seems to have worked fine for the last x years. Marking as internal/final sounds reasonable though, thanks.

PR #1240, opened 2018-12-09
Suggest JSON extension

gmponos:
I believe that this extension should be in the required ones...  🤔 

Although as a counter argument someone might say that he wants to use Monolog without using Handlers/Formatters that JSON encode messages that's why I added it at `suggest` for now.. 

Let me know where do you believe it belongs...


gmponos:
just for reference.. https://twitter.com/gmponos/status/1073611157792006144

Seldaek:
I think it's kinda pointless as json ext is virtually always there. Those that don't have it probably did so explicitly and can figure out their own problems.

PR #1241, opened 2018-12-09
Error handler strict comparison fix

gmponos:
According to official PHP documentation 

> Returns the value of the configuration option as a string on success, or an empty string for null values. Returns FALSE if the configuration option doesn't exist. 

This change seems kinda big... and I don't know how to test it

### DIFF
--- a/src/Monolog/ErrorHandler.php
+++ b/src/Monolog/ErrorHandler.php
@@ -58,9 +58,6 @@ class ErrorHandler
      */
     public static function register(LoggerInterface $logger, $errorLevelMap = [], $exceptionLevelMap = [], $fatalLevel = null): self
     {
-        //Forces the autoloader to run for LogLevel. Fixes an autoload issue at compile-time on PHP5.3. See https://github.com/Seldaek/monolog/pull/929
-        class_exists('\\Psr\\Log\\LogLevel', true);
-
         $handler = new static($logger);
         if ($errorLevelMap !== false) {
             $handler->registerErrorHandler($errorLevelMap);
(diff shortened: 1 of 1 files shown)


src/Monolog/ErrorHandler.php
```diff
@@ -58,9 +58,6 @@ public function __construct(LoggerInterface $logger)
      */
     public static function register(LoggerInterface $logger, $errorLevelMap = [], $exceptionLevelMap = [], $fatalLevel = null): self
     {
-        //Forces the autoloader to run for LogLevel. Fixes an autoload issue at compile-time on PHP5.3. See https://github.com/Seldaek/monolog/pull/929
-        class_exists('\\Psr\\Log\\LogLevel', true);
-
```
gmponos: I removed also this... I guess the bug from back then it is fixed now.. isn't it?

Seldaek: Yup good catch.

src/Monolog/ErrorHandler.php
```diff
@@ -171,7 +169,7 @@ public function handleException($e)
             call_user_func($this->previousExceptionHandler, $e);
         }
 
-        if (!headers_sent() && ini_get('display_errors') === 0) {
+        if (!headers_sent() && ini_get('display_errors') === '0') {
```
Seldaek: Probably safer to simply have `&& !ini_get('display_errors')`?

Seldaek: Yes sounds like it should.

Seldaek:
Thanks

PR #1243, opened 2018-12-09
Docblock fixes

gmponos:
As WIP to see if I will find more...

### DIFF
--- a/src/Monolog/Formatter/HtmlFormatter.php
+++ b/src/Monolog/Formatter/HtmlFormatter.php
@@ -37,7 +37,7 @@ class HtmlFormatter extends NormalizerFormatter
     ];
 
     /**
-     * @param ?string $dateFormat The format of the timestamp: one supported by DateTime::format
+     * @param string|null $dateFormat The format of the timestamp: one supported by DateTime::format
      */
     public function __construct(?string $dateFormat = null)
     {
@@ -50,6 +50,7 @@ class HtmlFormatter extends NormalizerFormatter
      * @param string $th       Row header content
      * @param string $td       Row standard cell content
      * @param bool   $escapeTd false if td content must not be html escaped
+     * @return string
      */
     protected function addRow(string $th, string $td = ' ', bool $escapeTd = true): string
     {
--- a/src/Monolog/Formatter/HtmlFormatter.php
+++ b/src/Monolog/Formatter/HtmlFormatter.php
@@ -50,7 +50,6 @@ class HtmlFormatter extends NormalizerFormatter
      * @param string $th       Row header content
      * @param string $td       Row standard cell content
      * @param bool   $escapeTd false if td content must not be html escaped
-     * @return string
      */
     protected function addRow(string $th, string $td = ' ', bool $escapeTd = true): string
     {
@@ -69,7 +68,7 @@ class HtmlFormatter extends NormalizerFormatter
      * @param  int    $level Error level
      * @return string
      */
-    protected function addTitle(string $title, int $level)
+    protected function addTitle(string $title, int $level): string
     {
         $title = htmlspecialchars($title, ENT_NOQUOTES, 'UTF-8');
 
--- a/src/Monolog/Handler/BufferHandler.php
+++ b/src/Monolog/Handler/BufferHandler.php
@@ -44,7 +44,7 @@ class BufferHandler extends AbstractHandler implements ProcessableHandlerInterfa
     {
         parent::__construct($level, $bubble);
         $this->handler = $handler;
-        $this->bufferLimit = (int) $bufferLimit;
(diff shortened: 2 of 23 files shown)


src/Monolog/Handler/BufferHandler.php
```diff
@@ -44,7 +44,7 @@ public function __construct(HandlerInterface $handler, int $bufferLimit = 0, $le
     {
         parent::__construct($level, $bubble);
         $this->handler = $handler;
-        $this->bufferLimit = (int) $bufferLimit;
+        $this->bufferLimit = $bufferLimit;
```
gmponos: Note that I fixed this also.. not needed to be converted to int again.

src/Monolog/Handler/RotatingFileHandler.php
```diff
@@ -47,7 +47,7 @@ class RotatingFileHandler extends StreamHandler
     public function __construct(string $filename, int $maxFiles = 0, $level = Logger::DEBUG, bool $bubble = true, ?int $filePermission = null, bool $useLocking = false)
     {
         $this->filename = $filename;
-        $this->maxFiles = (int) $maxFiles;
+        $this->maxFiles = $maxFiles;
```
gmponos: Note that I fixed this also.. not needed to be converted to int again.

src/Monolog/Processor/UidProcessor.php
```diff
@@ -24,7 +24,7 @@ class UidProcessor implements ProcessorInterface, ResettableInterface
 
     public function __construct(int $length = 7)
     {
-        if (!is_int($length) || $length > 32 || $length < 1) {
+        if ($length > 32 || $length < 1) {
```
gmponos: Note that I fixed this also.. not needed to check for int.

gmponos:
That's what I've found. :)

src/Monolog/ResettableInterface.php
```diff
@@ -27,5 +27,10 @@
  */
 interface ResettableInterface
 {
+    /**
+     * Function that executes the resetting of a Processor
+     *
+     * @return void
```
gmponos: I wanted to create another PR adding void return type to reset function but I guess this would mean a huge BC for symfony... right ? ping @lyrixx

lyrixx: In Symfony, we decided to not add `void` as return type to make it easier to change our mind later if needed.

src/Monolog/Formatter/HtmlFormatter.php
```diff
@@ -50,6 +50,7 @@ public function __construct(?string $dateFormat = null)
      * @param string $th       Row header content
      * @param string $td       Row standard cell content
      * @param bool   $escapeTd false if td content must not be html escaped
+     * @return string
```
lyrixx: This is useless, there is already a PHP type hint on the method.
This should be reverted,

gmponos: Yes.. I added there for consistency with other docblocks.. I saw that where ever a docblock existed it had the return typehint.. I will revert...

src/Monolog/ResettableInterface.php
gmponos: I will revert then...

src/Monolog/Formatter/HtmlFormatter.php
Seldaek: It's only added when it provides a description or sometimes because it was there in 1.x and was missed when converting. e.g. addTitle below has a return that should be moved to the method signature.

src/Monolog/ResettableInterface.php
Seldaek: I decided not to add it as it's brand new in 1.x and that would mean breaking new implementations that could otherwise be compatible with 1.x and 2.x. Didn't really seem worth it. I'm fine having it in the docblock though for clarity.

src/Monolog/ResettableInterface.php
```diff
@@ -27,5 +27,10 @@
  */
 interface ResettableInterface
 {
+    /**
+     * Function that executes the resetting of a Processor
```
Seldaek: This isn't Processor specific, and really this whole sentence adds no value, I'd remove it.

src/Monolog/Formatter/HtmlFormatter.php
gmponos: > addTitle below has a return that should be moved to the method signature.

I did that too.

>  I will revert...

reverted.

src/Monolog/ResettableInterface.php
gmponos: Removed.

Seldaek:
Thanks

PR #1244, opened 2018-12-09
Parameter formatter was never set.. would cause fatal

gmponos:
The parameter `$formatter` was never set and most probably it would have caused a fatal error.


Seldaek:
Thanks, definitely a copy paste from the setter :)

PR #1249, opened 2018-12-10
Drop sup for swiftmailer v5

gmponos:
According to https://github.com/swiftmailer/swiftmailer/pull/1154


Seldaek:
Rather not do this, as the 3 extra lines of code really don't hurt much to let older projects update to monolog v2 easily.

PR #1255, opened 2018-12-13
Update README link of Python Logbook

Zarthus:
The old URL is no longer in use, serving "Logbook's documentation is maintained elsewhere." and a redirect.

Convieniently also converts the URL to HTTPS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,5 +98,5 @@ Monolog is licensed under the MIT License - see the `LICENSE` file for details
 
 ### Acknowledgements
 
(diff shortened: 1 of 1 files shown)


localheinz:
@Zarthus 

Maybe this should go into [`1.x`](https://github.com/Seldaek/monolog/tree/1.x)?

Zarthus:
It can be backported (or just directly committed to) in either direction, it's not a particularly complex change.

If you'd like I can take care of targeting 1.x instead, but you're welcome to do it yourself as well :)

Seldaek:
Closed by #1266 - thanks!

PR #1260, opened 2018-12-16
Throw exception if int level does not exist in array of levels

gmponos:
- [ ] Need to add tests.
- [ ] Fix failing tests.


gmponos:
There are some tests that are using custom levels.. So I need to know if allowing custom int was by design or I must proceed fixing them...

Seldaek:
This isn't needed, this function is only meant to normalize strings. Custom levels can be added if someone so wishes.

PR #1267, opened 2018-12-26
Use object as message on docblock for PSR functions

gmponos:
Driven by this: https://github.com/Seldaek/monolog/pull/1260

I looked at the PSR-3 to see if custom levels are allowed.. and yes they are... but on the next lines I saw this:

> Every method accepts a string as the message, or an object with a __toString() method. Implementors MAY have special handling for the passed objects. If that is not the case, implementors MUST cast it to a string

An alternative could be to get rid of the docblocks on these methods and depend on the docblocks of PSR.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -458,9 +458,9 @@ class Logger implements LoggerInterface, ResettableInterface
      *
      * This method allows for compatibility with common interfaces.
      *
-     * @param mixed  $level   The log level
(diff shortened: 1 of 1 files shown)


gmponos:
Opened one PR here too 

https://github.com/php-fig/log/pull/60

Seldaek:
Closing as per my comments on https://github.com/php-fig/log/pull/60

PR #1268, opened 2018-12-26
[WIP] PopHandler and PopProcessor may return null

gmponos:
Driven by this: https://github.com/Seldaek/monolog/pull/1267

I used PHPStan to see if object will cause any issue... Level 7 discovered the following probable bug...

Thanks @ondrejmirtes for his great work on PHPStan...

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -184,7 +184,7 @@ class Logger implements LoggerInterface, ResettableInterface
      *
      * @throws \LogicException If empty handler stack
      */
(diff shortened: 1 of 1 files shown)


gmponos:
A test will prove it... marking it as [WIP]

gmponos:
```
         if (!$this->processors) {
            throw new \LogicException('You tried to pop from an empty processor stack.');
        }
```


@ondrejmirtes  👍 

how did I missed that? 🤦‍♂️

Issue #675, opened 2015-10-30
Why forcing $context to be an array ?


dmidz:
I am wondering why forcing $context to be an array, because such, even writing a custom handler, it forces to record an array, and then eventually transform the data in the handler.
For example I wrote my own browser handler, which write javascript console.* and important, use JMSSerializer, so it respect entities' properties exposure, like :
```php
// BrowserHandler
$serializer->serialize($record['context'], 'json');
// Controller
$logger->debug('anyentity', $entity);
```
But I am forced to do like :
```php
// Controller
$logger->debug('anyentity', [$entity]);// wrap the context in an array
// BrowerHandler
$arr = $record['context'];
if(count($arr) === 1) $arr = $arr[0];// extract the first index if array contains only one element
$serializer->serialize($arr, 'json');
```
So now it outputs well 
```javascript
console.log("anyentity", {"exposedproperty":"foo","exposedproperty2":"bar"});
```

Seldaek:
There are a few reason: standardization and to make sure we can always access it as an array, it just builds common assumptions that make it a bit easier to write code. Plus it is required by the PSR-3 interface as well so it's unlikely we'll change that. And finally IMO it helps you write more readable logs if you are forced to name things and not just pass values without names (like you are doing, I would prefer passing it like `['entityName' => $entity]`, then maybe you can loop over all the context data and pass any object from the `Foo\Entity\..` namespace into the serializer and let the rest be serialized using the usual monolog NormalizerFormatter::normalize?).

dmidz:
Thank you for the clarification, I thought well it was about standardization, and then because it is used by a so wide variety of output possibilities, so it is more easy and less code everywhere when always expecting the input as array. Yes I could parse first level of context and check if instance of Entity and so pass to serializer, code in the handler is not a problem, it is more where I use the logger, I had to bypass in each place for wrapping the context in an array, but well...I could also extends the logger or insert a service in between and bypass there if I acually need it...
Anyway, Monolog is so usefull !

PR #1272, opened 2019-01-02
Support indexed context arrays

purplekrayons:
To provide some context, I am trying to create a means of passing a method's classname into my logs as I find it quite handy. My aim was to make this as visually clean as I could in my code, so I don't want to use `sprintf()` or any other means of formatting the message. To start, here's what im passing as my log arguments:

`$this->logger->info("Hey there! I'm just testing!", [__CLASS__]);`

To those unfamiliar, `__CLASS__` is a magic constant that returns the classname, including the namespace.

By default, Monolog will stringify the `context` array. This produces a log record which looks like this:
_(note: i've custom formatted the log message so that the context appears before the message)_

`[2019-01-02 02:06:12] INFO: [["LFLAPI\\Services\\Ecom\\SomeService"]] Hey there! I'm just testing!`

I don't like how this appears in my logs, so I opted to make use of the `pushHandler()` to pull out just the first element, which is the string by itself.

```php
$logger->pushProcessor(function ($record) {
    $record['context'] = $record['context'][0];
    return $record;
});
```

This produces the log results im after:

`[2019-01-02 02:09:02] INFO: [LFLAPI\Services\Ecom\SomeService] Hey there! I'm just testing!`

But it also triggers a warning:

`PHP Warning:  Invalid argument supplied for foreach() in /var/www/api/vendor/monolog/monolog/src/Monolog/Formatter/LineFormatter.php on line 81`

If I've analyzed this correctly, it occurs because Monolog expects an associative array, whereas I am passing an indexed array. I would like to propose including support for indexed arrays by including a function that can determine if an array is indexed, or associative:

```php
protected $isAssoc;

public function isAssoc(array $record)
{
    $this->isAssoc = count(array_filter(array_keys($record), 'is_string')) > 0;
}

if ($this->isAssoc) {
    foreach ($vars['context'] as $var => $val) {
        if (false !== strpos($output, '%context.'.$var.'%')) {
            $output = str_replace('%context.'.$var.'%', $this->stringify($val), $output);
            unset($vars['context'][$var]);
        }
    }
}
```

I feel this adds value to Monolog by allowing for more customization, and ultimately, cleaner user-code. Here's a quick side-by-side comparison of what I was aiming to avoid:

```php
$this->logger->info("Hey there! I'm just testing!", [__CLASS__]);
$this->logger->info(sprintf("[%s] Hey there! I'm just testing!", __CLASS__));

$this->logger->info("Hey there! I'm just testing!", [$var]);
$this->logger->info(sprintf("[%s] Hey there! I'm just testing!", $var));
```

### DIFF
--- a/src/Monolog/Formatter/LineFormatter.php
+++ b/src/Monolog/Formatter/LineFormatter.php
@@ -29,6 +29,7 @@ class LineFormatter extends NormalizerFormatter
     protected $allowInlineLineBreaks;
     protected $ignoreEmptyContextAndExtra;
     protected $includeStacktraces;
+    protected $isAssoc;
 
     /**
      * @param string|null $format                     The format of the message
@@ -62,6 +63,11 @@ class LineFormatter extends NormalizerFormatter
         $this->ignoreEmptyContextAndExtra = $ignore;
     }
 
+    public function isAssoc(array $record)
+    {
+        $this->isAssoc = count(array_filter(array_keys($record), 'is_string')) > 0;
+    }
+
     /**
      * {@inheritdoc}
      */
@@ -78,10 +84,12 @@ class LineFormatter extends NormalizerFormatter
             }
         }
 
-        foreach ($vars['context'] as $var => $val) {
-            if (false !== strpos($output, '%context.'.$var.'%')) {
-                $output = str_replace('%context.'.$var.'%', $this->stringify($val), $output);
-                unset($vars['context'][$var]);
+        if ($this->isAssoc) {
+            foreach ($vars['context'] as $var => $val) {
+                if (false !== strpos($output, '%context.'.$var.'%')) {
+                    $output = str_replace('%context.'.$var.'%', $this->stringify($val), $output);
+                    unset($vars['context'][$var]);
+                }
             }
         }
 


purplekrayons:
Note: Build failing because of this unit test:
https://github.com/Seldaek/monolog/blob/master/tests/Monolog/Formatter/LineFormatterTest.php#L94

purplekrayons:
I'll further note, by implementing an array type check, it would not break existing key value use-cases:

```php
$this->logger->info("Hey there! I'm just testing!", ["className" => __CLASS__]);
```
> [2019-01-02 02:06:12] INFO: [{"className":"LFLAPI\\Controllers\\Ecom\\PromoController"}] Hey there! I'm just testing!

It will also not break things out of the box (ie: not using a `pushProcessor()` as I've done).

```php
$this->logger->info("Hey there! I'm just testing!", [__CLASS__]);
```

> [2019-01-05 17:11:28] INFO: [["LFLAPI\\Controllers\\Ecom\\PromoController"]] Hey there! I'm just testing!

Moreover, you voiced concerns in https://github.com/Seldaek/monolog/issues/675 over adhering to PSR-3 standards. I argue, that this pull request still adhere's to the recommend useage as described in PSR-3. In fact, I believe it stands truer to what PSR-3 states about Context implementation ([section 1.3)](https://www.php-fig.org/psr/psr-3/#13-context):

> The array can contain anything. Implementors MUST ensure they treat context data with as much lenience as possible.

Seldaek:
Take a look at https://github.com/Seldaek/monolog/blob/master/src/Monolog/Processor/IntrospectionProcessor.php instead.. this kinda does what you want and will leave your code even cleaner.

PR #1273, opened 2019-01-02
Add SentryHandler for sentry/sentry >= 2.0

HazAT:
Adds `SentryHandler` class to support new unified Sentry SDK API.
`SentryHandler` should be used for `sentry/sentry >= 2.0`, we still want to keep the `RavenHandler` for versions `< 2.0`.


HazAT:
I just wanted to ask if there is anything I can do so this will be merged.
This is blocking a new major release for our laravel SDK
https://github.com/getsentry/sentry-laravel/

maxailloud:
We could really use this PR.

src/Monolog/Handler/SentryHandler.php
```diff
@@ -0,0 +1,261 @@
...
+     * @var FormatterInterface The formatter to use for the logs generated via handleBatch()
+     */
+    protected $batchFormatter;
+
+    /**
+     * @param Hub $hub
+     * @param int $level The minimum logging level at which this handler will be triggered
+     * @param bool $bubble Whether the messages that are handled can bubble up the stack or not
+     */
+    public function __construct(Hub $hub, $level = Logger::DEBUG, bool $bubble = true)
```
idr0id: I think the first argument type should be `HubInterface`.

src/Monolog/Handler/SentryHandler.php
```diff
@@ -0,0 +1,261 @@
...
+                return Severity::warning();
+            case Logger::ERROR:
+                return Severity::error();
+            case Logger::CRITICAL:
+                return Severity::fatal();
+            case Logger::ALERT:
+                return Severity::fatal();
+            case Logger::EMERGENCY:
+                return Severity::fatal();
+
```
idr0id: Maybe needs `default` case to handle unsupported log level?

src/Monolog/Handler/SentryHandler.php
HazAT: Good catch!

fmasa:
Would you mind implementing breadcrumbs in separate handler [as was the case in sentry/sentry](https://github.com/getsentry/sentry-php/blob/1.10.0/lib/Raven/Breadcrumbs/MonologHandler.php)?

Right now users would have to write their own handler to send log records as breadcrumbs.

Also I think you have to write tests for this handler (I can help).

stof:
I suggest maintaining that as a separate composer package (which can then define its requirement properly) instead of adding it in Monolog itself, where requirements won't be checked by composer and where it will be unmaintained (many of the handlers added in core are actually very badly maintained, as the people contributing them initially don't help maintaining them, and the maintainer of Monolog does not use them).

HazAT:
We will move this handler into our `sentry/sentry` package to have more control over it.
Thanks all for contributing.

PR #1280, opened 2019-01-18
Remove SlackbotHandler from the docs

lightswitch05:
`SlackbotHandler` was deleted by https://github.com/Seldaek/monolog/commit/655b4284d80fcfa71f4d5f9c2c972b72be8b81b4 - also remove it from the docs


lightswitch05:
Would anyone be willing to give insights between the reliability of `SlackWebhookHandler` vs. `SlackHandler`? Just looking at implementations, it seems to me that `SlackHandler` is a bit more complicated, and therefore might be a little less reliable then `SlackWebhookHandler`. Having to read from the socket for the messages to go through definitely appears to be a hack.

Seldaek:
Sorry merged #1328 before I saw this but it does the same so closing here. Thanks for the PR tho.

PR #1281, opened 2019-01-25
Fix create duplicate records in *GroupHandler 1.x

GreYY:
Fix create duplicate records in *GroupHandler when there is more than one Processor.


Seldaek:
Oops, thanks :)

PR #1282, opened 2019-01-25
Fix create duplicate records in *GroupHandler

GreYY:
Fix create duplicate records in *GroupHandler when there is more than one Processor.


Seldaek:
Closing as #1281 is going to be merged into master

PR #1284, opened 2019-01-26
optimize MemoryUsage processors

TheDigitalOrchard:
Determining accurate memory usage should be an optimized process, not introducing any extra overhead. This pull request uses just a single variable and avoids a function call, if it's not needed.

1. Combine `$bytes` and `$formatted` into single `$usage` variable.
2. Check to see if formatting is needed before calling `formatBytes()`.


Seldaek:
Thanks

PR #1287, opened 2019-02-04
Check file was opened successfully before trying to read/write

### LINKED ISSUE #1286: DeduplicationHandler causes infinite loop and fills logs with warnings
Our server ran out of diskspace after a cronjob kept writing to a log file in an infinite loop caused by the DeduplicationHandler.

The infinite loop happens here:
https://github.com/Seldaek/monolog/blob/bfc9ebb28f97e7a24c45bdc3f0ff482e47bb0266/src/Monolog/Handler/DeduplicationHandler.php#L146-L151

And the output is these two lines infinitely:
```
PHP Warning:  feof() expects parameter 1 to be resource, boolean given in .../vendor/monolog/monolog/src/Monolog/Handler/DeduplicationHandler.php on line 146
PHP Warning:  fgets() expects parameter 1 to be resource, boolean given in .../vendor/monolog/monolog/src/Monolog/Handler/DeduplicationHandler.php on line 147
```

On further inspection I found out the cause to be insufficient file permissions, there was also a warning:
```
PHP Warning:  fopen(...): failed to open stream: Permission denied in .../vendor/monolog/monolog/src/Monolog/Handler/DeduplicationHandler.php on line 140
```

But anyway, should probably check that `$handle` is not `false` before going into the while loop.

cvuorinen:
Fixes #1286

### DIFF
--- a/src/Monolog/Handler/DeduplicationHandler.php
+++ b/src/Monolog/Handler/DeduplicationHandler.php
@@ -137,6 +137,11 @@ class DeduplicationHandler extends BufferHandler
         }
 
         $handle = fopen($this->deduplicationStore, 'rw+');
+
+        if (!$handle) {
+            throw new \RuntimeException('Failed to open file for reading and writing: ' . $this->deduplicationStore);
+        }
+
         flock($handle, LOCK_EX);
         $validLogs = [];
 


cvuorinen:
Tried to write a unit test for this, but couldn't even get into `collectLogs` method since this call:
https://github.com/Seldaek/monolog/blob/ebb804e432e8fe0fe96828f30d89c45581d36d07/src/Monolog/Handler/DeduplicationHandler.php#L166

causes infinite recursion and stack overflow here:
https://github.com/Seldaek/monolog/blob/ebb804e432e8fe0fe96828f30d89c45581d36d07/src/Monolog/ErrorHandler.php#L201

Seldaek:
Thanks makes sense

PR #1288, opened 2019-02-04
fixing MongoDBFormatter for new MongoDB driver

ravimisra:
For new MongoDB drivers, added a new formatter viz MongoDBFormatterForNewDriver which able to handle UTCDateTime.
to use this new formatter:
`$formatter = new MongoDBFormatterForNewDriver(0);`
instead of using:
`$formatter = new MongoDBFormatter(0);`


Seldaek:
For the new driver you should rather use Monolog 2.0-beta1 or above which supports the latest MongoDB extension.

PR #1290, opened 2019-02-14
use mb_ functions everywhere if possible

versh23:
when log a big cyrrillic message we neede use a mb_ functions.

and there https://github.com/bzikarsky/gelf-php/blob/master/src/Gelf/Transport/HttpTransport.php#L193

$rawMessage  return false, because string is corrupted

### DIFF
--- a/src/Monolog/Formatter/FlowdockFormatter.php
+++ b/src/Monolog/Formatter/FlowdockFormatter.php
@@ -82,7 +82,7 @@ class FlowdockFormatter implements FormatterInterface
         return $formatted;
     }
 
-    public function getShortMessage($message)
+    public function getShortMessage(string $message): string
     {
         static $hasMbString;
 
--- a/src/Monolog/Formatter/FlowdockFormatter.php
+++ b/src/Monolog/Formatter/FlowdockFormatter.php
@@ -11,6 +11,8 @@
 
 namespace Monolog\Formatter;
 
+use Monolog\Utils;
+
 /**
  * formats the record to be used in the FlowdockHandler
  *
@@ -84,22 +86,10 @@ class FlowdockFormatter implements FormatterInterface
 
     public function getShortMessage(string $message): string
     {
-        static $hasMbString;
-
-        if (null === $hasMbString) {
-            $hasMbString = function_exists('mb_strlen');
-        }
-
         $maxLength = 45;
 
-        if ($hasMbString) {
-            if (mb_strlen($message, 'UTF-8') > $maxLength) {
-                $message = mb_substr($message, 0, $maxLength - 4, 'UTF-8') . ' ...';
-            }
-        } else {
-            if (strlen($message) > $maxLength) {
-                $message = substr($message, 0, $maxLength - 4) . ' ...';
-            }
+        if (Utils::strlen($message) > $maxLength) {
+            $message = Utils::substr($message, 0, $maxLength - 4) . ' ...';
         }
 
         return $message;
--- a/src/Monolog/Formatter/FlowdockFormatter.php
+++ b/src/Monolog/Formatter/FlowdockFormatter.php
@@ -11,8 +11,6 @@
 
 namespace Monolog\Formatter;
 
-use Monolog\Utils;
-
 /**
  * formats the record to be used in the FlowdockHandler
  *
@@ -84,12 +82,24 @@ class FlowdockFormatter implements FormatterInterface
         return $formatted;
     }
 
-    public function getShortMessage(string $message): string
+    public function getShortMessage($message)
     {
+        static $hasMbString;
+
+        if (null === $hasMbString) {
+            $hasMbString = function_exists('mb_strlen');
+        }
+
         $maxLength = 45;
 
-        if (Utils::strlen($message) > $maxLength) {
-            $message = Utils::substr($message, 0, $maxLength - 4) . ' ...';
+        if ($hasMbString) {
+            if (mb_strlen($message, 'UTF-8') > $maxLength) {
+                $message = mb_substr($message, 0, $maxLength - 4, 'UTF-8') . ' ...';
+            }
+        } else {
+            if (strlen($message) > $maxLength) {
+                $message = substr($message, 0, $maxLength - 4) . ' ...';
+            }
         }
 
         return $message;
--- a/src/Monolog/Formatter/WildfireFormatter.php
+++ b/src/Monolog/Formatter/WildfireFormatter.php
@@ -12,6 +12,7 @@
 namespace Monolog\Formatter;
 
 use Monolog\Logger;
+use Monolog\Utils;
 
 /**
  * Serializes a log message according to Wildfire's header requirements
@@ -90,7 +91,7 @@ class WildfireFormatter extends NormalizerFormatter
         // The message itself is a serialization of the above JSON object + it's length
         return sprintf(
             '%d|%s|',
-            strlen($json),
+            Utils::strlen($json),
             $json
         );
     }
--- a/src/Monolog/Formatter/WildfireFormatter.php
+++ b/src/Monolog/Formatter/WildfireFormatter.php
@@ -12,7 +12,6 @@
 namespace Monolog\Formatter;
 
 use Monolog\Logger;
-use Monolog\Utils;
 
 /**
  * Serializes a log message according to Wildfire's header requirements
@@ -91,7 +90,7 @@ class WildfireFormatter extends NormalizerFormatter
         // The message itself is a serialization of the above JSON object + it's length
         return sprintf(
             '%d|%s|',
-            Utils::strlen($json),
+            strlen($json),
             $json
         );
     }
--- a/src/Monolog/Handler/BrowserConsoleHandler.php
+++ b/src/Monolog/Handler/BrowserConsoleHandler.php
@@ -176,7 +176,7 @@ class BrowserConsoleHandler extends AbstractProcessingHandler
             $args[] = '"font-weight: normal"';
 
             $pos = $match[0][1];
-            $format = Utils::substr($format, 0, $pos) . '%c' . $match[1][0] . '%c' . Utils::substr($format, $pos + Utils::strlen($match[0][0]));
+            $format = Utils::substr($format, 0, $pos) . '%c' . $match[1][0] . '%c' . Utils::substr($format, $pos + strlen($match[0][0]));
         }
 
         array_unshift($args, static::quote($format));
--- a/src/Monolog/Handler/BrowserConsoleHandler.php
+++ b/src/Monolog/Handler/BrowserConsoleHandler.php
@@ -13,6 +13,7 @@ namespace Monolog\Handler;
 
 use Monolog\Formatter\LineFormatter;
 use Monolog\Formatter\FormatterInterface;
+use Monolog\Utils;
 
 /**
  * Handler sending logs to browser's javascript console with no browser extension required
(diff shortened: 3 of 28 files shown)


Seldaek:
Sounds good overall to handle non-ascii better, but IMO this should be handled across all handlers where substr/strlen are done on user strings, and ideally this would use an internal wrapper so that if mbstring extension is missing we fallback to substr/strlen.

versh23:
thanks for reply
I will work on it

versh23:
@Seldaek is it enough ?

src/Monolog/Formatter/WildfireFormatter.php
```diff
@@ -90,7 +91,7 @@ public function format(array $record): string
         // The message itself is a serialization of the above JSON object + it's length
         return sprintf(
             '%d|%s|',
-            strlen($json),
+            Utils::strlen($json),
```
Seldaek: I'd guess this needs to be a real strlen at binary level and not counting unicode characters, but not 100% sure

src/Monolog/Handler/BrowserConsoleHandler.php
```diff
@@ -175,7 +176,7 @@ private static function handleStyles(string $formatted): array
             $args[] = '"font-weight: normal"';
 
             $pos = $match[0][1];
-            $format = substr($format, 0, $pos) . '%c' . $match[1][0] . '%c' . substr($format, $pos + strlen($match[0][0]));
+            $format = Utils::substr($format, 0, $pos) . '%c' . $match[1][0] . '%c' . Utils::substr($format, $pos + Utils::strlen($match[0][0]));
```
Seldaek: These should be undone as the $pos offset is in bytes and not unicode characters

src/Monolog/Handler/ChromePHPHandler.php
```diff
@@ -146,7 +147,7 @@ protected function send(): void
 
         $json = @json_encode(self::$json);
         $data = base64_encode(utf8_encode($json));
-        if (strlen($data) > 240 * 1024) {
+        if (Utils::strlen($data) > 240 * 1024) {
```
Seldaek: Also probably needs to be removed

src/Monolog/Handler/CubeHandler.php
```diff
@@ -134,7 +135,7 @@ private function writeUdp(string $data): void
             $this->connectUdp();
         }
 
-        socket_send($this->udpConnection, $data, strlen($data), 0);
+        socket_send($this->udpConnection, $data, Utils::strlen($data), 0);
```
Seldaek: The two in this file should also be undone as those are byte counts.

src/Monolog/Handler/FleepHookHandler.php
```diff
@@ -96,7 +97,7 @@ private function buildHeader(string $content): string
         $header = "POST " . static::FLEEP_HOOK_URI . $this->token . " HTTP/1.1\r\n";
         $header .= "Host: " . static::FLEEP_HOST . "\r\n";
         $header .= "Content-Type: application/x-www-form-urlencoded\r\n";
-        $header .= "Content-Length: " . strlen($content) . "\r\n";
+        $header .= "Content-Length: " . Utils::strlen($content) . "\r\n";
```
Seldaek: Also should be undone

src/Monolog/Handler/FlowdockHandler.php
```diff
@@ -107,7 +108,7 @@ private function buildHeader(string $content): string
         $header = "POST /v1/messages/team_inbox/" . $this->apiToken . " HTTP/1.1\r\n";
         $header .= "Host: api.flowdock.com\r\n";
         $header .= "Content-Type: application/json\r\n";
-        $header .= "Content-Length: " . strlen($content) . "\r\n";
+        $header .= "Content-Length: " . Utils::strlen($content) . "\r\n";
```
Seldaek: This also

src/Monolog/Handler/PushoverHandler.php
```diff
@@ -158,7 +159,7 @@ private function buildHeader(string $content): string
         $header = "POST /1/messages.json HTTP/1.1\r\n";
         $header .= "Host: api.pushover.net\r\n";
         $header .= "Content-Type: application/x-www-form-urlencoded\r\n";
-        $header .= "Content-Length: " . strlen($content) . "\r\n";
+        $header .= "Content-Length: " . Utils::strlen($content) . "\r\n";
```
Seldaek: This should be undone but the others above are fine.

src/Monolog/Handler/SlackHandler.php
```diff
@@ -129,7 +130,7 @@ private function buildHeader(string $content): string
         $header = "POST /api/chat.postMessage HTTP/1.1\r\n";
         $header .= "Host: slack.com\r\n";
         $header .= "Content-Type: application/x-www-form-urlencoded\r\n";
-        $header .= "Content-Length: " . strlen($content) . "\r\n";
+        $header .= "Content-Length: " . Utils::strlen($content) . "\r\n";
```
Seldaek: Also should be undone

src/Monolog/Handler/SocketHandler.php
```diff
@@ -325,7 +326,7 @@ private function setStreamChunkSize(): void
 
     private function writeToSocket(string $data): void
     {
-        $length = strlen($data);
+        $length = Utils::strlen($data);
```
Seldaek: Also should be undone

src/Monolog/Handler/SqsHandler.php
```diff
@@ -51,8 +52,8 @@ protected function write(array $record): void
         }
 
         $messageBody = $record['formatted'];
-        if (strlen($messageBody) >= static::MAX_MESSAGE_SIZE) {
-            $messageBody = substr($messageBody, 0, static::HEAD_MESSAGE_SIZE);
+        if (Utils::strlen($messageBody) >= static::MAX_MESSAGE_SIZE) {
+            $messageBody = Utils::substr($messageBody, 0, static::HEAD_MESSAGE_SIZE);
```
Seldaek: Also should be undone

src/Monolog/Handler/StreamHandler.php
```diff
@@ -146,8 +147,8 @@ private function getDirFromStream(string $stream): ?string
             return dirname($stream);
         }
 
-        if ('file://' === substr($stream, 0, 7)) {
-            return dirname(substr($stream, 7));
+        if ('file://' === Utils::substr($stream, 0, 7)) {
+            return dirname(Utils::substr($stream, 7));
```
Seldaek: These don't really matter but also don't really need to be using mbstring

src/Monolog/Handler/SyslogUdp/UdpSocket.php
```diff
@@ -47,13 +49,13 @@ protected function send(string $chunk): void
...
+        socket_sendto($this->socket, $chunk, Utils::strlen($chunk), $flags = 0, $this->ip, $this->port);
     }
 
     protected function assembleMessage(string $line, string $header): string
     {
-        $chunkSize = static::DATAGRAM_MAX_LENGTH - strlen($header);
+        $chunkSize = static::DATAGRAM_MAX_LENGTH - Utils::strlen($header);
 
-        return $header . substr($line, 0, $chunkSize);
+        return $header . Utils::substr($line, 0, $chunkSize);
```
Seldaek: These above all should be undone as they deal with udp bytes

src/Monolog/Utils.php
```diff
@@ -22,4 +22,33 @@ public static function getClass($object): string
...
+        if (null === $hasMbString) {
+            $hasMbString = extension_loaded('mbstring');
+        }
+
+        return $hasMbString;
+    }
+
+    public static function strlen(string $string, ?string $encoding = null): int
+    {
+        if (self::hasMbString()) {
```
Seldaek: I think inlining the `extension_loaded('mbstring')` instead of the function call is more efficient here, as opcache can optimize this call away AFAIK.

src/Monolog/Processor/UidProcessor.php
```diff
@@ -45,11 +46,11 @@ public function getUid(): string
...
     public function reset()
     {
-        $this->uid = $this->generateUid(strlen($this->uid));
+        $this->uid = $this->generateUid(Utils::strlen($this->uid));
     }
 
     private function generateUid(int $length): string
     {
-        return substr(bin2hex(random_bytes((int) ceil($length / 2))), 0, $length);
+        return Utils::substr(bin2hex(random_bytes((int) ceil($length / 2))), 0, $length);
```
Seldaek: This file also should be undone as it doesn't bring anything here

Seldaek:
@versh23 hey, thanks a lot for the work, and sorry if I wasn't clear, this shouldn't always be blindly applied as some cases need to work with raw bytes and some cases work better with unicode characters. If you don't mind reverting those I marked, I think the rest is safe and then it can be merged before 2.0 final comes out.

versh23:
there is https://www.php.net/manual/ru/function.mb-strcut.php function

may be i drop all Utils::strlen() and use mb-strcut instead mb_strlen in Utils::substr ?

It will be resolve discussion about length byte and nonbreakable substring multibyte string.

What do you think ? @Seldaek

Seldaek:
That sounds good for SqsHandler where substr was used to cut a string based on a byte limit, but the other cases IMO should not use mb_string at all.

versh23:
hmm. 
my problem was started in GelfMessageFormatter

we have a restriction on each fields = 32kb

if my error message\ctx\extra more than 32kb i need to truncate field data. 

how to normaly trucate data without mb_string if my log message contains unicode ?

we can truncate by bytes length(mb_strcut) or string length(mb_substr)

Seldaek:
I would say for gelf it should use mb strcut if mbstring is there or substr
if not.

As the limit as far as I know is in bytes, we want to cut in bytes and not
in characters, but if mbstring is missing then you have to cut with substr
and hope for the best..

versh23:
do you think there are problem only for gelf ? i dont think so, because if we subtr a string with unicode without MB_STRING = corrupted string

FlowdockFormatter for example.


so, i can find where use a truncate message string by maxLength and do it with mb_string with BC , its ok ?

Seldaek:
For UdpSocket, the strlen must remain a real strlen, not mb_strlen, as it counts bytes to be sent. The substr could indeed be mb_strcut when available.

For FlowdockFormatter yes mb_strcut when available makes sense too instead of substr.

versh23:
ok, i fix it. thank you for fast feedback!

src/Monolog/Handler/BrowserConsoleHandler.php
versh23: Utils::substr work with bytes count now

src/Monolog/Handler/SqsHandler.php
versh23: Utils::substr work with bytes count now
Utils::strlen - removed

versh23:
Im done

src/Monolog/Formatter/FlowdockFormatter.php
```diff
@@ -82,7 +82,7 @@ public function formatBatch(array $records): array
         return $formatted;
     }
 
-    public function getShortMessage(string $message): string
+    public function getShortMessage($message)
```
Seldaek: This change seems unnecessary?

Seldaek:
Hey, thanks for the update. Apart from that last thing that seems like a mistake, the rest looks good now!

versh23:
done

ogunkarakus:
@versh23 @Seldaek 

Maybe the mb_strlen and mb_substr functions should also be used. I don't think it will cause any performance loss. I think they should be used.

versh23:
we work with bytes, not a string length. 
if ext-mbstring is enable, string truncating by bytes will be with mb_strcut

versh23:
@Seldaek is it ok?

Seldaek:
Yup looking good now thanks!

PR #1291, opened 2019-02-21
Replace logrotate link with alternative page

Crecket:
The current link no longer exists so this is the most extensive and popular page for the logrotate command.


Seldaek:
Thanks

PR #1292, opened 2019-02-24
Check and reject duplicate handlers in the logger.

crayner:
Stop the duplication of log entries in the same log.

Tested:  Locally for a couple of weeks.  

See Issue [#1293](https://github.com/Seldaek/monolog/issues/1293)


Seldaek:
Thanks for the PR but I'd rather not do this. This is due to a misconfiguration and shoving the problem under the rug isn't really a solution, you should rather fix the root cause IMO.

PR #1304, opened 2019-03-22
Add possibility to use RFC3164 for udp syslog

dominikkukacka:


Seldaek:
Sorry but as I am not familiar with RFC3164 can you explain what the upside is of using this obsolete RFC vs RFC5424 (which is 10 years old and not exactly bleeding edge)?

dominikkukacka:
No immediate upsides. 
But there are certain tools which only support RFC3164, e.g. the syslog plugin of logstash (https://www.elastic.co/guide/en/logstash/current/plugins-inputs-syslog.html)

Seldaek:
Ok, well I guess it doesn't hurt to have this so will merge, thanks!

dominikkukacka:
@Seldaek should I implement it on master as well?

Seldaek:
No no it's been merged already

PR #1309, opened 2019-04-03
Check for JsonSerializable in normalizeException

alex2005git:
I needed this change because I have exceptions that I want to handle how they get json serialized.


Seldaek:
Thanks

PR #1311, opened 2019-04-04
Dont use more than 2000 chars for slack attachments field

Nyholm:
See docs here: https://api.slack.com/docs/interactive-message-field-guide

Slack will automatically limit this for us, but it will remove the last 3 back ticks so long messages is not properly formatted.


Nyholm:
Friendly ping

Seldaek:
Thanks

Nyholm:
Wohoo. Thank you for merging

PR #1312, opened 2019-04-05
For when the date changes during a long process.

yasutomog:
Laravel uses an instance of Handler in one process.

When using "RotatingFileHandler", if the date changes in one process, the error "Could not be opened: failed to open stream: No such file or directory in" is occured.

The date is checked with the "write" function of "RotatingFileHandler". If the date has changed, closing the file. However, "$dirCreated" of "StreamHandler" remains True. Therefore, a new directory can not be created and an error occurs.

Fixed "$dirCreated" to its initial value when closing a file.

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -65,6 +65,7 @@ class StreamHandler extends AbstractProcessingHandler
             fclose($this->stream);
         }
         $this->stream = null;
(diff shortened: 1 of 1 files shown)


Seldaek:
Merged in 1.x/master as https://github.com/Seldaek/monolog/commit/ab7e54e7df2a798bec317734a332642e88f9127a - thanks!

PR #1316, opened 2019-04-17
fix ProcessorInterface

keksa:
Every single processor that implements `ProcessorInterface` has only single record as argument. This is true even when I test the code in my project.

Having the ProcessorInterface argument for multiple records is seriously confusing and untrue.

### DIFF
--- a/src/Monolog/Processor/ProcessorInterface.php
+++ b/src/Monolog/Processor/ProcessorInterface.php
@@ -19,7 +19,7 @@ namespace Monolog\Processor;
 interface ProcessorInterface
 {
     /**
-     * @return array The processed records
+     * @return array The processed record
      */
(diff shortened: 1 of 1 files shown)


src/Monolog/Processor/ProcessorInterface.php
```diff
@@ -19,7 +19,7 @@
 interface ProcessorInterface
 {
     /**
-     * @return array The processed records
+     * @return array The processed record
      */
-    public function __invoke(array $records);
+    public function __invoke(array $record);
```
keksa: Ideally I would also add a return type, but that could be considered BC break. Let me know if you want me to add it.
```php
public function __invoke(array $record): array;
```

Seldaek:
Oh nice catch thanks

PR #1318, opened 2019-04-22
Adding curl sharing to fix #1313

### LINKED ISSUE #1313: Share CURL connection in LogglyHandler
The LogglyHandler seems to be very slow, one reason could be, that every log entry opens a new curl connection. An easy optimization would be to share the curl connection for every requests to counter i.e. a tcp slowstart (https://www.maxcdn.com/one/visual-glossary/tcp-slow-start/?fbclid=IwAR2aDgazXaH1RQPwlPuA9nQKg4ZVysGIBpBomphI7_s9fNvTba-XHbCjlKE), instead of creating a new connection for every request.

(An example: http://technosophos.com/2012/06/18/connection-sharing-curl-php-how-re-use-http-connections-knock-70-rest-network-time.html)

b3nl:
This change caches the curl handlers for every endoint and reuses them. This makes the logglyHandler two to three times faster.

### DIFF
--- a/src/Monolog/Handler/LogglyHandler.php
+++ b/src/Monolog/Handler/LogglyHandler.php
@@ -14,6 +14,7 @@ namespace Monolog\Handler;
 use Monolog\Logger;
 use Monolog\Formatter\FormatterInterface;
 use Monolog\Formatter\LogglyFormatter;
+use function array_key_exists;
 
 /**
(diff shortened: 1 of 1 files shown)


Seldaek:
Makes sense for handlers that are making lots of requests, thanks.

PR #1320, opened 2019-04-23
Documentation - Missing word "not"

patrickkusebauch:
There is a word missing that changes the meaning of the sentence.


Seldaek:
Thanks but it is actually not missing. `logrorate` should be used in place of this handler when you care about running things reliably.

PR #1321, opened 2019-04-23
Fixed recursive create directory.

Egishe:
Setting right permissions for each directory.


stof:
what is this fixing ?

Egishe:
> what is this fixing ?
`mkdir($dir, 0777, true);` it don't setting permissions 777 for each directory but my fixing do it

Seldaek:
I don't see what is wrong with mkdir(), this works fine in my tests.

PR #1327, opened 2019-05-28
Fixed gelf formatter

nikitaksv:
If the first element in the array has exceeded the maximum message length, then the following elements are ignored.


Seldaek:
Thanks

PR #1334, opened 2019-06-21
Added ElasticsearchLogstashHandler

lyrixx:
ATM, there are few options to push log to Elastic Stack in order to play with them:

* install logstash and use a Gelf handler. It works but you have to install logstash and configure it. Not an easy task. More over, it need an extra PHP package
* use the ES handler: It does not play well with context and extra: Kibana is not able to filter on nested object. And this handler is tightly coupled to the ElasticaFormatter formater. More over, it need an extra PHP package
* use something to parse file logs. This is really a bad idea since it involves a parsing... More over a daemon is needed to do that (file beat / logstash / you name it)

This is why I'm introducing a new Handler.

* There is not need to install anything (expect ES, of course)
* It play very well with Kibana, as it uses the Logstash format
* It requires symfony/http-client, but in a modern PHP application (SF 4.3) this dependency is already present 
* It slow down a bit the application since it trigger an HTTP request for each logs. But symfony/http-client is non-blocking. If you want to use it in production, I recommend to wrap this handler in a buffer handler or a cross-finger handle to have only one HTTP call.

**Note: This handler is not aimed to be used in production**

---

This is what you can expect **out of the box**
image

### DIFF
--- a/src/Monolog/Formatter/ElasticsearchLogstashHandler.php
+++ b/src/Monolog/Formatter/ElasticsearchLogstashHandler.php
@@ -1,0 +1,77 @@
+<?php
+
+namespace AppBundle\Monitoring\Monolog\Handler;
+
+use Monolog\Formatter\FormatterInterface;
+use Monolog\Formatter\LogstashFormatter;
+use Monolog\Handler\AbstractProcessingHandler;
+use Monolog\Logger;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\HttpClientInterface;
+
+/**
+ * Push logs directly to Elasticsearch and format them according to Logstash specification
+ *
+ * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ */
+class ElasticsearchLogstashHandler extends AbstractProcessingHandler
+{
+    private $endpoint;
+    private $index;
+    private $client;
+    private $responses;
+
+    public function __construct($endpoint = 'http://elasticsearch:9200', $index = 'monolog', HttpClientInterface $client = null, $level = Logger::DEBUG, $bubble = true)
+    {
+        if (!class_exists(HttpClientInterface::class)) {
+            throw new \LogicException(sprintf('%s handler needs symfony/http-client, please run `composer require symfony/http-client`.', __CLASS__));
+        }
+
+        parent::__construct($level, $bubble);
+        $this->endpoint = $endpoint;
+        $this->index = $index;
+        $this->client = $client ?: HttpClient::create();
+        $this->responses = new \SplObjectStorage();
+    }
+
+    protected function write(array $record)
+    {
+        $this->sendToElasticsearch(array($record));
+    }
+
+    protected function getDefaultFormatter()
+    {
+        return new LogstashFormatter('application', null, null, 'ctxt_', LogstashFormatter::V1);
+    }
+
+    private function sendToElasticsearch(array $documents)
+    {
+        $body = '';
+        foreach ($documents as $document) {
+            $body .= json_encode([
+                'index' => [
+                    '_index' => $this->index,
+                    '_type' => '_doc',
+                ],
+            ]);
+            $body .= "\n";
+            $body .= $document['formatted'];
+            $body .= "\n";
+        }
+
+        $response = $this->client->request('POST', $this->endpoint.'/_bulk', [
+            'body' => $body,
+            'headers' => [
+                'Content-Type' => 'application/json',
+            ],
+        ]);
+
+        $this->responses->attach($response);
+
+        foreach ($this->client->stream($this->responses, 0) as $response => $chunk) {
+            if (!$chunk->isTimeout() && $chunk->isFirst()) {
+                $this->responses->detach($response);
+            }
+        }
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 1.25.0 (xxxx-xx-xx)
+
+  * Added ElasticsearchLogstashHandler
+
 ### 1.24.0 (2018-11-05)
 
   * BC Notice: If you are extending any of the Monolog's Formatters' `normalize` method, make sure you add the new `$depth = 0` argument to your function signature to avoid strict PHP warnings.


src/Monolog/Formatter/ElasticsearchLogstashHandler.php
```diff
@@ -0,0 +1,75 @@
...
+    public function handleBatch(array $records)
+    {
+        $this->sendToElasticsearch($records);
+    }
+
+    private function sendToElasticsearch(array $documents)
+    {
+        $body = '';
+        foreach ($documents as $document) {
+            $body .= '{"index":{"_index":"'.$this->index.'","_type":"_doc"}}'."\n";
```
stof: shouldn't this use `json_encode`  to produce the JSON, to ensure proper escaping ?

src/Monolog/Formatter/ElasticsearchLogstashHandler.php
```diff
@@ -0,0 +1,75 @@
...
+        $response = $this->client->request('POST', $this->endpoint.'/_bulk', [
+            'body' => $body,
+            'headers' => [
+                'Content-Type' => 'application/json',
+            ],
+        ]);
+
+        $this->responses->attach($response);
+
+        foreach ($this->client->stream($this->responses, 0) as $response => $chunk) {
```
stof: this will block until we get some data from the response, while `sendToElasticsearch` is called. So this defeats the non-blocking nature of HttpClient if you expect that it will lead to a non-blocking `log` call.

lyrixx: Ah! @nicolas-grekas told me to do that :( Maybe he does not know this component! :joy: 

What I want to achieve:
* be non-blocking
* but Response::desctruct() will block, so I need to keep all response "somewhere"
* from time to time, clean "somewhere"

In my first implementation, I used a fixed array. But this is a bit stupid.
What do you suggest to fix this code ?

(Note: according to  [blackfire](https://blackfire.io/profiles/96195738-7c64-4d90-a2e7-6ac3521f9067/graph), it seems to work...)

stof: you spend 23% of the time in this handler, most of it being actually due to calling the HttpClient component. That's not non-blocking.

Should we instead perform this cleaning in `flush` and `close` ?

lyrixx: > Should we instead perform this cleaning in `flush` and `close` ?

No because In a CLI context, this may happen not so much.

stof: hmm, actually, the fact of passing `0` as timeout in the second argument indeed means that we will process only the chunks which have been received. So this is fine. I missed it initially (I thought about it while reviewing https://github.com/symfony/symfony/pull/32174).

Seldaek:
Thanks @lyrixx, I am medium happy about this PR tbh, for a few reasons:

- symfony/http-client dependency makes sense if you have symfony, but monolog is used in many contexts/frameworks that don't..
- if at all I think this belongs more in monolog 2.x
- and well lastly I think it probably makes more sense to publish in an extra package which can have the explicit dependency on symfony/http-client then.

src/Monolog/Formatter/ElasticsearchLogstashHandler.php
```diff
@@ -0,0 +1,77 @@
...
+    public function __construct($endpoint = 'http://elasticsearch:9200', $index = 'monolog', HttpClientInterface $client = null, $level = Logger::DEBUG, $bubble = true)
+    {
+        if (!class_exists(HttpClientInterface::class)) {
+            throw new \LogicException(sprintf('%s handler needs symfony/http-client, please run `composer require symfony/http-client`.', __CLASS__));
+        }
+
+        parent::__construct($level, $bubble);
+        $this->endpoint = $endpoint;
+        $this->index = $index;
+        $this->client = $client ?: HttpClient::create();
```
nicolas-grekas: If we want to remove the dependency to the component, we can remove this call and make the `$client` argument mandatory. That will only require the http-client-contracts, which is an abstract dependency.

nicolas-grekas:
> symfony/http-client dependency makes sense if you have symfony

what does this mean? it's a standalone component.

> have the explicit dependency on symfony/http-client then.

if this is problematic, I'd suggest relying only on symfony/http-client-contracts, which is an abstraction

src/Monolog/Formatter/ElasticsearchLogstashHandler.php
nicolas-grekas: we can also keep the argument optional, and throw an exception when the component is missing but not client has been provided, as done in https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/HttpClient/Psr18Client.php#L65

lyrixx:
HttpClient is not an hard dependency. This is the same as for the ElasticHandler or the LogstashHandler.
I don't understand why this should be different for this handler.

Seldaek:
The point is those other handlers were added ages ago, and removing them here would cause breakage for not much benefit, but for new handlers with third-party dependencies IMO the right way is to publish as a third-party package, with requires on monolog and on whichever dependency is needed. It lets Composer resolve everything which makes more sense really.

@nicolas-grekas requiring the contracts package doesn't really solve the above, so IMO still best to just release as a separate package. That's why there is https://github.com/Seldaek/monolog/wiki/Third-Party-Packages - we can still add support for that third party package in the monolog-bundle for example.

nicolas-grekas:
This means another maintainers need to deal with these per-handle packages. It may be easier to ask ppl to send PRs here: managing a package it very different than patching a class. I'd be more confident if you (monolog maintenance team) would deal with the releasing part.
Unless the old handlers create a significant maintenance burden, please consider this aspect.

Seldaek:
These handlers typically don't change much, so the package management aspect is very little.. We could do a monolog organization of some sort, but I don't really want to be managing a million handler packages for stuff that I don't know and don't care about, so IMO distributing this and having everyone maintaining their own handlers they are interested in makes more sense. I can see you may not agree with this seeing how the Symfony project handles this very differently, but that's life :)

Anyway this handler in particular being dev-time only which really isn't the main use case for monolog.. If you feel it makes sense for symfony projects it could go in the monolog bridge or the monolog bundle there I guess.

Seldaek:
P.S; I can see how it'd be good to have a way to write to kibana directly tho, if you can make this support handleBatch and skip the http-client dependency then it might make more sense to me to have a production ready thing.

lyrixx:
> These handlers typically don't change much, so the package management aspect is very little.. We could do a monolog organization of some sort, but I don't really want to be managing a million handler packages for stuff that I don't know and don't care about, so IMO distributing this and having everyone maintaining their own handlers they are interested in makes more sense. I can see you may not agree with this seeing how the Symfony project handles this very differently, but that's life :)

Yes, I already proposed a "Contrib" Repository. 
Back in the past, we did a `Contrib` namespace for Sismo https://github.com/FriendsOfPHP/Sismo/tree/master/src/Sismo/Contrib
It worked quite well

> P.S; I can see how it'd be good to have a way to write to kibana directly tho, 

Kibana is only a "reader" on top of ES. I mean you can not write directly to kibana. You push logs to ES, and then you visualize them with Kibana

> if you can make this support handleBatch and skip the http-client dependency then it might make more sense to me to have a production ready thing.

HandleBatch is really easy to do. Just 3/5 LOC to add. But I don't want to implement a new HTTP client here. Finally, http client is not added to the composer.json, it's only a soft deps, so I don't really see the issue.

But I understand having another Handler means another Handler to maintain.

PR #1336, opened 2019-06-25
Add telegram bot handler.

alexandrmazur96:
Add handler which can send logs to Telegram using [Telegram Bot API](https://core.telegram.org/bots/api).

Handler depends only on the cURL extension.

image

P.S. seems that travis-ci failed because of PHP-version ^7.1 in composer.json but [CI use 8.0.0-dev](https://travis-ci.org/Seldaek/monolog/jobs/550408854)

### DIFF
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -80,7 +80,7 @@ class TelegramBotHandler extends AbstractProcessingHandler
             'chat_id' => $this->channel,
         ]));
 
-        $result = curl_exec($ch);
+        $result = Curl\Util::execute($ch);
         $result = json_decode($result, true);
 
         if ($result['ok'] === false) {
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Monolog\Handler;
+
+use RuntimeException;
+use Monolog\Logger;
+
+/**
+ * Handler send logs to Telegram using Telegram Bot API.
+ *
+ * How to use:
+ *  1) Create telegram bot with https://telegram.me/BotFather
+ *  2) Create a telegram channel where logs will be recorded.
+ *  3) Add created bot from step 1 to the created channel from step 2.
+ *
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -0,0 +1,90 @@
...
+        $url = self::BOT_API . $this->apiKey . '/SendMessage';
+        curl_setopt($ch, CURLOPT_URL, $url);
+        curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+        curl_setopt($ch, CURLOPT_SSL_VERIFYPEER, true);
+        curl_setopt($ch, CURLOPT_POSTFIELDS, http_build_query([
+            'text' => $message,
+            'chat_id' => $this->channel,
+        ]));
+
+        $result = curl_exec($ch);
```
Seldaek: Please use `Curl\Util`, see MandrillHandler or SlackWebhookHandler for example usages.

alexandrmazur96: Thank you for your reply! 
I looked at Curl\Util and find that `execute()` does not return any response from the server and only throws an exception when got curl error. 

But, Telegram API send error in the response body and I think would be great if we say to the user what exactly wrong.

Maybe would be better add a method to the `Curl\Util` same as `execute` but with response return from `curl_exec()` call?

Seldaek: I think we could change Curl\Util::execute to always return the result, as it throws otherwise. The return type would then be `string` or `true` depending on CURLOPT_RETURNTRANSFER.

alexandrmazur96: I create a [pull request](https://github.com/Seldaek/monolog/pull/1341) for this feature.

I can fix this after merging it, or I can merge this pull request into this branch (the bad way I guess).

Seldaek: Merged it, please rebase this :)

alexandrmazur96: Rebased and apply changes.
I retest my code - all good ;)

Seldaek:
Thanks

PR #1338, opened 2019-07-02
timestamped index in elastica formatter

hervelaunay:
Like other elasticsearch data provider monolog handler should add timestamp in index definition. It allows to easily purge old data by safely removing matching indexes.

### DIFF
--- a/src/Monolog/Formatter/ElasticaFormatter.php
+++ b/src/Monolog/Formatter/ElasticaFormatter.php
@@ -79,10 +79,14 @@ class ElasticaFormatter extends NormalizerFormatter
      */
     protected function getDocument($record)
     {
+        // index should be timestamped to allow log rotating
+        $date = str_replace('-', '.', explode('T', $record['datetime'])[0]);
+        $index = $this->index . '-' . $date;
+
         $document = new Document();
         $document->setData($record);
         $document->setType($this->type);
-        $document->setIndex($this->index);
+        $document->setIndex($index);
 
         return $document;
     }


Seldaek:
cc @lyrixx any opinion here? I don't know the ES stuff enough to judge if this is a horrible idea or not.. Sounds like changing the index name is a BC break though.

src/Monolog/Formatter/ElasticaFormatter.php
```diff
@@ -79,10 +79,14 @@ public function getType()
      */
     protected function getDocument($record)
     {
+        // index should be timestamped to allow log rotating
+        $date = str_replace('-', '.', explode('T', $record['datetime'])[0]);
```
Seldaek: `$date = $record['datetime']->format('Y.m.d')` would achieve the same in a more obvious way as far as I can tell.

lyrixx:
This is is really opinionated and this will lead to some issue:

1. The elastica formatter is used in combinaison with ElasticHandler. And it push data in a blocking way. I would not recommend it in production. So this change is not really needed since it's not needed to log-rotate in dev env
1. This will create a new index each day. It means people need to use an index template to avoid bad surprise. I'm not sure this feature is well known by everybody
1. This will lead to many many indexes. So many many shards. So a big slow down penalty (more RAM, and more CPU). A good cleaning tooling is required for that

TL;DR: A configuration option is needed for that

Note: I uses to use this strategy, but I knew what I was doing ... putting theses defaults seems harmful without good knowledge of ES

hervelaunay:
I didn't see in the documentation that this handler is dedicated for dev only. Thanks for this precision.
I saw this PR https://github.com/Seldaek/monolog/pull/1334 that works in non blocking way but it is rejected and it does not include timestamp indexes and it is not recommanded for production :-(.
So what is your advise to send logs from a PHP application to elasticsearch in production (micro-service application architecture)?
Thanks for your help.
Do you wish I adapt my PR to make timestamped index optionnal ?

lyrixx:
> I didn't see in the documentation that this handler is dedicated for dev only.

Calling ES each time you got a log is a terrible idea. You can wrap it in a Buffer Handler or a Cross Fringer Handler but you will one extra HTTP call to ES while you are processing a Client Request.

More over, you will add pressure on ES with many documents to process.

People usually prefer to use LogStash (or similar tool) that is able to listen on lower and faster protocol (raw TCP or even better: UDP). Theses tools are also able to buffer many documents before sending them to ES and so reduce the presure on ES.

That's why I discourage people to use this handler in a production environment.

lyrixx:
> Do you wish I adapt my PR to make timestamped index optionnal ?

Yes, please

Seldaek:
I'd suggest maybe closing this and doing the change with an optional parameter only in https://github.com/Seldaek/monolog/pull/1339 for 2.0, this sounds like not really worth adding to 1.x anymore..

PR #1341, opened 2019-07-03
Change Curl\Util for return response from cURL request.

alexandrmazur96:
Simply rework `Curl\Util::execute()` method. So now this method returns HTTP response from the request.

Moreover, because this method returns nothing - these changes do not break anything.


Seldaek:
Thanks

PR #1345, opened 2019-07-09
Update DeduplicationHandler.php

billdanbury:
To resolve a report from Veracode scanning, avoid using md5 for fast-hash. Tests show sha1 is compatible and passes Veracode scans.

https://cwe.mitre.org/data/definitions/327.html

From Veracode:

Attack Vector: !php_standard_ns.md5

Number of Modules Affected: 1

Description: This function uses the !php_standard_ns.md5() function, which uses a hash algorithm that is considered weak. In recent years, researchers have demonstrated ways to breach many uses of previously-thought-safe hash functions such as MD5.

Remediation: Consider using a stronger algorithm in order to prevent attackers from being able to manipulate hash results. If this algorithm is being used to hash passwords, then consider using a strong computationally-hard algorithm such as PBKDF2 or bcrypt instead of a plain hashing algorithm.

### DIFF
--- a/src/Monolog/Handler/DeduplicationHandler.php
+++ b/src/Monolog/Handler/DeduplicationHandler.php
@@ -66,7 +66,7 @@ class DeduplicationHandler extends BufferHandler
     {
         parent::__construct($handler, 0, Logger::DEBUG, $bubble, false);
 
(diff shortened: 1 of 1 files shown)


versh23:
i think it is useless
md5() uses for hash file name, not cryptographic purpose

Seldaek:
Right, there is no gain here in changing this. This is not a security sensitive context.

Seldaek:
And btw sha1 also has had collision attacks reported IIRC so it's not any safer than md5.

Issue #1323, opened 2019-04-26
ZendMonitor TypeError


happydude:
Getting a TypeError in 2.0.0-beta1,

./vendor/monolog/monolog/src/Monolog/Handler/ZendMonitorHandler.php:68
zend_monitor_custom_event() expects parameter 1 to be string, integer given

Documentation at http://files.zend.com/help/Zend-Server/content/zendserverapi/zend_monitor-php_api.htm#function-zend_monitor_custom_event

Based on the zend server documentation
https://github.com/Seldaek/monolog/blob/ebb804e432e8fe0fe96828f30d89c45581d36d07/src/Monolog/Handler/ZendMonitorHandler.php#L68
 should be something like

```
zend_monitor_custom_event('',$message, $formatted, $level);
```

In addition the levelMap is not compatible with zend server monitor, they only support three log levels:
```
var_export(
[
	'ZEND_MONITOR_EVENT_SEVERITY_INFO' => \ZEND_MONITOR_EVENT_SEVERITY_INFO,
	'ZEND_MONITOR_EVENT_SEVERITY_WARNING' => \ZEND_MONITOR_EVENT_SEVERITY_WARNING,
	'ZEND_MONITOR_EVENT_SEVERITY_ERROR' => \ZEND_MONITOR_EVENT_SEVERITY_ERROR
]);
```
outputs
```
array (
  'ZEND_MONITOR_EVENT_SEVERITY_INFO' => -1,
  'ZEND_MONITOR_EVENT_SEVERITY_WARNING' => 0,
  'ZEND_MONITOR_EVENT_SEVERITY_ERROR' => 1,
)
```

Seldaek:
If you have time to send a PR, the handler would ideally adjust its behavior based on the version of ZendServer present, I am not familiar enough with it to say how to detect that.. Please target the 1.x branch for a fix.

happydude:
I did not find a way to detect zend server version, but the newer style should be compatible with zend server 6 (php 5.3) and newer.

I did not commit it, but the tests passed with declare(strict_types=1).

PR #1348, opened 2019-07-12
Compatibility with actively maintained versions of Zend Server

happydude:
Fix for #1323

### DIFF
--- a/src/Monolog/Handler/ZendMonitorHandler.php
+++ b/src/Monolog/Handler/ZendMonitorHandler.php
@@ -17,6 +17,7 @@ use Monolog\Logger;
  * Handler sending logs to Zend Monitor
  *
  * @author  Christian Bergau <cbergau86@gmail.com>
+ * @author  Jason Davis <happydude@jasondavis.net>
  */
 class ZendMonitorHandler extends AbstractProcessingHandler
 {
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

src/Monolog/Handler/ZendMonitorHandler.php
```diff
@@ -57,22 +62,23 @@ public function __construct($level = Logger::DEBUG, $bubble = true)
...
-     * @param int    $level
-     * @param string $message
-     * @param array  $formatted
+     * Write to Zend Monitor Events
+     * @param string $type Text displayed in "Class Name (custom)" field
+     * @param string $message Text displayed in "Error String"
+     * @param mixed $formatted Displayed in Custom Variables tab
+     * @param int $severity Set the event severity level (-1,0,1)
      */
-    protected function writeZendMonitorCustomEvent($level, $message, $formatted)
```
gmponos: :( this function is protected..  adding an extra argument might be a BC for some devs.

Seldaek: Yes.. but it also appears like this code never worked, so I am thinking probably nobody uses this or if they do it's breaking and they didn't notice. I doubt extending code is something to worry about.

Issue #1350, opened 2019-07-18
SyslogUdpHandler doesn't respect TimeZone


Pierre-Lannoy:
Hello all contributors, and thanks for this wonderful tool!

While trying 2.0.0-beta2 I found SyslogUdpHandler doesn't respect the timezone set in Logger.
I think this line is not fully compliant with timezone ;)
`
protected function getDateTime(): string
    {
        return date($this->dateFormats[$this->rfc]);
    }
`
Do you know how to proceed to fix it? Should I contact the author? Should I submit a PR (and if so, what's your input on how to do it in regards with the fact that TZ is now set to Logger class)?

Thanks.

yi-yang-github:
This fix actually doesn't work when the datetime is `DateTimeImmutable`, I'll create a new ticket.

PR #1352, opened 2019-07-22
Fix for SyslogUdpHandler timezone issue

Pierre-Lannoy:
Hello!
As described in [this issue](https://github.com/Seldaek/monolog/issues/1350), SyslogUdpHandler doesn't respect `Logger` timezone setting for IETF (RFC-5424) syslog format.
This PR is a fix for this issue.

### DIFF
--- a/src/Monolog/Handler/SyslogUdpHandler.php
+++ b/src/Monolog/Handler/SyslogUdpHandler.php
@@ -11,6 +11,7 @@
 
 namespace Monolog\Handler;
 
+use Monolog\DateTimeImmutable;
 use Monolog\Logger;
 use Monolog\Handler\SyslogUdp\UdpSocket;
 
@@ -57,7 +58,7 @@ class SyslogUdpHandler extends AbstractSyslogHandler
     {
         $lines = $this->splitMessageIntoLines($record['formatted']);
 
-        $header = $this->makeCommonSyslogHeader($this->logLevels[$record['level']]);
+        $header = $this->makeCommonSyslogHeader($this->logLevels[$record['level']], $record['datetime']);
 
         foreach ($lines as $line) {
             $this->socket->write($line, $header);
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/SyslogUdpHandler.php
```diff
@@ -93,10 +94,13 @@ protected function makeCommonSyslogHeader(int $severity): string
             $hostname = '-';
         }
 
-        $date = $this->getDateTime();
-
         if ($this->rfc === self::RFC3164) {
-            return "<$priority>" .
+	    	$datetime->setTimezone(new \DateTimeZone('UTC'));
```
Seldaek: The indent is off here if you can use spaces that'd be great.

src/Monolog/Handler/SyslogUdpHandler.php
```diff
@@ -93,10 +94,13 @@ protected function makeCommonSyslogHeader(int $severity): string
             $hostname = '-';
         }
 
-        $date = $this->getDateTime();
```
Seldaek: As you removed that function usage entirely I would say let's remove the function itself.

src/Monolog/Handler/SyslogUdpHandler.php
```diff
@@ -81,7 +82,7 @@ private function splitMessageIntoLines($message): array
     /**
      * Make common syslog header (see rfc5424 or rfc3164)
      */
-    protected function makeCommonSyslogHeader(int $severity): string
+    protected function makeCommonSyslogHeader(int $severity, DateTimeImmutable $datetime): string
```
Seldaek: This should be DateTimeInterface

Seldaek:
Thanks for the fix, if you can update according to the few comments I'm happy to get this merged in for 2.0 stable :)

src/Monolog/Handler/SyslogUdpHandler.php
Pierre-Lannoy: Hello! Thanks for the hint. I think it's fixed.

src/Monolog/Handler/SyslogUdpHandler.php
Pierre-Lannoy: Haha. Yes, of course... Done.

src/Monolog/Handler/SyslogUdpHandler.php
Pierre-Lannoy: Done. Sorry for the mistake.

Seldaek:
Thanks, tests still needed fixing but did that

Pierre-Lannoy:
Thanks a lot!

PR #1355, opened 2019-07-31
Fix error_handler return type declaration

ScullWM:
Hi,

Small patch to fix error handler function return type.
From php.net documentation, the error_handler should return a bool and not a void (https://www.php.net/manual/en/function.set-error-handler.php).


Seldaek:
Thanks

PR #1357, opened 2019-08-05
Support json scalar formatter

tadasauciunas:
There are cases when logstash doesn't support non-scalar values and json format is necessary as well. This formatter solves this case, as it makes all arrays into scalar values, and encodes the whole record to json afterwards.
I guess this wouldn't be necessary if we could push multiple formatters (i.e. JsonFormatter&ScalarFormatter) to the logger, but that's not possible yet.

### DIFF
--- a/src/Monolog/Formatter/JsonScalarFormatter.php
+++ b/src/Monolog/Formatter/JsonScalarFormatter.php
@@ -1,4 +1,4 @@
-<?php
+<?php declare(strict_types=1);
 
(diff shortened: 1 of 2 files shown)


Seldaek:
Sorry but this isn't quite clear to me why we need this in core, it seems very use case specific and maybe best only present in your project?

tadasauciunas:
Yeah I thought it might be useful but it really is quite case specific.

PR #1359, opened 2019-08-12
Upgrade PHPUnit to v8

MarioBlazek:
This PR upgrades PHPUnit to v8 and fixes deprecations along the way.

### DIFF
--- a/tests/Monolog/Formatter/GelfMessageFormatterTest.php
+++ b/tests/Monolog/Formatter/GelfMessageFormatterTest.php
@@ -86,13 +86,13 @@ class GelfMessageFormatterTest extends TestCase
      */
     public function testFormatInvalidFails()
     {
-        $this->expectException(\InvalidArgumentException::class);
-
         $formatter = new GelfMessageFormatter();
         $record = [
             'level' => Logger::ERROR,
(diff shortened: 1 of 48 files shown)


tests/Monolog/Formatter/GelfMessageFormatterTest.php
```diff
@@ -82,10 +83,11 @@ public function testFormatWithFileAndLine()
 
     /**
      * @covers Monolog\Formatter\GelfMessageFormatter::format
-     * @expectedException InvalidArgumentException
      */
     public function testFormatInvalidFails()
     {
+        $this->expectException(\InvalidArgumentException::class);
```
gmponos: The reason the annotations are deprecated is in order to explicitly be added before the class that will throw the exception.. So this must be moved below.. and excactly above the function that will throw the exception

gmponos: Same thing for other places as well.

MarioBlazek: @gmponos thanks for the review. I will update PR accordingly.

Seldaek:
Thanks

Issue #590, opened 2015-06-24
Fallback Logging


inssein:
Are there any plans to add fallback mechanisms to logging? For example, I plan to use the ElasticSearchHandler, but if for whatever reason it cannot reach the elastic search instance, maybe it could be logged to disk?

I was thinking about perhaps checking the elastic search connection before creating the handler itself, but then this means that it will always open a connection to elastic search. Where as with Monolog, it only opens a connection if there is something to send.

Seldaek:
I think for such specific cases it's better to create a custom handler that takes in two handler and calls the second one if the first one throws.

Neirda24:
Hi Reopening this issue as it may lead to a feature of some kind.
We had exactly the same problem. So we created a custom handler based of the `WhatFailureGroupHandler` but added a break if no error was thrown (stop at first succeed)

But then we wanted to log the one that had fail because it means that one or more handlers failed and it is never a good sign... We have no idea as to how to implement it.

masacc:
Hi, I'm in the same team than @Neirda24 and here is the solution we implemented : 

Create a FallbackGroupHandler : `src/LogBundle/Handler/FallbackGroupHandler.php`

```
<?php

namespace LogBundle\Handler;

use Monolog\Handler\GroupHandler;
use Throwable;

class FallbackGroupHandler extends GroupHandler
{
    /**
     * {@inheritdoc}
     */
    public function handle(array $record)
    {
        if ($this->processors) {
            foreach ($this->processors as $processor) {
                $record = call_user_func($processor, $record);
            }
        }

        foreach ($this->handlers as $handler) {
            try {
                $handler->handle($record);
                break;
            } catch (Throwable $throwable) {
                // do nothing
            }
        }

        return false === $this->bubble;
    }

    /**
     * {@inheritdoc}
     */
    public function handleBatch(array $records)
    {
        foreach ($this->handlers as $handler) {
            try {
                $handler->handleBatch($records);
                break;
            } catch (Throwable $throwable) {
                // do nothing
            }
        }
    }
}
```
Define it as a service and inject into it two other handlers that we have previously declared as nested : 

`src/LogBundle/Resources/config/services/handler.yml` : 

```
services:
    my_monolog.handler.fallback_group:
        class: 'LogBundle\Handler\FallbackGroupHandler'
        arguments:
            - ['@monolog.handler.elasticsearch', '@monolog.handler.filesystem']
```

`app/config/config_dev.yml` : 

```
monolog:
[...]

PR #1361, opened 2019-08-13
Add FallbackGroupHandler

inverse:
- Adds `FallbackGroupHandler` to attempt to send to a series of handlers, breaking once one is successfully handled. 
  - Code adapted from https://github.com/Seldaek/monolog/issues/590#issuecomment-359484395
- Moved `ExceptionTestHandler` to own class - exception thrown before handled to simulate real world situation. Otherwise TestHandler would have the record.


Seldaek:
Thanks

PR #1362, opened 2019-08-14
Add explicity void return type to Monolog\Logger

nicolas-grekas:
As part of its migration to PHP 7.2, Symfony is starting to require more explicit annotations for return-type. ~Of course, Monolog cannot add real ones here, due to BC.~ These declarations would help us (and would make our CI green again as we now have a check to ensure the code and its deps are in good shape in this regard, see e.g. https://travis-ci.org/symfony/symfony/jobs/571836979)

Cannot be applied on branch 1.x since it has conflicting `@return bool`.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -466,7 +466,7 @@ class Logger implements LoggerInterface, ResettableInterface
      * @param string $message The log message
      * @param array  $context The log context
      */
-    public function log($level, $message, array $context = [])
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Now with real return-types, all green.
The change doesn't affect neither the Symfony bridge not the bundle if I made no mistakes.

Seldaek:
Thanks, let's hope this goes as planned ;)

PR #1364, opened 2019-08-19
Backport Interface and Trait from master to 1.X

lyrixx:
New features should have been done on 1.X and 2.X should be for cleanup.
Now it's too late, but we can ease migrations by backporting some
code to 1.X.

This PR will help supporting monolog 1.X and 2.X at the same time.
See https://github.com/symfony/symfony/pull/33241

### DIFF
--- a/src/Monolog/Handler/FormattableHandlerInterface.php
+++ b/src/Monolog/Handler/FormattableHandlerInterface.php
@@ -1,0 +1,38 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Handler;
+
+use Monolog\Formatter\FormatterInterface;
+
+/**
+ * Interface to describe loggers that have a formatter
+ *
+ * @internal This interface is present in monolog 1.x to ease forward compatibility.
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ */
+interface FormattableHandlerInterface
+{
+    /**
+     * Sets the formatter.
+     *
+     * @param  FormatterInterface $formatter
+     * @return HandlerInterface   self
+     */
+    public function setFormatter(FormatterInterface $formatter): HandlerInterface;
+
+    /**
+     * Gets the formatter.
+     *
+     * @return FormatterInterface
+     */
+    public function getFormatter(): FormatterInterface;
+}
--- a/.php_cs
+++ b/.php_cs
@@ -13,6 +13,12 @@ $finder = Symfony\CS\Finder::create()
     ->files()
     ->name('*.php')
     ->exclude('Fixtures')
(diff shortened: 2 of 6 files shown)


src/Monolog/Handler/FormattableHandlerInterface.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+interface FormattableHandlerInterface
+{
+    /**
+     * Sets the formatter.
+     *
+     * @param  FormatterInterface $formatter
+     * @return HandlerInterface   self
+     */
+    public function setFormatter(FormatterInterface $formatter): HandlerInterface;
```
stof: not compatible with the PHP 5.3 min version.

lyrixx: We don't use theses interface / trait in the core.
I think we can just ignore theses files in the linter, isn't ?

stof: well, the issue is that someone using them is forced to have a much different support matrix than monolog 1 itself. So if a project is supporting monolog 1 due to older PHP version, this won't help the migration.

nicolas-grekas: But then, that's not Monolog's problem.

lyrixx: For exemple, in Symfony 5, we support Monolog 1 and 2. but we also support only PHP 7+
I agree this is not perfect, but I try to fill the holes.
If remove the typehint, there will be a missmatch in Symfony, isn't ?
What do you suggest?

stof: well, if the only reason to add these interfaces in 1.x is to help the migration, I think it is.

nicolas-grekas: It's only an issue for not-yet written code that would like to use these interfaces while still supporting PHP 5.3. I think that Monolog doesn't need to care for such an edge case. The burden should be on the authors that have these crazy policies, which sound totally unrealistic btw to me.

nicolas-grekas:
FYI, I merged symfony/symfony#33241 and it fixed our tests, but we still have the issue that our `ElasticsearchLogstashHandler` cannot implement the new interface, because we want to keep compat with v1.

@lyrixx tests are still red, the linter doesn't skip :)

lyrixx:
> @lyrixx tests are still red, the linter doesn't skip :)

I know. I prefer to be sure it could be merged before spending some time on this PR.

Seldaek:
Sounds fine to merge for me, but I'd maybe add in the class-level docblocks that these are only for forward compat with 2.0 and not really used by monolog 1.x

Seldaek:
@lyrixx if you can fix the tests I'd like to release soonish :)

lyrixx:
@Seldaek It should be OK :) thanks for the review

Seldaek:
👍 thanks

Seldaek:
1.25 is out with this

lyrixx:
@Seldaek Thanks.
Actually, It does not work well, I open a new PR https://github.com/Seldaek/monolog/pull/1374
Sorry for the noise

PR #1367, opened 2019-08-30
Use svg for build status directly from travis-ci

pavog:
Since there is no reason to use shields for the build status badge we can use it directly from travis-ci.


Seldaek:
Thanks, was too lazy to figure out why this didn't show up :)

PR #1370, opened 2019-09-05
Remove versions from Laravel integration

driesvints:
We recently released v6.0 so maybe we shouldn't mention the versions anymore. Monolog will probably always remain with the framework.

Thanks for all your work on Monolog and Composer @Seldaek!


Seldaek:
👍 thanks

PR #1374, opened 2019-09-06
Fixed forward compatibilty layer

lyrixx:
I'm sorry the extra work. But while working on https://github.com/symfony/symfony/pull/33492 I notice traits were not forward compatible.

So I:

* removed the `@internal ` flag, as the goal of theses traits is to be used.
* fixed the signature, to be compatible with the current signature of `Mono
log\Handler\AbstractHandler::pushProcessor($callback)`

The error was:

```
  1x: The "Monolog\Handler\ProcessableHandlerTrait" trait is considered internal This interface is present in monolog 1.x to ease forward compatibility. It may change wit
hout further notice. You should not use it from "Symfony\Bridge\Monolog\Handler\ElasticsearchLogstashHandler".                                                            
    1x in DeprecationErrorHandler::shutdown from Symfony\Bridge\PhpUnit                                                                                                   
                                                                                                                                                                          
  1x: The "Monolog\Handler\FormattableHandlerTrait" trait is considered internal This interface is present in monolog 1.x to ease forward compatibility. It may change wit
hout further notice. You should not use it from "Symfony\Bridge\Monolog\Handler\ElasticsearchLogstashHandler".                                                            
    1x in DeprecationErrorHandler::shutdown from Symfony\Bridge\PhpUnit                                                                                                   
```

### DIFF
--- a/src/Monolog/Handler/FormattableHandlerInterface.php
+++ b/src/Monolog/Handler/FormattableHandlerInterface.php
@@ -16,7 +16,8 @@ use Monolog\Formatter\FormatterInterface;
 /**
  * Interface to describe loggers that have a formatter
  *
- * @internal This interface is present in monolog 1.x to ease forward compatibility.
+ * This interface is present in monolog 1.x to ease forward compatibility.
+ *
  * @author Jordi Boggiano <j.boggiano@seld.be>
  */
 interface FormattableHandlerInterface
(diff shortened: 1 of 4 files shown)


Seldaek:
Ah well, shit happens ;) 1.25.1 coming soon..

PR #1383, opened 2019-09-25
Support Proxy by Telegram Bot Handler

finagin:


src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -46,21 +46,41 @@ class TelegramBotHandler extends AbstractProcessingHandler
...
+     * Proxy
+     *
+     * @var string
+     */
+    private $proxy;
+
+    /**
+     * Proxy schemas
+     */
+    const SCHEMAS = [
```
stof: why making this a public constant ?

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -46,21 +46,41 @@ class TelegramBotHandler extends AbstractProcessingHandler
...
     /**
      * @param string $apiKey  Telegram bot access token provided by BotFather
      * @param string $channel Telegram channel name
+     * @param string $proxy   Proxy
      * @inheritDoc
      */
     public function __construct(
         string $apiKey,
         string $channel,
+        string $proxy = null,
```
stof: adding a new parameter in a non-final position is a BC break.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -83,6 +103,29 @@ protected function send(string $message): void
         $url = self::BOT_API . $this->apiKey . '/SendMessage';
         curl_setopt($ch, CURLOPT_URL, $url);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+
+        if ($this->proxy !== null) {
+            $porxy = parse_url($this->proxy);
```
stof: typo in the variable name.

tests/Monolog/Handler/TelegramBotHandlerTest.php
```diff
@@ -32,12 +32,13 @@ public function testSendTelegramRequest(): void
...
-     * @param string $channel
+     * @param string      $apiKey
+     * @param string      $channel
+     * @param string|null $proxy
      */
-    private function createHandler(string $apiKey = 'testKey', string $channel = 'testChannel'): void
+    private function createHandler(string $apiKey = 'testKey', string $channel = 'testChannel', string $proxy = null): void
     {
-        $constructorArgs = [$apiKey, $channel, Logger::DEBUG, true];
+        $constructorArgs = [$apiKey, $channel, $proxy, Logger::DEBUG, true];
```
alexandrmazur96: You'll fail this test - proxy parameter should be the last one.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -83,6 +103,29 @@ protected function send(string $message): void
...
         curl_setopt($ch, CURLOPT_URL, $url);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+
+        if ($this->proxy !== null) {
+            $proxy = parse_url($this->proxy);
+
+            $proxyHost = $proxy['host'];
+
+            if (array_key_exists('port', $proxy)) {
+                $proxyHost .= ':'.$proxy['port'];
```
alexandrmazur96: I guess you can highlight concatenation with spaces to make code more readable.
Also on line 119.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -83,6 +103,29 @@ protected function send(string $message): void
         $url = self::BOT_API . $this->apiKey . '/SendMessage';
         curl_setopt($ch, CURLOPT_URL, $url);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
+
+        if ($this->proxy !== null) {
+            $proxy = parse_url($this->proxy);
+
+            $proxyHost = $proxy['host'];
+
+            if (array_key_exists('port', $proxy)) {
```
alexandrmazur96: what about using `isset` instead of `array_key_exists`?
if key `port` in `$proxy` is `null` - you can get some issues with this.

finagin:
> Shouldn't curl already respect the `HTTP_PROXY` env variable automatically ?

You can use proxies only for telegrams

Seldaek:
I don't really agree this is an improvement.. Firstly either this should be done in a more generic way for all curl handlers, or not at all. 

But I also tend to agree with @stof that HTTP_PROXY env var should be enough here.. What's the use case for proxying only some of the traffic? That seems very odd and not something Monolog should support by default.

alexandrmazur96:
Use case - telegram is forbidden in Russia, but @finagin want to use it..

alexandrmazur96:
Maybe you are right, it seems that someone who want to use telegram with proxy can do it over the inheritance.

PR #1384, opened 2019-10-03
JsonFormatter: Don't even try to attempt normalizing iterators or generators in context

amino-garricklam:
Iterators and Generators may not be rewindable, so foreach is not safe
to use on them.

Iterators and especially Generators may trigger irreversible actions on
calling next(), so iterating over all values can potentially cause harm,
e.g. imagine an iterator over a set of HTTP POST requests that are sent
when the next value is requested . The only sufficiently safe thing to
iterate and include here are primitive arrays.

similar fix is applied to `NormalizerFormatter`: https://github.com/Seldaek/monolog/pull/891/files

It throws `Exception : Cannot traverse an already closed generator vendor/monolog/monolog/src/Monolog/Formatter/JsonFormatter.php:152`:
```php
    public function getGenerator(): \Generator
    {
        yield 111;
        yield 222;
    }

    public function testNormalizeClosedGenerator()
    {
        $formatter = new JsonFormatter();

        $generator = $this->getGenerator();
        foreach ($generator as $i) {
            // Consume generator...
        }

        $res = $formatter->format(array(
            'level_name' => 'CRITICAL',
            'channel' => 'test',
            'message' => 'bar',
            'context' => array($generator),
            'datetime' => new \DateTime,
            'extra' => array(),
        ));
    }
```

### DIFF
--- a/src/Monolog/Formatter/JsonFormatter.php
+++ b/src/Monolog/Formatter/JsonFormatter.php
@@ -145,7 +145,7 @@ class JsonFormatter extends NormalizerFormatter
             return 'Over 9 levels deep, aborting normalization';
         }
 
-        if (is_array($data) || $data instanceof \Traversable) {
+        if (is_array($data)) {
             $normalized = array();
 
             $count = 1;


src/Monolog/Formatter/JsonFormatter.php
```diff
@@ -145,7 +145,7 @@ protected function normalize($data, $depth = 0)
             return 'Over 9 levels deep, aborting normalization';
         }
 
-        if (is_array($data) || $data instanceof \Traversable) {
```
alexandrmazur96: This code throw fatal error only for `Generator`s . 

Can you just `is_array($data) || ($data instanceof \Traversable && !($data instanceof \Generator))`?

I guess idea to use only arrays is very bad.

amino-garricklam: thanks, changed

Seldaek: I'd tend to say rather remove this behavior entirely from Traversables, much like #891 did, otherwise there is a behavior difference between NormalizerFormatter and JsonFormatter.

amino-garricklam: Changed again.

Seldaek:
Thanks for the updates

PR #1385, opened 2019-10-11
Fix BrowserConsoleHandler formatting

trvrnrth:
This resolves an issue whereby all styles would be applied in reverse order to the formatting markers.

### Reproduction code
```
$logger = new Logger('test');
$handler = new BrowserConsoleHandler();
$handler->setFormatter(new LineFormatter(
	'[[blue text]]{color: blue} or [[green background]]{background-color: green; color: white}'
));
$logger->setHandlers([$handler]);
$logger->info('woo');
```

### Before
Selection_593
```
<script>(function (c) {if (c && c.groupCollapsed) {
c.log("%c%cblue text%c or %cgreen background%c", "font-weight: normal", "background-color: green; color: white", "font-weight: normal", "color: blue", "font-weight: normal");
}})(console);</script>
```

### After
Selection_592
```
<script>(function (c) {if (c && c.groupCollapsed) {
c.log("%c%cblue text%c or %cgreen background%c", "font-weight: normal", "color: blue", "font-weight: normal", "background-color: green; color: white", "font-weight: normal");
}})(console);</script>
```

### DIFF
--- a/src/Monolog/Handler/BrowserConsoleHandler.php
+++ b/src/Monolog/Handler/BrowserConsoleHandler.php
@@ -167,21 +167,22 @@ class BrowserConsoleHandler extends AbstractProcessingHandler
 
     private static function handleStyles(string $formatted): array
     {
-        $args = [static::quote('font-weight: normal')];
+        $args = [];
         $format = '%c' . $formatted;
         preg_match_all('/\[\[(.*?)\]\]\{([^}]*)\}/s', $format, $matches, PREG_OFFSET_CAPTURE | PREG_SET_ORDER);
 
         foreach (array_reverse($matches) as $match) {
-            $args[] = static::quote(static::handleCustomStyles($match[2][0], $match[1][0]));
             $args[] = '"font-weight: normal"';
+            $args[] = static::quote(static::handleCustomStyles($match[2][0], $match[1][0]));
 
             $pos = $match[0][1];
             $format = Utils::substr($format, 0, $pos) . '%c' . $match[1][0] . '%c' . Utils::substr($format, $pos + strlen($match[0][0]));
         }
 
-        array_unshift($args, static::quote($format));
+        $args[] = static::quote('font-weight: normal');
+        $args[] = static::quote($format);
 
-        return $args;
+        return array_reverse($args);
     }
 
     private static function handleCustomStyles(string $style, string $string): string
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/BrowserConsoleHandler.php
```diff
@@ -167,18 +167,19 @@ private static function generateScript(): string
...
     {
-        $args = [static::quote('font-weight: normal')];
+        $args = [];
         $format = '%c' . $formatted;
         preg_match_all('/\[\[(.*?)\]\]\{([^}]*)\}/s', $format, $matches, PREG_OFFSET_CAPTURE | PREG_SET_ORDER);
 
         foreach (array_reverse($matches) as $match) {
-            $args[] = static::quote(static::handleCustomStyles($match[2][0], $match[1][0]));
-            $args[] = '"font-weight: normal"';
+            array_unshift($args, '"font-weight: normal"');
```
alexandrmazur96: I don't see why you use `array_unshift` instead of `$arr[] = ...`

If you weren't use reversed styles maybe then would be better to "unreverse" it back? Because recalculating array index (**O(n)**) on every iteration looks terrible...

trvrnrth: The reason for doing it in reverse is to be able to substr in the %c placeholders a few lines below which I've not touched.

I guess the easiest alternative would be to build up args, then reverse as the last action.

alexandrmazur96: I guess it would be better.

P.S. Did you understand why `array_reverse($matches)` is used? Because I don't :(

trvrnrth: Yeah, it's because of the way the string substitution for replacing the format specifier matches with the console %c placeholder works based on the offsets returned by preg_match_all. In order for the offsets to remain correct as each match is processed it has to work from the end of the string. It's possible that this could be achieved in some other manner but I didn't fancy touching it!

I've pushed a change to replace the array_unshifts with a single array_reverse on return which is indeed a bit nicer.

Seldaek:
Thanks for the fix, merged in 1.x and master as 0ff3a9bd0399ddcf02f6b1ab80efca8f3ea2d6e2

PR #1393, opened 2019-10-30
Format exception log for Elasticsearch

Quynh-Nguyen:
For resolving the issues when working with Laravel

Class Error: `QueryException`

```php
        "status" => 400
        "error" => array:3 [
          "type" => "mapper_parsing_exception"
          "reason" => "failed to parse field [context.exception.code] of type [long] in document with id 'sPh3G24Bss4rAxTOzNRI'"
          "caused_by" => array:2 [
            "type" => "illegal_argument_exception"
            "reason" => "For input string: "42S22""
          ]
        ]
```

Or got the error the message log too long.

### DIFF
--- a/src/Monolog/Formatter/ElasticsearchFormatter.php
+++ b/src/Monolog/Formatter/ElasticsearchFormatter.php
@@ -50,6 +50,10 @@ class ElasticsearchFormatter extends NormalizerFormatter
     {
         $record = parent::format($record);
 
+        if ($this->recordHasContext($record) && $this->contextHasException($record['context'])) {
+            $record['context']['exception'] = $this->getContextException($record['context']['exception']);
+        }
+
         return $this->getDocument($record);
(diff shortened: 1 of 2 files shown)


Seldaek:
It seems from your error message that the problem is the exception's getCode() returned a `long` instead of `int`. That seems to be due to PDOException returning error codes as strings. I guess NormalizerFormatter should cast the code to int which would fix it at the source.

Quynh-Nguyen:
@Seldaek 
Yes I understood sir.
Beside the getCode() return `long` instead of `int`. The QueryException also sending the `previous` exception which make the exception message too long for sending to `Elasticsearch`.
Let me update my pull request.

Seldaek:
I fixed the type error at least. The length one I am not sure we can do much about, as this is kind of dependent on the exact exception and content of what you log?

PR #1400, opened 2019-11-15
Fix misleading @throws tag in StreamHandler doc

tpraxl:
The `\Exception` for non-existing and not creatable paths is thrown earliest on first write.
Not during StreamHandler instantiation, as claimed by the constructor's `@throws` tag.

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -32,7 +32,7 @@ class StreamHandler extends AbstractProcessingHandler
     private $dirCreated;
 
     /**
-     * @param resource|string $stream         If a missing path can't be created, an Exception will be thrown on first write
+     * @param resource|string $stream         If a missing path can't be created, an UnexpectedValueException will be thrown on first write
      * @param string|int      $level          The minimum logging level at which this handler will be triggered
      * @param bool            $bubble         Whether the messages that are handled can bubble up the stack or not
      * @param int|null        $filePermission Optional file permissions (default (0644) are only for owner read/write)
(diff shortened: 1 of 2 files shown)


tests/Monolog/Handler/StreamHandlerTest.php
```diff
@@ -184,33 +184,40 @@ public function testWriteNonExistingFileResource()
...
+        try {
+            $handler = new StreamHandler($nonExistingAndNotCreatablePath);
+        } catch (\Exception $fail) {
+            $this->fail(
+                'A non-existing and not creatable path should throw an Exception earliest on first write.
+                 Not during instantiation.'
+            );
+        }
+
         $this->expectException(\Exception::class);
```
Seldaek: This could be changed to `\UnexpectedValueException::class` I believe?

src/Monolog/Handler/StreamHandler.php
```diff
@@ -32,13 +32,12 @@ class StreamHandler extends AbstractProcessingHandler
     private $dirCreated;
 
     /**
-     * @param resource|string $stream
+     * @param resource|string $stream         If a missing path can't be created, an Exception will be thrown on first write
```
Seldaek: Should say UnexpectedValueException not Exception?

tests/Monolog/Handler/StreamHandlerTest.php
tpraxl: Sorry, I somehow missed your review. I just committed an update, that expects an `\UnexpectedValueException::class`.

src/Monolog/Handler/StreamHandler.php
tpraxl: The latest update fixes it.

Seldaek:
Thanks

Issue #1401, opened 2019-11-15
1.25.2 issue in FilterHandler:  "Notice: Undefined offset: -1"


dmaicher:
Monolog version 1.25.2

Since this new release we started to get some errors:

```json
   "ctxt_exception": {
      "code": 0,
      "class": "ErrorException",
      "message": "Notice: Undefined offset: -1",
      "file": "/var/www/foo/releases/2019_11_14_11_04_12/vendor/monolog/monolog/src/Monolog/Handler/FilterHandler.php:131",
      "trace": [
        "/var/www/foo/releases/2019_11_14_11_04_12/vendor/monolog/monolog/src/Monolog/Handler/FingersCrossedHandler.php:92",
        "/var/www/foo/releases/2019_11_14_11_04_12/vendor/monolog/monolog/src/Monolog/Handler/FingersCrossedHandler.php:113",
        "/var/www/foo/releases/2019_11_14_11_04_12/vendor/monolog/monolog/src/Monolog/Logger.php:344",
        "/var/www/foo/releases/2019_11_14_11_04_12/vendor/monolog/monolog/src/Monolog/Logger.php:707",
        "/var/www/foo/releases/2019_11_14_11_04_12/src/TheController.php:51",
        "/var/www/foo/releases/2019_11_14_11_04_12/vendor/symfony/http-kernel/HttpKernel.php:151",
        "/var/www/foo/releases/2019_11_14_11_04_12/vendor/symfony/http-kernel/HttpKernel.php:68",
        "/var/www/foo/releases/2019_11_14_11_04_12/vendor/symfony/http-kernel/Kernel.php:198",
        "/var/www/foo/releases/2019_11_14_11_04_12/web/app.php:21"
      ]
   }
```

It happens because of this change:
https://github.com/Seldaek/monolog/commit/65f1f304d4e68759970a5c663ad2ae46faa47013#diff-d69c48eda429ffdda49a91d61f9126e5R131

For us in those cases `$filtered` is an empty array after the foreach loop.

PR #1402, opened 2019-11-15
FilterHandler: fix handle empty batch of filtered records

dmaicher:
This fixes https://github.com/Seldaek/monolog/issues/1401


Seldaek:
Thanks

PR #1406, opened 2019-12-06
Fix a bug when using string-based loglevels in the `tr_TR` locale

mpdude:
PHP's `strtoupper()` is locale-aware, and `strtoupper('i')` (= `strtoupper(chr(105))`) will not be `I`(= `chr(73)`) but `chr(221)`, the "latin capital letter i with dot above" İ character.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -527,8 +527,13 @@ class Logger implements LoggerInterface, ResettableInterface
      */
     public static function toMonologLevel($level)
     {
-        if (is_string($level) && defined(__CLASS__.'::'.strtoupper($level))) {
-            return constant(__CLASS__.'::'.strtoupper($level));
+        if (is_string($level)) {
+            // Contains chars of all log levels and avoids using strtoupper() which may have
+            // strange results depending on locale (for example, "i" will become "İ")
+            $upper = strtr($level, 'abcdefgilmnortuwy', 'ABCDEFGILMNORTUWY');
+            if (defined(__CLASS__.'::'.$upper)) {
+                return constant(__CLASS__ . '::' . $upper);
+            }
         }
 
         return $level;


mpdude:
What is the cause of the PHPUnit failures here? 

https://travis-ci.org/Seldaek/monolog/jobs/621819572#L235

🤷🏻‍♂️

src/Monolog/Logger.php
```diff
@@ -527,8 +527,11 @@ public static function getLevelName($level)
      */
     public static function toMonologLevel($level)
     {
-        if (is_string($level) && defined(__CLASS__.'::'.strtoupper($level))) {
-            return constant(__CLASS__.'::'.strtoupper($level));
+        if (is_string($level)) {
+            $upper = mb_strtoupper($level);
```
jderusse: `ext-mbstring` should be added to composer.json.

Seldaek: Nope.. we should rather not require mb_strtoupper. This can be done in another way like `strtoupper(str_replace('i', 'I', $level))` - hackish but seems to work.

Seldaek:
Not sure what the travis failures are about, could be due to `sudo:true` changing the build env.. IMO you can skip the test and CI change, and rather add a comment to the function explaining why it doesn't use a straight up strtoupper to make sure nobody "optimizes" this fix away in the future.

src/Monolog/Logger.php
jderusse: Indeed.. May I suggest to use `strtr($level, 'abcdefgilmnortuwy', 'ABCDEFGILMNORTUWY');`which is faster?

List of chars are extract from current level, but have to be updated if new levels are added (test have to cover it)

mpdude: @jderusse I think your suggestion is pretty clever, will try that

mpdude:
Fixed as suggested & tests green.

Seldaek:
Thanks

PR #1409, opened 2019-12-10
Support Proxy by Slack webhook Handler

MilesChou:
Like #1383 , we have some use case to proxy only for Slack.


gmponos:
Wanna check this? https://github.com/gmponos/monolog-slack

PR #1410, opened 2019-12-10
Fix JSON formatting of resource type

theravel:
I think it is well-known that `json_encode` does not handle variables of `resource` type properly. Easiest example is:

    php -r "var_dump(json_encode(['test' => curl_init()]));" 
    // returns bool(false)

JsonFormatter of Monolog is affected as well, so it returns empty line if there is a resource in "context".

This pull request leverages already existing code from `NormalizerFormatter` to convert resources to `sprintf('[resource(%s)]', get_resource_type($data))`.


Seldaek:
Thanks, merged in 1.x & master

PR #1411, opened 2019-12-11
Correct pronoun in readme for contributors

finwe:
"which" is associated with things, "who" is more suitable for humans.


Seldaek:
Thanks

PR #1412, opened 2019-12-12
Set phpredis multi() mode parameter

fbnfgc:
Set the default function parameter to avoid potential issues when a Redis proxy is used instead of the internal class.

This issue has been discussed in https://github.com/phpredis/phpredis/issues/1476#issuecomment-445968070 and https://github.com/snc/SncRedisBundle/issues/442.

This may be Symfony specific, but I'd be happy to hear your thoughts on this issue.

### DIFF
--- a/src/Monolog/Handler/RedisHandler.php
+++ b/src/Monolog/Handler/RedisHandler.php
@@ -73,7 +73,8 @@ class RedisHandler extends AbstractProcessingHandler
     protected function writeCapped(array $record)
     {
         if ($this->redisClient instanceof \Redis) {
-            $this->redisClient->multi()
+            $mode = defined('\Redis::MULTI') ? \Redis::MULTI : 1;
+            $this->redisClient->multi($mode)
                 ->rpush($this->redisKey, $record["formatted"])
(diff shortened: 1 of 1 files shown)


Seldaek:
This is fine I guess, but should be rebased on 1.x branch as the problem exists there too.

fbnfgc:
Just cherry-picked the commit and applied it to the `1.x` branch. 

Tests now fails but it shouldn't. 

```
1) Monolog\Handler\RedisHandlerTest::testRedisHandleCapped
Error: Undefined class constant 'MULTI'
```

I need to dig into, some tests were skipped because `phpredis` is not available on the CI.

Seldaek:
Yeah the extension isn't loaded I believe in the tests.. Either mark more tests as skipped or maybe fix the patch so it uses `defined('Redis::MULTI') ? Redis::MULTI : 1` or whatever the value of Redis::MULTI is when the ext is loaded..

fbnfgc:
Thanks, fixed as suggested.

Seldaek:
Thanks

PR #1419, opened 2020-01-07
Update html color

noniagriconomie:
Hi,

Just a PR that remap new html colors on bootstrap ones, that IMO a lot of developers are visually used to :)
cf https://github.com/twbs/bootstrap/blob/master/site/data/colors.yml
(exepts for the `alert` where  i took the `critical` but darker)

before:
before

after:
HtmlFormatter_ logLevels_color


Seldaek:
Thanks

PR #1425, opened 2020-01-14
Introduce Stackdriver formatter

theravel:
We actively use such formatter in production for a long time, so I decided to propose making it available for the community.

The formatter is useful with the following setup:

- Application is containerized and orchestrated using Google Kubernetes Engine (GKE);
- Application writes logs to stderr/stdout;
- Log entries are collected from stdout/stderr by GKE, then [forwarded to Stackdriver](https://cloud.google.com/monitoring/kubernetes-engine/).

The problem:
Stackdriver uses JSON-structured logs, but does not understand "level" field, therefore it cannot understand the level of log entry. On the other hand, level names match Monolog level names, the difference is just a field name ("severity" instead of "level" or "level_name").

Other references:
People use similar approach for other technical stack as well, for example Python article: https://medium.com/retailmenot-engineering/formatting-python-logs-for-stackdriver-5a5ddd80761c

### DIFF
--- a/src/Monolog/Formatter/StackdriverFormatter.php
+++ b/src/Monolog/Formatter/StackdriverFormatter.php
@@ -1,0 +1,34 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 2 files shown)


theravel:
@Seldaek can you take a look please?

theravel:
@Seldaek  can you take a look please?

Seldaek:
Please rather publish this as a standalone package and add it to https://github.com/Seldaek/monolog/wiki/Third-Party-Packages - it makes more sense than baking support for everything into monolog, we already have too many handlers/formatters.

PR #1426, opened 2020-01-17
Apply php-cs-fixer fix

DQNEO:
just applied `php-cs-fixer fix` by the latest version of php-cs-fixer.


Seldaek:
cs fixes are to be applied by maintainers only, thanks.

PR #1430, opened 2020-01-30
Fix typehint for AbstractHandler level parameter

alexander-schranz:
The level parameter of the `AbstractHandler` supports also the PSR levels as strings e.g. `\Psr\Log\LogLevel::ERROR` or from a ENV variable `CUSTOM_ERROR_HANDLER_LEVEL=error`.

This will fix issues with phpstan or other static code analyzer when given the level as string.


Seldaek:
Thanks

Issue #1435, opened 2020-03-06
Quotation error ( ` )


rickserrat:
Monolog version 1|2

Error:
NOTICE: PHP message: PHP Fatal error:  Uncaught TypeError: preg_match() expects parameter 2 to be string, null given in /var/www/html/vendor/monolog/monolog/src/Monolog/Processor/GitProcessor.php:54

Remove: 

> ```$branches = `git branch -v --no-abbrev`;```

**Update:**` $branches = 'git branch -v --no-abbrev';`

rickserrat:
@Seldaek Can I make a pull request?

Seldaek:
This is not a quotation error, the backticks are executing the process https://www.php.net/manual/en/language.operators.execution.php. But I guess it should check that $branches is not null before calling preg_match. Feel free to PR that to 1.x branch.

rickserrat:
Pull Request 1.x #1438

PR #1438, opened 2020-03-13
Check that  is not null before calling preg_match

rickserrat:
Regarding Issue #1435


rickserrat:
@Seldaek it's my first PR, what reason for error?

Seldaek:
I think you can ignore the error. Thanks for the fix.

PR #1439, opened 2020-03-20
Make null handler processable

m-h-1:
to fix https://github.com/symfony/monolog-bundle/issues/335


Seldaek:
Sorry but that seems more like a bug in the MonologBundle than in Monolog. It is misusing a handler.. There is a ProcessableHandlerInterface which can be checked before calling stuff like pushProcessor

PR #1441, opened 2020-03-27
Ensure usage of global curl_* functions

### LINKED ISSUE #1444: "Attempted to call function "curl_init" from namespace "Monolog\Handler"."
Monolog version 2

Using the Slack Webhook handler broke on our PHP 7.3 install (Ubuntu 18.04) with a Uncaught PHP Exception Symfony\Component\ErrorHandler\Error\UndefinedFunctionError: "Attempted to call function "curl_init" from namespace "Monolog\Handler".". Prefixing the curl_* calls with a backslash to ensure the global methods are used fixed it, and as far as I know there's no downsides to being more precise with these things.

doenietzomoeilijk:
Using the Slack Webhook handler broke on our PHP 7.3 install (Ubuntu 18.04) with a `Uncaught PHP Exception Symfony\Component\ErrorHandler\Error\UndefinedFunctionError: "Attempted to call function "curl_init" from namespace "Monolog\Handler"."`. Prefixing the `curl_*` calls with a backslash to ensure the global methods are used fixed it, and as far as I know there's no downsides to being more precise with these things.

Fixes #1444

### DIFF
--- a/src/Monolog/Handler/SlackWebhookHandler.php
+++ b/src/Monolog/Handler/SlackWebhookHandler.php
@@ -95,7 +95,7 @@ class SlackWebhookHandler extends AbstractProcessingHandler
         $postData = $this->slackRecord->getSlackData($record);
         $postString = Utils::jsonEncode($postData);
 
-        $ch = curl_init();
+        $ch = \curl_init();
         $options = array(
             CURLOPT_URL => $this->webhookUrl,
             CURLOPT_POST => true,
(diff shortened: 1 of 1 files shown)


Seldaek:
I'd like to understand exactly what causes this failure.. Because it sounds to me like the failure you see is only possible if `curl` extension is missing entirely, which would not be fixed by adding backslashes.

PR #1442, opened 2020-04-01
JsonFormatter: add option to ignore empty `context` and `extra` fields

carusogabriel:
Similar to efe572cb1074.

### DIFF
--- a/src/Monolog/Formatter/JsonFormatter.php
+++ b/src/Monolog/Formatter/JsonFormatter.php
@@ -28,16 +28,18 @@ class JsonFormatter extends NormalizerFormatter
 
     protected $batchMode;
     protected $appendNewline;
+    protected $ignoreEmptyContextAndExtra;
 
     /**
      * @var bool
      */
     protected $includeStacktraces = false;
 
-    public function __construct(int $batchMode = self::BATCH_MODE_JSON, bool $appendNewline = true)
+    public function __construct(int $batchMode = self::BATCH_MODE_JSON, bool $appendNewline = true, bool $ignoreEmptyContextAndExtra = false)
     {
         $this->batchMode = $batchMode;
         $this->appendNewline = $appendNewline;
+        $this->ignoreEmptyContextAndExtra = $ignoreEmptyContextAndExtra;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Monolog/Formatter/JsonFormatter.php
```diff
@@ -68,13 +70,24 @@ public function isAppendingNewlines(): bool
...
+
         if (isset($normalized['context']) && $normalized['context'] === []) {
             $normalized['context'] = new \stdClass;
         }
         if (isset($normalized['extra']) && $normalized['extra'] === []) {
             $normalized['extra'] = new \stdClass;
         }
 
+        if ($this->ignoreEmptyContextAndExtra) {
+            if ((array) $normalized['context'] === []) {
```
Seldaek, line 82: I think it might make more sense to add a `if $(this->ignoreEmptyContextAndExtra)` check above at line 75/78 to do unset in that case instead of creating a stdClass and then casting it to array here and unsetting it again.. That seems a little wasteful.

Seldaek:
Also may I know what the use case is for this? For LineFormatter it kinda made sense to clean up the output as it's meant for human consumption, but in the json output I see less value.

carusogabriel:
@Seldaek We have an internal logging system that is manipulating each log record.

At the end of the process, `extra` is an empty array, and this system expects an object. 

We'll force an object for it with something like `{"extra": {"_force_object": true}}`, but ideally we'd like to remove if it's empty :)

Seldaek:
But the new stdClass code there was exactly about forcing it to be an object.. Are you decoding/re-encoding it somehow? Because it seems to me like the type should be correct when coming out of JsonFormatter

carusogabriel:
> But the new stdClass code there was exactly about forcing it to be an object

Indeed: `$normalized['extra'] = new \stdClass;`

> Are you decoding/re-encoding it somehow? Because it seems to me like the type should be correct when coming out of JsonFormatter

I'll check that, we have some custom Monolog Processors as well, they might be changing that 👍

Seldaek:
Processors happen before the formatter. If you have problems I'd imagine it's something you do later with the log file, if you json_decode with $assoc=true, then you get an empty array, and if you json_encode that you'll encode it back into an array instead of an object.

Anyway I guess the option here doesn't hurt so I'll merge, but technically it sounds like you're doing something wrong somewhere.

PR #1451, opened 2020-04-10
Telegram Handler: support additional API parameters

Pascal-So:
The telegram API for sendMessage supports some additional parameters that where not yet covered with the current implementation. See the API documentation here:

https://core.telegram.org/bots/api#sendmessage

In my use case I'm mostly just interested in the `parse_mode` parameter but I thought that while I'm on it I could also add support for `disable_web_page_preview` and `disable_notification`.

I didn't add support for the other remaining options because in my opinion they will never make sense when using the API for logging. `reply_to_message_id` can only be set for replies, but when logging we're not replying to a message but rather sending independent messages. `reply_markup` lets the bot display an interactive interface for the user receiving the message, but the log handler can't deal with replies anyway.

Something that I'm not sure about whether it's okay like this is the function argument order to the constructor. Should I move the new arguments to the end?

### DIFF
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -59,19 +59,19 @@ class TelegramBotHandler extends AbstractProcessingHandler
      * The kind of formatting that is used for the message.
      * See available options at https://core.telegram.org/bots/api#formatting-options
      * or in AVAILABLE_PARSE_MODES
-     * @var string
+     * @var string|null
      */
     private $parseMode;
 
     /**
      * Disables link previews for links in the message.
-     * @var bool
+     * @var bool|null
      */
     private $disableWebPagePreview;
 
     /**
      * Sends the message silently. Users will receive a notification with no sound.
-     * @var bool
+     * @var bool|null
      */
     private $disableNotification;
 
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -54,13 +73,19 @@ class TelegramBotHandler extends AbstractProcessingHandler
     public function __construct(
         string $apiKey,
         string $channel,
+        string $parseMode = null,
+        bool $disableWebPagePreview = null,
+        bool $disableNotification = null,
```
Seldaek, line 78: New args should be added to the end to be backwards compatible. And adding setters for those custom options is usually a good idea to allow for a more readable configuration code.. e.g. https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/SlackHandler.php#L192

Pascal-So:
I've added some simple validation for the parse mode in the setter, plus tests for the validation.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -46,6 +55,26 @@ class TelegramBotHandler extends AbstractProcessingHandler
...
+
+    /**
+     * Disables link previews for links in the message.
+     * @var bool
+     */
+    private $disableWebPagePreview;
+
+    /**
+     * Sends the message silently. Users will receive a notification with no sound.
+     * @var bool
```
Seldaek, line 74: As far as I can tell all 3 vars will default to null and if not configured will remain so. Those `@var` should probably include `|null`.

Seldaek:
Thanks for the updates!

PR #1462, opened 2020-05-15
Fix failures when SoapFault Detail is an object.

neclimdul:


neclimdul:
I don't know if this is the right approach but it is _an_ approach :-D Let me know if you'd like it implemented in a different way.

Seldaek:
Thanks, switched it to use $this->toJson so encoding errors can be ignored but otherwise the approach is fine IMO

PR #1466, opened 2020-05-27
Update ErrorLogHandler.php

jagui:
`$record['formatted']` contains no `\r\n` escape sequences but ` $record['message']` does, hence should be the source string for split.


Seldaek:
`$record['formatted']` may contain newlines, that depends on how it gets formatted. If the formatted string includes stack traces for example it will have new lines.

PR #1472, opened 2020-06-04
Fix array offset access on null in RavenHandler

jdecool:
PHP 7.4 introduces backward incompatible changes and not allow anymore to [access a null variable like an array](https://www.php.net/manual/en/migration74.incompatible.php#migration74.incompatible.core.non-array-access).

The `RavenHandler` use an `array_reduce` with the default initial value that is `null` but acces it like an array which generate a PHP notice.

before


Seldaek:
Thanks

jdecool:
Thanks @Seldaek 

Is it possible to tag a version with this fix ?

Seldaek:
1.25.5 / 2.1.1 are out

jdecool:
Nice ! 

Thanks

PR #1475, opened 2020-06-14
Static analysis improvements

gmponos:
Hi.. although you are not using static analysis these docblocks can improve static analysis on userland.

- I used `psalm-*` in order for the normal annotations to not mess the IDEs
- The `psalm-*` annotation works both for psalm and phpstan. I think that also psalm reads `phpstan-*` annotations.
- Since you don't use static analysis I only changed the public api.

More things could be done. Like adding to levels something like this:

```php
psalm-param 100|200|300|400|550|600|LogLevel::* $level
```

But it wont be 100% accurate since level can accept `info` and `INFO` and `iNFO` so I believe it's not worth the overhead.

### DIFF
--- a/src/Monolog/Handler/FilterHandler.php
+++ b/src/Monolog/Handler/FilterHandler.php
@@ -49,6 +49,8 @@ class FilterHandler extends Handler implements ProcessableHandlerInterface, Rese
     protected $bubble;
 
     /**
+     * @psalm-param callable(array, HandlerInterface): HandlerInterface|HandlerInterface $handler
+     *
      * @param callable|HandlerInterface $handler        Handler or factory callable($record|null, $filterHandler).
      * @param int|array                 $minLevelOrList A list of levels to accept or a minimum level if maxLevel is provided
      * @param int|string                $maxLevel       Maximum level to accept, only used if $minLevelOrList is not an array
--- a/src/Monolog/Handler/FilterHandler.php
+++ b/src/Monolog/Handler/FilterHandler.php
@@ -49,7 +49,7 @@ class FilterHandler extends Handler implements ProcessableHandlerInterface, Rese
     protected $bubble;
 
     /**
-     * @psalm-param callable(array, HandlerInterface): HandlerInterface|HandlerInterface $handler
+     * @psalm-param HandlerInterface|callable(?array, HandlerInterface): HandlerInterface $handler
      *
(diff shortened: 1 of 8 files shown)


src/Monolog/Logger.php
```diff
@@ -140,6 +140,8 @@ class Logger implements LoggerInterface, ResettableInterface
     protected $exceptionHandler;
 
     /**
+     * @psalm-param array<callable(array): array>
```
Seldaek, line 143: This is missing a variable name for which $param this applies to

src/Monolog/Handler/SwiftMailerHandler.php
```diff
@@ -28,6 +28,8 @@ class SwiftMailerHandler extends MailHandler
     private $messageTemplate;
 
     /**
+     * @psalm-param callable(string, array): Swift_Message|Swift_Message $message
```
Seldaek, line 31: This seems ambiguous especially in light of php8's union types.. Is this a callable which returns `Swift_Message OR Swift_Message`, or is it a `callable OR a Swift_Message` (that's the correct one of course, but it reads super weird IMO right now).

src/Monolog/Handler/FilterHandler.php
```diff
@@ -49,6 +49,8 @@ class FilterHandler extends Handler implements ProcessableHandlerInterface, Rese
     protected $bubble;
 
     /**
+     * @psalm-param callable(array, HandlerInterface): HandlerInterface|HandlerInterface $handler
```
Seldaek, line 52: should be array|null, as per the comment at line 54

gmponos: Fixed.

src/Monolog/Handler/SwiftMailerHandler.php
gmponos: I changed the order in all similar parts add first the `Swift_message` and then the `callable`

src/Monolog/Logger.php
gmponos: fixed.

src/Monolog/Handler/SwiftMailerHandler.php
Seldaek: Thanks yeah that's better :)

PR #1477, opened 2020-06-19
[Sendgrid] Depend on the time of the first record

gmponos:
Shouldn't this time depend on the time of the first record instead?

### DIFF
--- a/src/Monolog/Handler/SendGridHandler.php
+++ b/src/Monolog/Handler/SendGridHandler.php
@@ -11,6 +11,7 @@
 
 namespace Monolog\Handler;
 
+use Monolog\DateTimeImmutable;
 use Monolog\Logger;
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Hmm, I am not so sure actually. In SwiftMailerHandler we use current time, in NativeMailerHandler no date is specified so it's also current time.. in use cases with long running scripts and such, the first log might be very old, while the email is sent whenever an error occurs usually and thus I think it's probably better to keep the email date at $now.

If you really want to know when something happen, look at the date of the log record you are interested in, that's much more reliable.

gmponos:
True.. let's not merge this now.. let's give it a second though and come back...

Seldaek:
I think I'd rather not change this as per my comment above.

PR #1480, opened 2020-07-08
Replace call_user_func

gmponos:
Based on 

https://github.com/fab2s/call_user_func

and 
https://github.com/symfony/symfony/pull/29309

thanks @fab2s

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -312,7 +312,7 @@ class Logger implements LoggerInterface, ResettableInterface
 
         try {
             foreach ($this->processors as $processor) {
-                $record = call_user_func($processor, $record);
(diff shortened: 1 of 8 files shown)


src/Monolog/Logger.php
```diff
@@ -312,7 +312,7 @@ public function addRecord(int $level, string $message, array $context = []): boo
 
         try {
             foreach ($this->processors as $processor) {
-                $record = call_user_func($processor, $record);
+                $record = $processor($record);
```
gmponos, line 315: I believe this is the place where we might see the results of this.. all the others are not offering much in performance.. just in order to be consistent.

Seldaek:
Thanks

PR #1481, opened 2020-07-13
Added support for working with BufferHandler

gvozdb:
Added `handleBatch` method and fixed `write` method for `TelegramBotHandler`.

This is necessary to work with `BufferHandler`, `OverflowHandler`, etc.

### DIFF
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -149,15 +149,14 @@ class TelegramBotHandler extends AbstractProcessingHandler
      */
     protected function write(array $record): void
     {
-        $this->send($record['formatted'], [$record['formatted']]);
+        $this->send($record['formatted']);
     }
 
     /**
      * Send request to @link https://api.telegram.org/bot on SendMessage action.
      * @param string $message
-     * @param array $records
      */
-    protected function send(string $message, array $records): void
+    protected function send(string $message): void
     {
         $ch = curl_init();
         $url = self::BOT_API . $this->apiKey . '/SendMessage';
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -150,6 +150,7 @@ class TelegramBotHandler extends AbstractProcessingHandler
     /**
      * Send request to @link https://api.telegram.org/bot on SendMessage action.
      * @param string $message
+     * @param array $records
      */
     protected function send(string $message, array $records): void
     {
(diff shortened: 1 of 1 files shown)


gvozdb:
Why checking failed – http://joxi.ru/Grq0DGEck5BxpA ? I don't understand.

axell-brendow:
> Why checking failed – http://joxi.ru/Grq0DGEck5BxpA ? I don't understand.

Checking failed for PHP:nightly

gvozdb:
@axell-brendow What is this? It's important?

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -122,12 +122,31 @@ public function disableNotification(bool $disableNotification = null): self
...
+        }
+    }
+
     /**
      * @inheritDoc
      */
     protected function write(array $record): void
     {
-        $this->send($record['formatted']);
+        $this->send($record['formatted'], [$record['formatted']]);
```
Seldaek, line 149: I don't see how this fixes anything? The send method receives only one argument.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -122,12 +122,31 @@ public function disableNotification(bool $disableNotification = null): self
...
+        foreach ($records as $record) {
+            if ($record['level'] < $this->level) {
+                continue;
+            }
+            $messages[] = $this->processRecord($record);
+        }
+
+        if (!empty($messages)) {
+            $this->send((string) $this->getFormatter()->formatBatch($messages), $messages);
+        }
```
Seldaek, line 141: This should look more like the parent handle method IMO:

```suggestion
        $messages = [];

        foreach ($records as $record) {
            if (!$this->isHandling($record)) {
                continue;
            }

            if ($this->processors) {
                $record = $this->processRecord($record);
            }

            $messages[] = $record;
        }

        if (!empty($messages)) {
            $this->send((string) $this->getFormatter()->formatBatch($messages));
        }
```

src/Monolog/Handler/TelegramBotHandler.php
gvozdb: @Seldaek Yeah. I missed it. I've added some corrections to the next commit.

Seldaek: I still don't get the point, $records isn't used in the send method, so why is this passed at all?

Seldaek: Also handleBatch calls send with a single argument, so it's broken now that you added the second required argument.

Seldaek:
Ok fixed things up and merged, thanks

PR #1484, opened 2020-07-22
Improve exception message

axell-brendow:
This exception is commonly thrown in Laravel applications when using Docker.
But it is not clear why the system could not open a file that has read permission
for all users. So, when I came to the code of Monolog, I saw that you try to open it
in append mode, so I think that just adding this information could help a lot other
users.


axell-brendow:
@Seldaek CI failing in this case is because of PHP:nightly exclusively right ?

Seldaek:
Yeah you can ignore the failure.

Seldaek:
Thanks, merged  in 1.x/master as bcca19cee0ed507f342d39688a071e7c989393d8

PR #1488, opened 2020-08-13
Adds support for unix sockets

kimpepper:
The SyslogUdpHandler currently hard codes the `AF_INET` socket domain, however we can also easily support local unix sockets with `AF_UNIX`. See https://www.php.net/manual/en/function.socket-create.php

If the port is 0, the we can assume it is a local unix socket.

### DIFF
--- a/src/Monolog/Handler/SyslogUdpHandler.php
+++ b/src/Monolog/Handler/SyslogUdpHandler.php
@@ -51,7 +51,7 @@ class SyslogUdpHandler extends AbstractSyslogHandler
         $this->ident = $ident;
         $this->rfc = $rfc;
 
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/SyslogUdpHandler.php
```diff
@@ -51,7 +51,7 @@ public function __construct(string $host, int $port = 514, $facility = LOG_USER,
         $this->ident = $ident;
         $this->rfc = $rfc;
 
-        $this->socket = new UdpSocket($host, $port ?: 514);
+        $this->socket = new UdpSocket($host, $port);
```
kimpepper, line 54: The default is already set as a default parameter.

PR #1489, opened 2020-08-13
Speed up Logger::addRecord

AntonAcc:
I was wondering if next/current is faster than foreach. The test

```<?php

$startTime = microtime(true);
for ($i = 1; $i < 1000000; $i++) {
    $handlerList = [];
    $handlerList[] = 'a';
    $handlerList[] = 'b';
    $handlerList[] = 'c';
    $handlerList[] = 'd';

    foreach ($handlerList as $key => $handler) {
    }
}
echo 'foreach      - ' . (microtime(true) - $startTime) . "\n";

$startTime = microtime(true);
for ($i = 1; $i < 1000000; $i++) {
    $handlerList = [];
    $handlerList[] = 'a';
    $handlerList[] = 'b';
    $handlerList[] = 'c';
    $handlerList[] = 'd';

    while ($handler = current($handlerList)) {
        next($handlerList);
    }
}
echo 'next/current - ' . (microtime(true) - $startTime) . "\n";
```
```
foreach      - 0.17546820640564
next/current - 0.38347506523132
```
http://sandbox.onlinephpfunctions.com/code/0c1d5b1a723309ce669c1d8348b3617def6d3d6b

shows that foreach is faster than next/current.

Then I've done tests

```<?php

class Handler
{
    private $level = 100;

    private $isHandling = false;

    public function __construct($isHandling)
    {
        $this->isHandling = $isHandling;
    }

    public function isHandling(array $record)
    {
        $result = $record['level'] >= $this->level;

        return $this->isHandling;
    }

    public function handle(array $record)
    {
        $this->isHandling($record);
    }
}

// Current algorithm
$startTime = microtime(true);
for ($i = 1; $i < 100000; $i++) {
    $handlerList = [];
    $handlerList[] = new Handler(false);
    $handlerList[] = new Handler(false);
    $handlerList[] = new Handler(true);
    $handlerList[] = new Handler(false);

    $record = ['level' => 300];
// check if any handler will handle this message so we can return early and save cycles
    $handlerKey = null;
    foreach ($handlerList as $key => $handler) {
        if ($handler->isHandling($record)) {
            $handlerKey = $key;
            break;
        }
    }

    if (null === $handlerKey) {
        continue;
    }

// advance the array pointer to the first handler that will handle this record
    reset($handlerList);
    while ($handlerKey !== key($handlerList)) {
        next($handlerList);
    }

    while ($handler = current($handlerList)) {
        if (true === $handler->handle($record)) {
            break;
        }

        next($handlerList);
    }
}
echo (microtime(true) - $startTime) . "\n";

// Preparing $handleableHandlers algorithm
$startTime = microtime(true);
for ($i = 1; $i < 100000; $i++) {
    $handlerList = [];
    $handlerList[] = new Handler(false);
    $handlerList[] = new Handler(false);
    $handlerList[] = new Handler(true);
    $handlerList[] = new Handler(false);

    $record = ['level' => 300];

    $handleableHandlers = [];
    foreach ($handlerList as $handler) {
        if ($handler->isHandling($record)) {
            $handleableHandlers[] = $handler;
        }
    }

    if (!$handleableHandlers) {
        continue;
    }

    foreach ($handleableHandlers as $handler) {
        if (true === $handler->handle($record)) {
            break;
        }
    }
}
echo (microtime(true) - $startTime) . "\n";

// Suggested algorithm
$startTime = microtime(true);
for ($i = 1; $i < 100000; $i++) {
    $handlerList = [];
    $handlerList[] = new Handler(false);
    $handlerList[] = new Handler(false);
    $handlerList[] = new Handler(true);
    $handlerList[] = new Handler(false);

    $record = ['level' => 300];

    $offset = 0;
    foreach ($handlerList as $handler) {
        if ($handler->isHandling($record)) {
            break;
        }

        $offset++;
    }
    // cut off checked not handling handlers
    $remainedHandlers = array_slice($handlerList, $offset);

    if (!$remainedHandlers) {
        continue;
    }

    foreach ($remainedHandlers as $handler) {
        if (true === $handler->handle($record)) {
            break;
        }
    }
}
echo (microtime(true) - $startTime) . "\n";
```
```
0.16189312934875
0.12785387039185
0.13638997077942
```
http://sandbox.onlinephpfunctions.com/code/6323a783b033b0d6b8217336d5f1b358e15d2f47

As a result, a current algorithm is slowest among others.
The algorithm with preparing $handleableHandlers is faster than current, but slower than suggested. However it has more clean code than suggested.
The suggested algorithm is fastest among others and it is more useful in context of logger than more clean code of algorithm with preparation of $handleableHandlers 

PS: However tests on http://sandbox.onlinephpfunctions.com shows that algorithm with preparation of $handleableHandlers is fastest. Plus, it has a more clean code. Maybe, it worth to use this one. I'll be glad to read your opinions.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -285,16 +285,18 @@ class Logger implements LoggerInterface, ResettableInterface
      */
     public function addRecord(int $level, string $message, array $context = []): bool
     {
-        // check if any handler will handle this message so we can return early and save cycles
-        $handlerKey = null;
-        foreach ($this->handlers as $key => $handler) {
+        $offset = 0;
+        foreach ($this->handlers as $handler) {
             if ($handler->isHandling(['level' => $level])) {
-                $handlerKey = $key;
                 break;
             }
-        }
 
-        if (null === $handlerKey) {
+            $offset++;
+        }
+        // cut off checked not handleable handlers
+        $remainedHandlers = array_slice($this->handlers, $offset);
+
+        if (!$remainedHandlers) {
             return false;
         }
 
@@ -315,18 +317,10 @@ class Logger implements LoggerInterface, ResettableInterface
                 $record = $processor($record);
             }
 
-            // advance the array pointer to the first handler that will handle this record
-            reset($this->handlers);
-            while ($handlerKey !== key($this->handlers)) {
-                next($this->handlers);
-            }
-
-            while ($handler = current($this->handlers)) {
+            foreach ($remainedHandlers as $handler) {
                 if (true === $handler->handle($record)) {
                     break;
                 }
-
-                next($this->handlers);
             }
         } catch (Throwable $e) {
             $this->handleException($e, $record);


Seldaek:
Thanks, merged and simplified this even further in https://github.com/Seldaek/monolog/commit/df4d93f14828b8052665699bd4ba2ab257f6dca2 as I realized we can actually just do it one loop, I think it's faster than doing array_slice etc, and it reads less confusing.

PR #1496, opened 2020-09-03
Support for extended RFC5424 timestamp format

Pierre-Lannoy:
As described in [RFC5424 Timestamp format](https://tools.ietf.org/html/rfc5424#section-6.2.3), it is allowed to send second fraction as doted mili (3 digits) or micro (6 digits) in syslog messages. I suggest this modest PR to support it.


Seldaek:
Thanks

PR #1499, opened 2020-09-15
Split SlackRecord into SlackRecord + SlackFormatter

TimFinucane:
Moves the message formatting parts of SlackRecord into SlackFormatter. This allows the SlackFormatter to be swapped out by users to create new message formats.

100% coverage on SlackFormatter and SlackRecord.

Api Changes:

- Deprecates:
  - SlackHandler::useAttachment
  - SlackHandler::useShortAttachment
  - SlackHandler::includeContextAndExtra
  - SlackHandler::excludeFields
  - SlackRecord::COLOR_*

Doesn't deprecate the arguments in SlackHandler's constructor or SlackWebhookHandler's constructor, as the SlackFormatter is created by default to preserve the majority of use-cases.

- Breaking API Changes:
  - When using SetFormatter on SlackHandler or SlackWebhookHandler. Doing so will mean useAttachment, useShortAttachment, includeContextAndExtra, and excludeFields are all ignored, and any formatted text produced by the formatter will be directly added to the text in the sent slack message.
  - SlackRecord's useAttachment, useShortAttachment, includeContextAndExtra, and excludeFields.
  - Direct usage of SlackRecord will result in different outputs, however messages sent by SlackHandler and SlackWebhookHandler are preserved.

Benefits:
  - Users can now easily design and create their own slack message formatters, which can produce any kind of slack message, including e.g. using the new Blocks api: https://api.slack.com/reference/block-kit/blocks

### DIFF
--- a/src/Monolog/Formatter/SlackFormatter.php
+++ b/src/Monolog/Formatter/SlackFormatter.php
@@ -1,0 +1,255 @@
+<?php
+
+
+namespace Monolog\Formatter;
+
+
+use Monolog\Logger;
+use Monolog\Utils;
+
+/**
+ * Formats the message as a Slack message with the schema {@link https://api.slack.com/messaging/composing/layouts}
+ *
+ * @author Tim Finucane <timfinucane@outlook.com>
+ */
+class SlackFormatter extends NormalizerFormatter
+{
+    public const COLOR_DANGER = 'danger';
+
+    public const COLOR_WARNING = 'warning';
+
+    public const COLOR_GOOD = 'good';
+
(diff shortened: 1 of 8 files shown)


Seldaek:
The breaking changes are kind of a no-go here I'm afraid. I'd suggest maybe using https://github.com/gmponos/monolog-slack instead and contributing there. I guess the slack handler here should maybe be deprecated in favor of that project.

Seldaek:
That said, thanks for the PR, it seems like you did a decent job of it, but I can't just break BC in a minor release, and have currently no plans for a major release.

PR #1501, opened 2020-09-28
Mark FormattableHandlerTrait::$formatter as nullable

erikjwaxx:
This fixes a Psalm error at level 2 and above when extending `AbstractProcessingHandler`:

```
$formatter is not defined in constructor...and in any private or final methods called in the constructor (see https://psalm.dev/074)
```

### DIFF
--- a/src/Monolog/Handler/FormattableHandlerTrait.php
+++ b/src/Monolog/Handler/FormattableHandlerTrait.php
@@ -22,9 +22,9 @@ use Monolog\Formatter\LineFormatter;
 trait FormattableHandlerTrait
 {
     /**
-     * @var FormatterInterface
+     * @var ?FormatterInterface
      */
-    protected $formatter;
+    protected $formatter = null;
 
     /**
      * {@inheritdoc}
--- a/src/Monolog/Handler/FormattableHandlerTrait.php
+++ b/src/Monolog/Handler/FormattableHandlerTrait.php
@@ -24,7 +24,7 @@ trait FormattableHandlerTrait
     /**
      * @var ?FormatterInterface
      */
-    protected $formatter = null;
+    protected $formatter;
 
     /**
      * {@inheritdoc}


src/Monolog/Handler/FormattableHandlerTrait.php
```diff
@@ -22,9 +22,9 @@
 trait FormattableHandlerTrait
 {
     /**
-     * @var FormatterInterface
+     * @var ?FormatterInterface
      */
-    protected $formatter;
+    protected $formatter = null;
```
stof, line 27: this change is useless. It is already the case in PHP

erikjwaxx: With an untyped property, yes, the `null` default is implicit.

If this were formally a PHP 7.4 typed property `protected ?FormatterInterface $formatter`, then `FormattableHandlerTrait::getHandler` would throw an `E_ERROR` due to the access of an uninitialized property. Example [here](https://3v4l.org/O0TgZ).

So my reasoning is as follows:
* Under PHP < 7.4 and/or not using typed properties, this change is _redundant_ but not _wrong_.
* Under PHP >= 7.4 with typed properties, it is no longer _redundant_ and is in fact _necessary_.
* On first inspection, there doesn't seem to be any difference in the compiled opcodes between the two versions.
* Hence, adding this initialization implicitly gives you forward-compatibility if you choose to implement typed properties in some future iteration of the library, and at least appears to not have any runtime impact.

This is bikeshedding to the extreme, I know. I am not attached to the explicit instantiation here in any case: what I'm really after is the type annotation that `$formatter` is nullable so Psalm level 2 will quit barking about my custom log handler not being initialized :smiley: .

pallagun: I'm also seeing this issue in v7.4.11.  If the fix operates as claimed it'd be beneficial to have this.

stof: but this trait does not use typed properties.
And I did not say that the change was wrong, but that it was useless (and inconsistent with the coding standards used throughout the package)

erikjwaxx: That's fine then, I can remove it

e87d182

Seldaek:
Thanks, already fixed this earlier as I was adding phpstan to the project tho.

PR #1503, opened 2020-10-01
Adding Redis Pub/Sub handler (closes #574)

### LINKED ISSUE #574: Add Pub/Sub (channel) support to RedisHandler
Currently the RedisHandler is only providing the 'rpush' option for lists, while logstash redis input plugin provides support for both lists and channels ( https://www.elastic.co/guide/en/logstash/current/plugins-inputs-redis.html ). Having the channel option is useful when we want to read the input from multiple keys, using 'psubscribe' to read from channels based on a certain pattern.

gfaugere:
This adds a basic Handler which pushes messages to a Pub/Sub Redis channel.

Let me know if you think it should be rewritten as part of the existing Redis handler - or if this is fine!

### DIFF
--- a/src/Monolog/Handler/RedisPubSubHandler.php
+++ b/src/Monolog/Handler/RedisPubSubHandler.php
@@ -1,0 +1,67 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks, looks fine to me as is, very simple.

PR #1507, opened 2020-10-27
Allow string numeric log levels

adrianpanicek:
This allows for log level to be string numeric in case of setting from environment variables or other sources with ambivalent type


Seldaek:
Right, good fix I guess although in those cases I'd prefer using the human readable name like ERROR instead of 400 because nobody knows what that means.

PR #1512, opened 2020-11-07
Adds return value in handler wrapper

perevoshchikov:
Method `setFormatter` in `HandlerWrapper` always throws an exception, fixed it.


Seldaek:
Woops good catch :)

PR #1514, opened 2020-11-19
remove @internal annotation

idevin:
getClass is not internal method


stof:
Well, why not ?

Seldaek:
Eh well, I guess it's harmless to allow third parties to use this. It was internal initially because it was added on the Registry class where it really did not make a ton of sense, but now on the Utils class it's unlikely to ever go away.

PR #1515, opened 2020-11-26
Added JSON_PARTIAL_OUTPUT_ON_ERROR to the default json encoding options

zerkms:
Especially when monolog logs the exception traces (including arguments) there is a chance one of the arguments would contain recursive references, which `json_encode` by default cannot handle.

`JSON_PARTIAL_OUTPUT_ON_ERROR` option turns those recursive references to `null` instead.

It's available from php 5.5 hence I think should be okay to be added by default.


zerkms:
Okay, I realised that it breaks the corresponding test: `testIgnoresRecursiveObjectReferences`

So please hold on then.

The question then is: would the project want to have it at all? (then I will fix tests and write more if necessary)

Seldaek:
I think it makes sense yeah, we rather keep as much data as possible in there, fixed the tests quickly.

Seldaek:
Thanks for the tip, didn't realize this was there somehow :)

PR #1516, opened 2020-11-30
make methods protected access level so developer can override

milos-silni:


stof:
This makes the maintenance a lot harder. Allowing protected access to properties means it is almost impossible to refactor the class without it being a BC break.

Please describe the use case you have that would justify that cost.

milos-silni:
> This makes the maintenance a lot harder. Allowing protected access to properties means it is almost impossible to refactor the class without it being a BC break.
> 
> Please describe the use case you have that would justify that cost.

I wanted to override writingIsTimedOut method and add option to use ms instead of seconds

milos-silni:
in that case i will create a new pull request with changes from comment above

Seldaek:
Closing as #1517 is indeed a much better approach than overriding code :)

PR #1517, opened 2020-11-30
add support to define write-timeout in ms

milos-silni:


src/Monolog/Handler/SocketHandler.php
```diff
@@ -368,7 +369,10 @@ private function writingIsTimedOut(int $sent): bool
             usleep(100);
         }
 
-        if ((time() - $this->lastWritingAt) >= $writingTimeout) {
+        // convert to ms
+        $lastWritingMs = (time() - $this->lastWritingAt) * 1000;
```
Seldaek, line 373: I guess this (and lastWritingAt) var should be based on microtime(true) and not time(). That'd make it more precise (because otherwise if you determine a write-timeout as 0.5, this code would anyway only timeout after a second has passed), and you would not have to multiply anything by 1000 anymore.

Seldaek: Fixed this myself in https://github.com/Seldaek/monolog/commit/3ee78ae7313327e199b4fcc2a28f1d2d0d48d0ba

PR #1523, opened 2020-12-11
Support for Elastica client up to v7.0.0, with optional drop support for ES types

### LINKED ISSUE #1399: ElasticaHandler is incompatible with Elastica 7
Starting from version 7 (which is in a beta stage now), the `\Elastica\Type` class was removed (since specifying types in requests is deprecated in ES 7.x).

So the handler will result in:
`Call to undefined method Elastica\Document::setType`

I can cook up a PR if you want.

patrickkusebauch:
If you use the Elastica 7 client, which is only compatible with v7 of ElasticSearch, support of types has been dropped. This PR tests for the Elastica client version and conditionally drops types if they are no longer supported.

It replaces #1476 that mistakes `OpType` (operation type) parameter for ES types.

Resolves #1399

### DIFF
--- a/tests/Monolog/Handler/ElasticaHandlerTest.php
+++ b/tests/Monolog/Handler/ElasticaHandlerTest.php
@@ -288,7 +288,7 @@ class ElasticaHandlerTest extends TestCase
     protected function getDocSourceFromElastic(Client $client, $index, $type, $documentId)
     {
         if($type === null) {
-            $path  = "/{$index}/{$documentId}";
+            $path  = "/{$index}/_doc/{$documentId}";
         } else {
             $path  = "/{$index}/{$type}/{$documentId}";
         }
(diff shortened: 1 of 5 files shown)


tests/Monolog/Handler/ElasticaHandlerTest.php
```diff
@@ -226,13 +281,18 @@ protected function getCreatedDocId(Response $response)
...
-     * @param  string $type
+     * @param  ?string $type
      * @param  string $documentId
      * @return array
      */
     protected function getDocSourceFromElastic(Client $client, $index, $type, $documentId)
     {
-        $resp = $client->request("/{$index}/{$type}/{$documentId}", Request::GET);
+        if($type === null) {
+            $path  = "/{$index}/{$documentId}";
```
stof, line 291: the new API endpoint to get a document is `/index_name/_doc/document_id`

patrickkusebauch: Fixed in https://github.com/Seldaek/monolog/pull/1523/commits/79d12135dd08837f5a595d9699fa56a5303f1c05

Seldaek:
Thanks

PR #1531, opened 2021-01-15
Reset wrapped handler of the FilterHandler

marcvdm:
When we reset the FilterHandler we should call the handler we are wrapping.


Seldaek:
Thanks

PR #1532, opened 2021-01-20
Replace api_user and api_key with Bearer token

wolfram-agentis:
The Bearer Token seems to be how the v3 API will be handling authorization, which the v2 API currently accepts as well.

### DIFF
--- a/src/Monolog/Handler/SendGridHandler.php
+++ b/src/Monolog/Handler/SendGridHandler.php
@@ -75,8 +75,6 @@ class SendGridHandler extends MailHandler
     protected function send(string $content, array $records): void
     {
         $message = [];
-        $message['api_user'] = $this->apiUser;
-        $message['api_key'] = $this->apiKey;
         $message['from'] = $this->from;
         foreach ($this->to as $recipient) {
             $message['to[]'] = $recipient;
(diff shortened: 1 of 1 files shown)


stof:
is this the same value to be passed ?

stof:
this does not use the apiUser at all anymore AFAICT.

wolfram-agentis:
Yup, that's correct. I based this off of Twilio's v2 API docs which shows the cURL requests using only the Bearer auth (https://www.twilio.com/docs/sendgrid/api/v2/mail#curl-examples).

Seldaek:
Not against this, but then we should make sure apiUser accepts null too, as it's not strictly required anymore.

Seldaek:
And well, it seems v2 API is still supported so I'd say there isn't a lot of value in this change unless we also change to v3 API which is more future proof I guess: https://sendgrid.com/docs/api-reference/

selcukcukur:
It would be more correct to apply this in a way that it works in harmony with both versions instead of applying it this way. For this, we can add an api version information and format it accordingly.

wolfram-agentis:
should I just close this PR?

Seldaek:
Well it depends if you are up for adjusting the PR or not.

troyporter:
I am very late to this discussion but i would like to add now that this handler just doesn't work as-is without making the exact change the original poster suggested.  

I kept getting a bad username/password response from the API and was confused until i dug into it and found this in the logs:

image

Once I added the bearer token header it worked perfectly.  I think this change to the handler is needed at this time.

Can we reconsider this PR?

Seldaek:
I'm happy to reconsider if someone is willing to make a PR migrating to api v3 completely. That'd be more sensible IMO.

stof:
Note that if the API Key expected in v2 and v3 is different, doing such change would be a BC break for any project using it in a working way with v2 credentials.

Seldaek:
Yeah the question is.. is this working for anyone? We should just deprecate/remove the handler entirely and let someone publish it as an external lib IMO

PR #1533, opened 2021-02-04
Fluentd formatter: format as dictionnary for Record Accessor support

sbocahu:
Fluentbit, a lightweight Fluentd shipper, only supports dictionnary records for its record accessor functionnality ( https://docs.fluentbit.io/manual/administration/configuring-fluent-bit/record-accessor )
Using keys instead of array indexes is generally easier and less error-prone in any case.
This PR changes the Fluentd formatting to return a dictionnary instead of an array.

### DIFF
--- a/src/Monolog/Formatter/FluentdFormatter.php
+++ b/src/Monolog/Formatter/FluentdFormatter.php
@@ -73,7 +73,7 @@ class FluentdFormatter implements FormatterInterface
             $message['level_name'] = $record['level_name'];
         }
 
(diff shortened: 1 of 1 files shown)


Seldaek:
Two issues I have here:

- The tests are broken
- This may not be backwards compatible? I am not knowledgeable enough about Fluentd to say if this would cause issues for existing setups?

selcukcukur:
@Seldaek Test file corruption is normal. This requires new controls and is unnecessary, there is no situation that will cause an error right now. I think this withdrawal request should not be sent.

PR #1536, opened 2021-02-28
Stop using deprecated MockBuilder::setMethods in unit tests

pryazhnikov:
`MockBuilder::setMethods()` is marked as deprecated since PHPUnit 8, but it's used in monolog tests.
In this pull request I've replaced these calls to `onlyMethods()` / `addMethods()`.

Summary of changes:
1. the replace `setMethods()` -> `onlyMethods()` was enough in most cases
2. `DynamoDbHandlerTest` has a tricky case (Aws\DynamoDb\DynamoDbClient:: formatAttributes exists in aws/aws-sdk-php v2, but was removed in v3.0, but its call was tested based on the version)
3. AmqpHandlerTest requires the checking of `\PhpAmqpLib\Channel\AMQPChannel::__destruct` existence before the onlyMethods() usage (because __destruct was removed in php-amqplib/php-amqplib:2.5.0)
4. RedisHandlerTest requires the usage of `addMethods()` because Predis\Client uses magic method `__call()` for commands processing

### DIFF
--- a/tests/Monolog/Handler/AmqpHandlerTest.php
+++ b/tests/Monolog/Handler/AmqpHandlerTest.php
@@ -33,7 +33,7 @@ class AmqpHandlerTest extends TestCase
         $messages = [];
 
         $exchange = $this->getMockBuilder('AMQPExchange')
-            ->setMethods(['publish', 'setName'])
+            ->onlyMethods(['publish', 'setName'])
             ->disableOriginalConstructor()
             ->getMock();
 
@@ -85,7 +85,7 @@ class AmqpHandlerTest extends TestCase
         $messages = [];
 
         $exchange = $this->getMockBuilder('PhpAmqpLib\Channel\AMQPChannel')
-            ->setMethods(['basic_publish', '__destruct'])
+            ->onlyMethods(['basic_publish', '__destruct'])
             ->disableOriginalConstructor()
             ->getMock();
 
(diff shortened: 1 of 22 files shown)


Seldaek:
Thanks. Looks all fine to me except for the 7.2/lowest CI build which uses aws-sdk 2.4.9 and outputs:

```
1) Monolog\Handler\DynamoDbHandlerTest::testHandle
Trying to configure method "formatAttributes" which cannot be configured because it does not exist, has not been specified, is final, or is static
```

Not sure why that is to be honest, as the code change you did looks good to me.

pryazhnikov:
Thanks for noticing this.
This warning definitely should be fixed. I'll investigate this problem later.

pryazhnikov:
Finally I've found out the reasons of this warning.

`Aws\DynamoDb\DynamoDbClient` from aws/aws-sdk-php 2.4.9 has method `formatAttributes`. So mock creation call looks like 
```
        $this->client = $this->getMockBuilder('Aws\DynamoDb\DynamoDbClient')
            ->onlyMethods(['__call', 'formatAttributes'])
            ->addMethods([])
            ->disableOriginalConstructor()
            ->getMock();
```

`addMethods` with an empty array breaks the list of methods given by `onlyMethods` call.
So I have fixed it by calling addMethods for non-empty absent methods list only.

I've checked the outputs of all test runs and looks like this warning was fixed completely.

Seldaek:
Great, thanks!

Issue #1542, opened 2021-03-17
TelegramBotHandler throws when Telegram API returns "message is too long" error


sergiks:
Monolog version 2.0.0

Telegram Bot API [sendMessage()][1] method has a message length limit of "4096 characters after entities parsing"

Added logging to my Telegram bot. Since stack trace might get long, Telegram API returns an error which is thrown unhandled by TelegramBotHandler:

    Fatal error: Uncaught RuntimeException: Telegram API error.
    Description: Bad Request: message is too long 
    in /var/www/prod/lumen/vendor/monolog/monolog/src/Monolog/Handler/TelegramBotHandler.php:96

Solution options:

* trim message length to 4096
* split long message into multiple messages
* attach long messages as a text file attachment to a message
* strip stack trace from message

[1]: https://core.telegram.org/bots/api#sendmessage

alexandrmazur96:
Hi!

I have introduced PR some time ago. Before (**and if!**) it will be merged you can alternatively handle this by yourself:

Create a new `TelegramBotHandler` class in your project and extend it from `Monolog\TelegramBotHandler`, then override proper methods in your class (you can use code from my PR).

Seldaek:
Fixed by #1616

PR #1544, opened 2021-03-26
Fixed TelegramBotHandler throws unhandled runtime exception "message is too long"

alexandrmazur96:
Fix for issue - https://github.com/Seldaek/monolog/issues/1542

### DIFF
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -195,7 +195,7 @@ class TelegramBotHandler extends AbstractProcessingHandler
      * Create a curl resource for further requesting Telegram API.
      *
      * @param string $message
-     * @return CurlHandle|false|resource
+     * @return \CurlHandle|false|resource
      */
     protected function prepareCurlHandle(string $message)
     {
@@ -247,7 +247,7 @@ class TelegramBotHandler extends AbstractProcessingHandler
                 // the message will be chunked into separate messages with the maximum length limit.
                 if (function_exists('mb_str_split')) {
                     $splittedMessages = mb_str_split($message, self::MESSAGE_LENGTH_LIMIT - 3);
-                    if ($splittedMessages === false) {
+                    if (!is_array($splittedMessages)) {
                         throw new RuntimeException('Unable to split too long message');
                     }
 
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -149,33 +167,105 @@ public function handleBatch(array $records): void
...
+                    }
+
+                    $lastChunk = end($splittedMessages);
+                    foreach ($splittedMessages as &$splittedMessage) {
+                        if ($lastChunk === $splittedMessage) {
+                            break;
+                        }
+
+                        $splittedMessage .= '...';
+                    }
```
Seldaek, line 261: There is an early return missing

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -149,33 +167,105 @@ public function handleBatch(array $records): void
...
+                }
+
+                return str_split($message, self::MESSAGE_LENGTH_LIMIT);
+            // It's most likely cause 'message is too long' exception from Telegram API.
+            case self::DO_NOTHING_WITH_TOO_LONG_MESSAGE:
+            default:
+                return [$message];
```
Seldaek, line 268: This seems kinda pointless, I assume you did it for BC but if it just causes failures always, I'd rather see some reasonable default strategy in place instead of keeping an old broken strategy. I'd say trimming by default sounds good, and offering splitting as an alternative.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -30,6 +30,11 @@
  */
 class TelegramBotHandler extends AbstractProcessingHandler
 {
+    public const MESSAGE_LENGTH_LIMIT = 4096;
+    public const TRIM_TOO_LONG_MESSAGE = 'trim';
+    public const SPLIT_TOO_LONG_MESSAGE_INTO_MULTIPLE_MESSAGES = 'split';
```
Seldaek, line 35: I'd rename these to LONG_MESSAGE_TRIM and LONG_MESSAGE_SPLIT I think, then we can more easily type-hint this for phpstan/psalm as being a parameter of type `self::LONG_MESSAGE_*`.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -30,6 +30,11 @@
  */
 class TelegramBotHandler extends AbstractProcessingHandler
 {
+    public const MESSAGE_LENGTH_LIMIT = 4096;
```
Seldaek, line 33: Probably should be private const? Or at least protected.

Seldaek:
Closing as it is fixed by #1616

PR #1545, opened 2021-03-27
MailHandler: Use array access to compare first character

glensc:


src/Monolog/Handler/MailHandler.php
```diff
@@ -70,7 +70,7 @@ protected function getHighestRecord(array $records): array
 
     protected function isHtmlBody(string $body): bool
     {
-        return substr($body, 0, 1) === '<';
+        return $body[0] === '<';
```
xabbuh, line 73: This will break when an empty string is passed.

glensc: @xabbuh `$body` is always `string` (type hinted), so that can not happen!

Seldaek:
Sorry if I'm stupid here but.. what's the point of this change exactly?

glensc:
@Seldaek my ide suggested to use array access rather function call. typically it's considered better practice to use array offset access for single byte rather substr call.

Seldaek:
Ok sure why not I suppose..

PR #1546, opened 2021-04-07
Removes unused html features.

selcukcukur:
HTML does not use these features anymore, it is considered invalid, so it is used with CSS markup.

See : https://dev.w3.org/html5/pf-summary/obsolete.html

### DIFF
--- a/src/Monolog/Formatter/HtmlFormatter.php
+++ b/src/Monolog/Formatter/HtmlFormatter.php
@@ -59,7 +59,7 @@ class HtmlFormatter extends NormalizerFormatter
             $td = '<pre>'.htmlspecialchars($td, ENT_NOQUOTES, 'UTF-8').'</pre>';
         }
 
-        return "<tr style=\"padding: 4px;text-align: left;\">\n<th style=\"vertical-align: top;background: #ccc;color: #000\" width=\"100\">$th:</th>\n<td style=\"padding: 4px;text-align: left;vertical-align: top;background: #eee;color: #000\">".$td."</td>\n</tr>";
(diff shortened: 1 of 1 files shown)


stof:
This HTMLFormatter is targetting email clients, which are known to be bad at supporting modern HTML.

selcukcukur:
> This HTMLFormatter is targetting email clients, which are known to be bad at supporting modern HTML.

@stof I can understand this, but still these features are no longer functional for HTML. They need to be formatted with CSS.

stof:
@selcukcukur are they really not functional anymore in these old email clients implementing the HTML from 15 years ago ?
And if modern clients don't support them anymore while older clients don't support setting them through CSS, the solution is to set it both ways, not only in the modern way.

selcukcukur:
> @selcukcukur are they really not functional anymore in these old email clients implementing the HTML from 15 years ago ?
> And if modern clients don't support them anymore while older clients don't support setting them through CSS, the solution is to set it both ways, not only in the modern way.

To be clear, I believe this formatter needs to be further edited, but for now, this solution will suffice. Do you mean the old client doesn't support the css code?

stof:
I know that email clients are quite bad at supporting CSS styles (see https://www.caniemail.com/). but I don't know whether it is an issue for these particular changes.

selcukcukur:
> I know that email clients are quite bad at supporting CSS styles (see https://www.caniemail.com/). but I don't know whether it is an issue for these particular changes.

this may be true for advanced formatting, but this level is not a problem for simple formatting.

Seldaek:
Frankly the current format seems to work very well, so I am not so keen on making changes as it requires careful testing and if it's not broken I don't see much point in trying to modernize the email format. It's only risks and very little benefit.

Seldaek:
Closing as does not seem worth the risk.

PR #1547, opened 2021-04-07
Improvements have been applied to HTML Formatter.

selcukcukur:


src/Monolog/Formatter/HtmlFormatter.php
```diff
@@ -136,4 +133,9 @@ protected function convertToString($data): string
 
         return Utils::jsonEncode($data, JSON_PRETTY_PRINT | Utils::DEFAULT_JSON_FLAGS, true);
     }
+    
+    protected function convertToSpecialChars($data): string
```
stof, line 137: why making this a protected method ?

selcukcukur: I forgot to fix it, after copying from the above function, now I'm fixing it.

Seldaek: I don't think public is better :) Probably should be private?

src/Monolog/Formatter/HtmlFormatter.php
```diff
@@ -54,12 +54,11 @@ public function __construct(?string $dateFormat = null)
...
     {
-        $th = htmlspecialchars($th, ENT_NOQUOTES, 'UTF-8');
-        if ($escapeTd) {
-            $td = '<pre>'.htmlspecialchars($td, ENT_NOQUOTES, 'UTF-8').'</pre>';
+       if ($escapeTd) {
+            $td = "<pre>{$this->convertToSpecialChars($td)}</pre>";
         }
 
-        return "<tr style=\"padding: 4px;text-align: left;\">\n<th style=\"vertical-align: top;background: #ccc;color: #000\" width=\"100\">$th:</th>\n<td style=\"padding: 4px;text-align: left;vertical-align: top;background: #eee;color: #000\">".$td."</td>\n</tr>";
+        return "<tr style=\"padding: 4px;text-align: left;\">\n<th style=\"vertical-align: top;background: #ccc;color: #000\" width=\"100\">{$th}:</th>\n<td style=\"padding: 4px;text-align: left;vertical-align: top;background: #eee;color: #000\">{$this->convertToSpecialChars($td)}</td>\n</tr>";
```
Seldaek, line 61: This change isn't correct. th is not escaped anymore above, and td is escaped once or twice depending on $escapeTd

src/Monolog/Formatter/HtmlFormatter.php
```diff
@@ -71,9 +70,7 @@ protected function addRow(string $th, string $td = ' ', bool $escapeTd = true):
      */
     protected function addTitle(string $title, int $level): string
     {
-        $title = htmlspecialchars($title, ENT_NOQUOTES, 'UTF-8');
-
-        return '<h1 style="background: '.$this->logLevels[$level].';color: #ffffff;padding: 5px;" class="monolog-output">'.$title.'</h1>';
+        return "<h1 style=\"background: {$this->logLevels[$level]}; color: #ffffff;padding: 5px;\" class=\"monolog-output\">{$this->convertToSpecialChars($title)}</h1>";
```
Seldaek, line 73: Frankly I'm not sure if having function calls within an interpolated string is what I would consider an improvement here..

PR #1551, opened 2021-04-19
SyslogHandler.php phpdoc update

selcukcukur:
By default, this setting is made. Therefore, the handler can now be added directly to the stack.


Seldaek:
Thanks but this is just an example, it shows how you can configure things, doesn't mean it's required IMO.

PR #1553, opened 2021-05-07
Set StreamHandler stream chunk size, fixes #1552

### LINKED ISSUE #1552: Interleaved log records using StreamHandler
Monolog version 1.24.0
PHP version 7.0.33

We get interleaved log records while using `StreamHandler`. The under-laying PHP stream writer implements a chunked write with a default chunk size of 8192 bytes. In an application with many concurrent processes writing to a single log file this leads to interleaved log records.

Example (if the chunk size would be 4):
```
my log line 1
my lmy log line 3
og line 2 
my log line 4
```

This issue could be fix in a simular way as in #1129 by calling `stream_set_chunk_size`.

dsch:
This change ensures atomic log record writing to the output stream.

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -23,6 +23,8 @@ use Monolog\Utils;
  */
 class StreamHandler extends AbstractProcessingHandler
 {
+    protected const MAX_CHUNK_SIZE = 2147483647;
+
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks, I'm hoping this won't have side effects for non-file:// streams tho, because that's a LARGE chunk size :)

PR #1554, opened 2021-05-14
Fix deprecation warning on PHP 8.1

derrabus:
PHP 8.1 issues a deprecation warning for Monolog:

> DateTime::__construct(): Passing null to parameter #1 ($datetime) of type string is deprecated

This PR attempts to fix the problem.


Seldaek:
Oh interesting, I assumed null was the most efficient way to get "now" without triggering string parsing for dates, but I guess not :)

Thanks

PR #1557, opened 2021-05-28
Add Record/Level/LevelName type aliases and improve phpstan type coverage to level 8

Seldaek:
Many new type annotations, a few small code tweaks/fixes 👍🏻

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -1,5 +1,5 @@
 parameters:
-    level: 6
+    level: 8
 
     treatPhpDocTypesAsCertain: false
     reportUnmatchedIgnoredErrors: false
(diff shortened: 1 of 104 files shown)


holtkamp:
🚀 , @Seldaek minor casing-consistency-thingy not sure it is worth submitting an issue for: both `@inheritDoc` and `@inheritdoc` are used in PHPDoc. https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5141 indicates that `@inheritDoc` is the "correct" approach...

Seldaek:
@holtkamp ok, not entirely sure if that is still accurate as fixing it didn't seem to change anything (I'd have expected something to break at least). But anyway it's fixed, thanks for pointing it out.

Issue #1459, opened 2020-05-05
Set date format in Logger


delrio-design:
Some background: I am using Fluent Bit as a log forwarder for a containerized PHP application with Monolog, and want to standardize date formats across my log streams.

Some formatters, such as LineFormatter and NormalizeFormatter allow this already. I am using JsonFormatter for structured logs, it inherits from NormalizeFormatter, but I don't believe I can set the date format without subclassing it and overriding the normalize method myself. I would like to be able to set the date format globally or consistently per formatter. Given that some streams require certain date formats for ingestion by third party services, it would be nice if each step from logger to handler to stream inherited the default or global date format unless explicitly set.

Seldaek:
Indeed JsonFormatter relies on jsonSerialize's implementation on `Monolog\DateTimeImmutable`. 

I think fixing it to allow choosing a date format sounds possible and reasonable, but I am not sure about adding a default global date format option, that sounds like it's going to be a mess to handle.

delrio-design:
> Indeed JsonFormatter relies on jsonSerialize's implementation on `Monolog\DateTimeImmutable`.
> 
> I think fixing it to allow choosing a date format sounds possible and reasonable, but I am not sure about adding a default global date format option, that sounds like it's going to be a mess to handle.

Thanks for the consideration! You're right about the global format, that's a broad request.

Seldaek:
If you feel like sending a PR for adding this to JsonFormatter please do :)

Shiva-Kandagatla:
I'm also stuck on the exact same problem. @delrio-design Can you please explain a bit more how you solved it.

PR #1561, opened 2021-07-01
JsonFormatter::normalize() : respect date format from Formatter

thomasphansen:
Hello! :)

This seems to be a leftover from #1459: JsonFormatter is not respecting the content of FormatterInterface->dateFormat.
The problem is because JsonFormatter extends the `normalize` method, but doesn't handle `DateTimeInterface` objects the same way `parent::normalize()` does. This PR fixes the issue. :)


Seldaek:
Makes sense, thanks

thomasphansen:
Uh, sounds like I wasn't very careful before sending the push request... Sorry for that. I basically copied the lines and pasted on your code, without checking the code style and missing the use clause for DateTimeImmutable... :( Will do it better next time, I promise! :)

Seldaek:
No worries at all, easy enough to fix. The main problem was the test failures but I figured out the issue so now should be good to merge 👍🏻

PR #1563, opened 2021-07-12
Fix Utils::getClass() with PHP8

villfa:
The results of this function differ over PHP versions since from PHP8 anonymous classes names don't necessarily start by _class_.

This PR takes also into account that _get_class_parent()_ may return _false_.

### DIFF
--- a/src/Monolog/Utils.php
+++ b/src/Monolog/Utils.php
@@ -19,7 +19,15 @@ final class Utils
     {
         $class = \get_class($object);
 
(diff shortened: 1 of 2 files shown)


Seldaek:
As per https://3v4l.org/KRpcr it seems like the parent class is anyway already returned in php8, so do we need this code at all or is get_class() not enough on php8+ ?

villfa:
Indeed this function has little value on PHP8+ and could be dropped at the same time than PHP7.

Seldaek:
Ok sorry never mind, I had a closer look now and I see your code definitely makes more sense than including stuff like `class@anonymous/in/0DSEO:27$0`.

Thanks 👍🏻

PR #1564, opened 2021-07-14
Allow psr/log 2.0

Jean85:
`psr/log` 2.0 was just tagged: https://twitter.com/phpfig/status/1415413578191024130

3.0 has been tagged too, but it would require a version bump and dropping support for `psr/log` 1.0, so we'll need to wait for now.

Details are available in the new meta document: https://www.php-fig.org/psr/psr-3/meta/
As always, changes are done in two steps to make a smoother transition path, as per the [PHP-FIG evolution bylaw](https://www.php-fig.org/bylaws/psr-evolution/)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "psr-4": {"Monolog\\": "tests/Monolog"}
     },
     "provide": {
-        "psr/log-implementation": "1.0.0"
+        "psr/log-implementation": "1.0.0 || 2.0.0"
     },
     "extra": {
         "branch-alias": {
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     ],
     "require": {
         "php": ">=7.2",
-        "psr/log": "^1.0.1"
+        "psr/log": "^1.0.1 || ^2.0"
     },
     "require-dev": {
         "aws/aws-sdk-php": "^2.4.9 || ^3.0",


derrabus (approved):
Does the `provide` block need an update as well?

Jean85:
Oh! Good question... I think it does, but I'm not totally sure it works with something like `^1 || ^2`

stof:
The provide rule should indeed be updated to `1.0.0 || 2.0.0`

derrabus (changes requested):
Monolog's test suite uses `Psr\Log\Test\LoggerInterfaceTest` which has been removed in php-fig/log#76.

derrabus:
Should be fixed by #1565.

derrabus:
> 3.0 has been tagged too, but it would require a version bump and dropping support for `psr/log` 1.0

No, it wouldn't. All implementations of the `psr/log` interfaces that this repository provides already have the necessary return types that are required for 3.0 compatibility. I've run the test suite locally with `psr/log` 3.0 and it passes (with one error related to an implementation from a different package).

Jean85:
@derrabus oh you're right, it's because Monolog already did the major bump to add return types! Good!

Jean85:
PHPStan failure seems unrelated... Rebased onto `main` hoping it would fix the CI.

Seldaek:
The CI works because psr/log 2.0 isn't used due to transient dependencies:

```
$ c why-not psr/log 2.0.0
elasticsearch/elasticsearch  v7.13.1  requires  psr/log (~1.0)
graylog2/gelf-php            1.7.0    requires  psr/log (~1.0)
rollbar/rollbar              v1.8.0   requires  psr/log (^1.0.1)
ruflin/elastica              5.1.0    requires  psr/log (~1.0)
```

But technically yeah this will break the tests once 2.0 is used, so will wait until #1565 is mergeable.

Anyway if you want to help move the ecosystem forward, see the list of packages above :D

Jean85:
Done (where missing)!

Let's hope for a quick turnaround...

ezimuel:
@Seldaek I just released `elasticsearch-php`  [7.14.0](https://github.com/elastic/elasticsearch-php/releases/tag/v7.14.0) including the `psr/log v2` support. Thanks to @derrabus for https://github.com/elastic/elasticsearch-php/pull/1154.

derrabus:
> > Would this be able to also allow for `psr/log` 3.0? It was tagged on July 14.
> 
> Not without dropping support for 1.0

Again, I disagree. We've had that discussion already. 🙂

Jean85:
> > Not without dropping support for 1.0
> 
> Again, I disagree. We've had that discussion already. slightly_smiling_face

Sorry but I forgot about the discussion... And while searching for that I found the reason! :smile: all return types in `psr/log` are `void`, so that's a special case that doesn't trigger Liskov incompatibility as in the general case! :tada:

See https://github.com/php-fig/log/pull/77

derrabus:
> all return types in `psr/log` are `void`

… and already implemented by Monolog.

> [EDIT] Still, we can't force it downstream, and we're blocked by them to have a CI that runs all the checks.

At some point, we run into a vicious circle of blockers. Monolog is a key implementation for `psr/log`. If we don't allow the new interfaces here, a large share of the ecosystem is blocked.

If the CI is really a problem, we could add a reduced CI job that only runs core tests and ignores dependencies that are not ready yet.

Seldaek:
Closing as this commit was included in #1587 - thanks everyone!

PR #1565, opened 2021-07-18
Don't use the deprecated LoggerInterfaceTest class

derrabus:
`Psr\Log\Test\LoggerInterfaceTest` is deprecated and has been removed in `psr/log` 2.0. This PR inlines the test cases from that class. That should unlock #1564.

### DIFF
--- a/tests/Monolog/PsrLogCompatTest.php
+++ b/tests/Monolog/PsrLogCompatTest.php
@@ -11,16 +11,130 @@
 
 namespace Monolog;
 
-use Monolog\Handler\TestHandler;
 use Monolog\Formatter\LineFormatter;
+use Monolog\Handler\TestHandler;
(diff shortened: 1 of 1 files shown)


Seldaek:
I think I'd rather wait on https://github.com/php-fig/log-util/pull/2 to be finished so we can rely on that instead of copy pasting it all in here.. I don't think supporting psr/log 2.0 is that urgent?

fabpot:
Not sure if it's urgent to support psr/log 2.0, but as we do support it in Symfony, 2.0 is installed by default and then, doing `composer req symfony/monolog-bundle` will fail. Using `-W` fixes the problem in the meantime.

Seldaek:
Right, log-util seems stalled so I merged #1587 now which does pretty much the same as this PR. Thanks!

PR #1572, opened 2021-07-27
add check for required extension

### LINKED ISSUE #1569: undefined constant AF_INET
Monolog version 2.3.2

I'm trying to write factories for the handlers and got this error:

```shell
1) Mimmi20Test\LoggerFactory\Handler\SyslogUdpHandlerFactoryTest::testInvoceWithConfig
Use of undefined constant AF_INET - assumed 'AF_INET' (this will throw an Error in a future version of PHP)

C:\Users\Besitzer\Documents\GitHub\monolog-laminas-factory\vendor\monolog\monolog\src\Monolog\Handler\SyslogUdp\UdpSocket.php:32
C:\Users\Besitzer\Documents\GitHub\monolog-laminas-factory\vendor\monolog\monolog\src\Monolog\Handler\SyslogUdpHandler.php:62
C:\Users\Besitzer\Documents\GitHub\monolog-laminas-factory\src\Handler\SyslogUdpHandlerFactory.php:94
C:\Users\Besitzer\Documents\GitHub\monolog-laminas-factory\tests\Handler\SyslogUdpHandlerFactoryTest.php:88
```

Which extension is required to avoid this error?

mimmi20:
The UdpSocket used in the SyslogUdpHandler requires the sockets extension. This PR will add a check for this.

Fixes #1569

### DIFF
--- a/src/Monolog/Handler/SyslogUdpHandler.php
+++ b/src/Monolog/Handler/SyslogUdpHandler.php
@@ -49,11 +49,16 @@ class SyslogUdpHandler extends AbstractSyslogHandler
      * @param bool       $bubble   Whether the messages that are handled can bubble up the stack or not
      * @param string     $ident    Program name or tag for each log message.
      * @param int        $rfc      RFC to format the message for.
+     * @throws MissingExtensionException
      *
      * @phpstan-param self::RFC* $rfc
      */
     public function __construct(string $host, int $port = 514, $facility = LOG_USER, $level = Logger::DEBUG, bool $bubble = true, string $ident = 'php', int $rfc = self::RFC5424)
     {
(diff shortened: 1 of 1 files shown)


juan-morales:
I believe that this change should determine the overall behavior in the package.

What I mean is ... if we choose to throw an exception if requirements are not satisfied, then we would have to replicate this behavior in every handler used, so developers can expected this.

Other option would be to silently disable the handler is requirements are not satisfied. 

I am not quite sure about the right approach, and I don't have time at the moment to take a look at the rest of the handlers to see this.

I wonder what the rest thinks about this.

@Seldaek @gmponos @stof

juan-morales (commented):
:+1:

Seldaek:
Thanks

PR #1575, opened 2021-07-28
Documentation. Enrich documentation on "Customizing log format".

### LINKED ISSUE #889: Formatter Syntax (Documentation Not Available)
There's no obvious cheat sheet for what substitutions one can make for the Formatters (I'm specifically thinking of the "LineFormatter". I've tried to follow the code backwards via inheritance and can see that it extends the NormalizeFormatter but its not clear to me how or where the substitutions are made.

If it were, I'd have proposed a Pull Request for such documentation.

1. It's probably somewhere really obvious - but if it is can someone point me to it (and possibly make it more obvious)?
2. If it's really obvious, as in the code where the substitutions are made is obvious, could someone point me to the code so I can read/understand and make a Pull Request to help out?

If it's not really obvious, how might we better document what one can put into the format string for the LineFormatter and possibly other relevant loggers?

I'm using the latest version of Monolog from composer (composer update monolog/monolog) and looking in the latest MASTER on this GitHub.

(I'd add this to the documentation label but I'm not sure how to do that either :()

juan-morales:
Fixes #889 

The idea behind this pull request is to put a little bit more light and understanding on the "Customizing log format" section of the help.

Not only mentioned in the issue referenced here but also happened to some colleges too.

Hopefully is good enough.

Ideas and suggestions are welcome too.

### DIFF
--- a/doc/01-usage.md
+++ b/doc/01-usage.md
@@ -194,31 +194,48 @@ $securityLogger = $logger->withName('security');
 ## Customizing the log format
 
 In Monolog it's easy to customize the format of the logs written into files,
-sockets, mails, databases and other handlers. Most of the handlers use the
+sockets, mails, databases and other handlers; by the use of "Formatters". 
 
+As mentioned before, a *Formatter* is attached to a *Handler*, and as a general convention, most of the handlers use the
 ```php
 $record['formatted']
 ```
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks

PR #1576, opened 2021-07-29
Fix - SocketHandler. Set chunkSize default value

### LINKED ISSUE #1570: TypeError: Return value of Monolog\Handler\SocketHandler::getChunkSize()
Monolog version 2.3.2

I'm trying to write factories for the handlers and got this error:

```shell
TypeError: Return value of Monolog\Handler\SocketHandler::getChunkSize() must be of the type int, null returned
```

This is the used code:

```php
$handler = new SocketHandler(
    $connectionString,
    $level,
    $bubble
);

$handler->getChunkSize();
```

juan-morales:
Fixes #1570 

Sets default value of chunkSize to 65023 (a good reasonable number like a UDP datagram size packet).

At the beggining I tried to set it to PHP_INT_MAX (according the documentation in the php website) but I got the following errors while execution the test `testWriteContentWithPlainTextMessage()`:

```

stream_set_chunk_size(): The chunk size cannot be larger than 2147483647
 monolog/src/Monolog/Handler/SocketHandler.php:272
 monolog/src/Monolog/Handler/SocketHandler.php:364
 monolog/src/Monolog/Handler/SocketHandler.php:339
 monolog/src/Monolog/Handler/SocketHandler.php:316
 monolog/src/Monolog/Handler/SocketHandler.php:74
 monolog/src/Monolog/Handler/SlackHandler.php:150
 monolog/src/Monolog/Handler/AbstractProcessingHandler.php:48
 monolog/tests/Monolog/Handler/SlackHandlerTest.php:108
 ```
I got it with php 7.4

So according the error , the number candidate `2147483647` is the max allowed.

But then I changed my mind and used number `65023`, the same used for UDP packets in the UdpHandler in Monolog.

If another number should be use as default, please let me know and I can change the PR asap!

I also fixed/changed some php-doc related issues in the file, and added a simple unit test for this.

Ideas/Suggestions are welcome.

### DIFF
--- a/src/Monolog/Handler/SocketHandler.php
+++ b/src/Monolog/Handler/SocketHandler.php
@@ -11,7 +11,11 @@
 
 namespace Monolog\Handler;
 
+use InvalidArgumentException;
+use LogicException;
 use Monolog\Logger;
+use RuntimeException;
+use UnexpectedValueException;
 
 /**
  * Stores to any socket - uses fsockopen() or pfsockopen().
@@ -24,6 +28,8 @@ use Monolog\Logger;
  */
 class SocketHandler extends AbstractProcessingHandler
 {
+    /** @const int */
+    const DEFAULT_CHUNK_SIZE = 65023;
     /** @var string */
     private $connectionString;
     /** @var float */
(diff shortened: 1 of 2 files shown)


mimmi20 (approved):

Seldaek:
Thanks but IMO null is a better default, so it uses the default PHP stream chunk size unless explicitly set by a user.

Issue #1577, opened 2021-07-29
bug: OOM errors on stream handler as of Monolog 2.3.0


bshaffer:
Monolog version 2 (only on 2.3.0+)

Error: `PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 2147483647 bytes) in...`

This can be recreated with the following file (note the php ini setting `memory_limit=2048M`):

```php
ini_set('memory_limit', '2048M');

$stream = tmpfile();
new Monolog\Handler\StreamHandler($stream);
stream_get_contents($stream);
```

@dwsupplee tracked the issue down to [this change](https://github.com/Seldaek/monolog/pull/1553/files). When I remove the calls to `stream_set_chunk_size` in the handler, the OOM error disappears.

I don't know the proper solution, if this change should just be reverted, or if we should do something to ensure the memory limit is less than the chunk size... maybe there's some specific handling necessary for `tmpfile`? Either way, this is a pretty nasty error so any help is appreciated!

juan-morales:
I Will create a PR during The morning... to propose a solution for this

juan-morales:
The pull request is created but it cannot pass a phpstan test, and I dont know if is my change or something else.

Any help?

Seldaek:
Should be fixed by #1578

PR #1578, opened 2021-07-30
Fix - OutOfMemory error in StreamHandler

juan-morales:
Fixes Issue #1577 

This change is meant to prevent setting a stream chunk size bigger than the available amount of memory.

The default value is still set, but can get change after calculations based on the php settings and the memory currently used by the php process.

There is a hard limit for an stream size of 2G ... is does not matter if the PHP process gets 1000 G of memory, the stream size cannot be bigger than 2G! and I suspect is because of some C lib used behind the scenes to implement streams in PHP.

Also ... the official PHP documentation states that `stream_set_chunk_size()` can hold a value of PHP_INT_MAX , but this is not completely true, because no matter what, it cannot be bigger than 2G! (because of the reason I already mentioned.

I suspect that if maybe the libs used in PHP are compiled with a 64 bit version... maybe the size can get bigger than 2G ... but just to be safe ... I would prefer to keep it in 2G.

This topic is quite tricky, please, someone elses opinion on this? ... thanks in advance

Ref: [PHP official documentation for stream_set_chunk_size()](https://www.php.net/manual/en/function.stream-set-chunk-size.php)

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -26,9 +26,7 @@ use Monolog\Utils;
 class StreamHandler extends AbstractProcessingHandler
 {
     /** @const int */
-    const SAFE_MEMORY_OFFSET = 1024;
-    /** @const int */
-    const MAX_CHUNK_SIZE = 2147483647;
+    protected const MAX_CHUNK_SIZE = 100 * 1024 * 1024;
     /** @var int */
     protected $streamChunkSize = self::MAX_CHUNK_SIZE;
     /** @var resource|null */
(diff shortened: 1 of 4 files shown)


juan-morales:
@Seldaek sorry to bother you. Can you help me with the phpstan errors I get in this PR?

Seldaek:
Feel free to ignore this for now, I need to find time to review the rest of the PR first, then will happily help with that.

juan-morales:
> Feel free to ignore this for now, I need to find time to review the rest of the PR first, then will happily help with that.

Thanks for Your quick response.

juan-morales:
@Seldaek your last change fixed the problem I had on this pull request during the automatic checks. Thanks.

src/Monolog/Handler/StreamHandler.php
```diff
@@ -50,9 +54,26 @@ class StreamHandler extends AbstractProcessingHandler
...
             $this->stream = $stream;
-            stream_set_chunk_size($this->stream, self::MAX_CHUNK_SIZE);
+
+            try {
+                stream_set_chunk_size($this->stream, $this->streamChunkSize);
+            } catch (\Exception $exception) {
+                throw new \RuntimeException('Impossible to set the stream chunk size.'
+                .PHP_EOL.'Error: '.$exception->getMessage()
+                .PHP_EOL.'Trace: '.$exception->getTraceAsString());
+            }
```
Seldaek, line 76: Why rethrowing here and not simply letting the original exception do its thing?

Seldaek: I'll revert this change for now as I don't understand it and I'd like to get the rest merged, if you think it's important please explain and we can do it in another PR.

tests/Monolog/Handler/StreamHandlerTest.php
```diff
@@ -220,4 +221,90 @@ public function provideNonExistingAndNotCreatablePath()
...
+        $exceptionRaised = false;
+
+        try {
+            $handler = new StreamHandler($stream);
+            stream_get_contents($stream, 1024);
+        } catch (\RuntimeException $exception) {
+            /*
+             * At this point, stream_set_chunk_size() failed in the constructor.
+             * Probably because not enough memory.
+             * The rest of th test depends on the success pf stream_set_chunk_size(), that is why
```
stof, line 278: several typos in that sentence

Seldaek:
cc @dsch this reduces the chunk size again, but IMO it'll still be large enough in most cases to fit in enough data to make writes atomic

Thanks @juan-morales for the PR! I tweaked things a bit to make it all simpler and not dependent on the current memory usage, will let you take a look at my commit.

PR #1580, opened 2021-08-06
Removing mention of monolog-cascade

pluc:
monolog-cascade doesn't support monolog ^2.0 or php8

see related issues on theorchard/monolog-cascade:

https://github.com/theorchard/monolog-cascade/issues/90

https://github.com/theorchard/monolog-cascade/issues/102


Seldaek:
Thanks

PR #1589, opened 2021-09-14
Allow psr/log 3

derrabus:
#1587 enabled psr/log 2, but let's not stop there. 🚀

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -51,20 +51,12 @@ jobs:
           key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.json') }}
           restore-keys: ${{ runner.os }}-composer-
 
-      - name: Remove elasticsearch/elasticsearch on PHP 8
-        if: "startsWith(matrix.php-version, '8.')"
-        run: 'composer remove elasticsearch/elasticsearch --dev --no-update'
-
       - name: Add require for mongodb/mongodb to make tests runnable
         run: 'composer require ${{ env.COMPOSER_FLAGS }} mongodb/mongodb --dev --no-update'
 
       - name: "Handle lowest dependencies update"
         if: "contains(matrix.dependencies, 'lowest')"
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -73,9 +71,11 @@ jobs:
...
 
-      - name: "Run tests with rollbar"
+      - name: "Run tests with rollbar, elastica, gelf"
         run: |
-          composer require psr/log:'^1.1|^2' --no-update
-          composer require rollbar/rollbar:^1.3 --no-update
+          composer require --no-update psr/log:'^1.1|^2|^3'
+          composer require --no-update --dev rollbar/rollbar:^1.3
+          composer require --no-update --dev 'ruflin/elastica:>=0.90,<8'
+          composer require --no-update --dev graylog2/gelf-php:^1.4.2
```
stof, line 79: instead of having to sync those constraints with the `require-dev` in the package.json, it would make more sense to rely on jobs using the unaltered composer.json IMO.

derrabus: So, I would basically swap the jobs: first run the tests with all deps and then just remove those that block psr/log 3?

derrabus: Done.

nicolas-grekas (approved):
Nice one thanks!

stof (approved):

fabpot (approved):
Got the problem today while trying to "run" the Symfony book on Symfony 6.0.

Seldaek:
Thanks, I'll make another release soon

Seldaek:
Did some more interface upgrades (in phpdoc for now) in https://github.com/Seldaek/monolog/commit/437e7a1c50044b92773b361af77620efb76fff59 - I hope that makes sense.

Seldaek:
2.3.4 is out if you want to try again @fabpot

fabpot:
Perfect! Thank you.

PR #1590, opened 2021-09-14
Enable PHP 8.1 in CI

derrabus:


derrabus:
Seems like MongoDB is also not ready for 8.1 yet. Nevermind.

PR #1591, opened 2021-09-15
Allow psr/log 3.0

nicolas-grekas:
I should have spotted this was possible when working on #1587...


tests/Monolog/Handler/RollbarHandlerTest.php
```diff
@@ -39,6 +37,10 @@ class RollbarHandlerTest extends TestCase
 
     protected function setUp(): void
     {
+        if (!class_exists(RollbarLogger::class)) {
+            $this->markTestSkipped("rollbar/rollbar not installed");
```
nicolas-grekas, line 41: for consistency with existing similar checks

Seldaek:
Please see #1589 too

nicolas-grekas:
Oh nice, closing then!

PR #1593, opened 2021-09-17
Allow easier extend of BrowserConsoleHandler.php

Warxcell:
This will allow to get rid of reimplementation of wheel here https://github.com/symfony/symfony/pull/43065/files#diff-b1c0588c5bb7078514aaa70581e920b852b49d9614f687950b6be6495afa7693R29-R42

### DIFF
--- a/src/Monolog/Handler/BrowserConsoleHandler.php
+++ b/src/Monolog/Handler/BrowserConsoleHandler.php
@@ -11,9 +11,16 @@
 
 namespace Monolog\Handler;
 
-use Monolog\Formatter\LineFormatter;
 use Monolog\Formatter\FormatterInterface;
+use Monolog\Formatter\LineFormatter;
 use Monolog\Utils;
+
+use function count;
+use function headers_list;
+use function stripos;
+use function trigger_error;
+
+use const E_USER_DEPRECATED;
 
 /**
  * Handler sending logs to browser's javascript console with no browser extension required
@@ -28,6 +35,9 @@ class BrowserConsoleHandler extends AbstractProcessingHandler
     protected static $initialized = false;
     /** @var FormattedRecord[] */
     protected static $records = [];
+
+    public const FORMAT_HTML = 'html';
+    public const FORMAT_JS = 'js';
 
     /**
      * {@inheritDoc}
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/BrowserConsoleHandler.php
```diff
@@ -131,21 +138,27 @@ protected static function getResponseFormat(): string
...
 
         return 'html';
     }
 
+    protected static function getResponseFormatFromContentType(string $contentType): ?string
+    {
+        // This handler only works with HTML and javascript outputs
+        // text/javascript is obsolete in favour of application/javascript, but still used
+        if (stripos($contentType, 'application/javascript') !== false || stripos($contentType, 'text/javascript') !== false) {
+            return static::FORMAT_JS;
```
stof, line 153: why not `self::` ?

Warxcell: https://github.com/Seldaek/monolog/pull/1160

Seldaek: Yeah on the other hand you used self:: below at line 166. I don't particularly think it matters as nobody should change/override these constants, but it'd be good to at least be consistent about it :)

src/Monolog/Handler/BrowserConsoleHandler.php
```diff
@@ -131,21 +145,27 @@ protected static function getResponseFormat(): string
...
+        }
+        if (stripos($contentType, 'text/html') === false) {
+            return null;
```
Seldaek, line 163: I do question the value of doing this change to `null`.. why not keep `'unknown'`, so it always returns a string? Right now getResponseFormat's return value is broken for example as you didn't update it.

I don't really see the value in deprecating this, except causing more fragmentation in implementations, and added work.

src/Monolog/Handler/BrowserConsoleHandler.php
```diff
@@ -29,6 +36,9 @@ class BrowserConsoleHandler extends AbstractProcessingHandler
     /** @var FormattedRecord[] */
     protected static $records = [];
 
+    public const FORMAT_HTML = 'html';
+    public const FORMAT_JS = 'js';
```
Seldaek, line 40:
```suggestion
    protected const FORMAT_HTML = 'html';
    protected const FORMAT_JS = 'js';
```

I don't think there's value in this being public

Seldaek:
Thanks

PR #1595, opened 2021-09-20
Allow php-amqplib/php-amqplib v3

siwinski:
`php-amqplib/php-amqplib` has released v3 https://github.com/php-amqplib/php-amqplib/releases


Seldaek:
Thanks

PR #1599, opened 2021-09-25
feat: Syslog formatter

### LINKED ISSUE #1594: Syslog formatter
First of, I know the Syslog handler exists.

The suggestion here is to create a Syslog formatter, which would allow using it without messing around with PHP's settings.

Use case:
- formatter is created and Symfony enables it by default in dev mode
- the format of the log file is by default recognized by various logging helpers like https://lnav.org/ without any configuration, super-charging the logging information
- the user (developer) gets great DX by default, using standard best-of-kind apps

Syslog seems like the most obvious choice since it's very widely supported, if there's another alternative, that also works.

dkarlovi:
Closes #1594.

### DIFF
--- a/src/Monolog/Formatter/SyslogFormatter.php
+++ b/src/Monolog/Formatter/SyslogFormatter.php
@@ -1,0 +1,63 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Formatter;
+
+use Monolog\Logger;
+
+/**
+ * Serializes a log message according to RFC 5424
+ *
+ * @author Dalibor Karlović <dalibor.karlovic@sigwin.hr>
+ */
+class SyslogFormatter extends LineFormatter
+{
+    private const SYSLOG_FACILITY_USER = 1;
+    private const LEVELS = [
+        Logger::EMERGENCY => 0,
+        Logger::ALERT => 1,
+        Logger::CRITICAL => 2,
+        Logger::ERROR => 3,
+        Logger::WARNING => 4,
+        Logger::NOTICE => 5,
+        Logger::INFO => 6,
+        Logger::DEBUG => 7,
+    ];
+    // [%datetime%] %channel%.%level_name%: %message% %context% %extra%
+    private const FORMAT = '<%extra.priority%>1 %datetime% %extra.hostname% %channel% %extra.procid% %channel% %extra.structured-data% %message%';
+    private const NILVALUE = '-';
+
+    public function __construct()
+    {
+        parent::__construct(self::FORMAT, 'Y-m-d\TH:i:s.uP');
+
+        $this->extra = [
+            'hostname' => (string) gethostname(),
+            'procid' => (int) getmypid(),
+        ];
+    }
+
+    public function format(array $record): string
+    {
+        $extra = $this->extra;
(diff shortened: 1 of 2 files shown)


Seldaek:
So the point is to output a fake, syslog-style log, to a non-syslog file, right? Seems reasonable I guess, although I'd question why not simply logging to syslog then, but I'm sure you have reasons.

src/Monolog/Formatter/SyslogFormatter.php
```diff
@@ -0,0 +1,68 @@
...
+        return (self::SYSLOG_FACILITY_USER * 8) + self::LEVELS[$level];
+    }
+
+    private function formatExtra(array $record): array
+    {
+        $extra = $this->extra;
+        $extra['priority'] = self::calculatePriority($record['level']);
+        $extra['structured-data'] = self::NILVALUE;
+        
+        return $extra;
```
Seldaek, line 66: Is this getting rid of the original record's extra data on purpose? I'd be better to array_merge it than replace it entirely IMO.

Really I guess this should use a nested LineFormatter to format the record, and use the formatted record as %message% when formatting to syslog format.

dkarlovi:
> although I'd question why not simply logging to syslog then

The general idea behind this is to allow flexibility of a stream handler (arbitrary file(s) as targets, for one) without much fuss, but getting a format which is common enough to get recognized by other related tooling without needing special configuration. See #1594 for one use case which makes sense to me.

src/Monolog/Formatter/SyslogFormatter.php
dkarlovi: The target format is very specific about the details, no arbitrary extras can be added or it will no longer be Syslog compatible. Still need to format/add the contextual data (param just before message).

> I guess this should use a nested LineFormatter to format the record, and use the formatted record as %message% when formatting to syslog format.

Not sure what you mean here, it does use the LineFormatter, no?

Seldaek: I mean when you use a SyslogHandler it will format a log record with extra/context data using a LineFormatter, resulting in something like `"message {"contextKey": "value"} {"extrakey": "value"}"` in `$record['formatted']`. This whole string then gets passed to syslog in https://github.com/Seldaek/monolog/blob/fd4380d6fc37626e2f799f29d91195040137eba9/src/Monolog/Handler/SyslogHandler.php#L65

While with your formatter here, everything gets lost except for the message. If you created a LineFormatter internally to format the record, and then decorate it with a syslog format, you should be able to keep the same level of functionality.

dkarlovi: Syslog natively supports structured data, this is still pending to get implemented here
https://datatracker.ietf.org/doc/html/rfc5424#section-6.3

Seldaek: Ok, I mean if you want to implement it using that it sounds ok to me too, although the ASCII restriction seems a bit outdated.. I don't really mind how the data makes it in, but IMO context/extra should not disappear.

ro0NL:
friendly ping :)

image


IIUC this pr improves that output?

dkarlovi:
@ro0NL it does, I've used https://lnav.org/ when using this with Symfony's dev log file and it's **really** nice already, you should be able to copy/paste this class into your app and register as custom log formatter to see it in action: you can navigate by type of error, 

IIRC I need to support metadata encoding, it shouldn't be all that much work, but I don't have the time right now.

Seldaek:
@dkarlovi sorry what's the state here? Just curious as you marked as draft, and I'm preparing slowly a 2.4.0 release.

dkarlovi:
Closing in favor of #1689.

PR #1600, opened 2021-10-01
Add constructor injection for `SocketHandler` and its children

boesing:
Hey there,

I'd like to propose a PoC to allow constructor injection for the `SocketHandler` and its children.
Is that something you want to support within monolog?

If not, I'd be fine with closing this. I can wrap around this and implement this as a feature in a dedicated abstraction layer tho.

Thanks in advance,
Max

### DIFF
--- a/src/Monolog/Handler/FleepHookHandler.php
+++ b/src/Monolog/Handler/FleepHookHandler.php
@@ -45,8 +45,16 @@ class FleepHookHandler extends SocketHandler
      * @param  string                    $token  Webhook token
      * @throws MissingExtensionException
      */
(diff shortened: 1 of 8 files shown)


Seldaek:
I guess why not, although it makes the signatures much longer and complex to grasp, I do see the value/ease of use too.

Can you please update the phpdoc to add the 5 new params in SocketHandler tho? No need to duplicate the descriptions everywhere IMO, as these get inherited, but it'd be nice to have details on the parent class.

boesing:
I've added the new arguments to phpdoc 🤙🏼

PR #1601, opened 2021-10-02
Fixed ElasticsearchHandler.php throws associative context map to text error

mohammad1444:
Using associative array causes this error:
Elasticsearch\Common\Exceptions\RuntimeException: mapper_parsing_exception: failed to parse field [context] of type [text] in document with id vendor\monolog\monolog\src\Monolog\Handler\ElasticsearchHandler.php on line 198

For resolving this error I've added a Json encode in ElasticsearchFormatter class to convert associative contexts to Json string

### DIFF
--- a/src/Monolog/Formatter/ElasticsearchFormatter.php
+++ b/src/Monolog/Formatter/ElasticsearchFormatter.php
@@ -1,4 +1,6 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
 
 /*
  * This file is part of the Monolog package.
@@ -12,6 +14,7 @@
 namespace Monolog\Formatter;
 
 use DateTimeInterface;
+use Monolog\Utils;
 
 /**
  * Format a log message into an Elasticsearch record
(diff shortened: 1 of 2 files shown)


Seldaek:
I kinda have a hard time believing this is a problem and nobody else noticed yet that associative context keys cause issues.. I'm wondering if you're doing something else that is causing this.

mohammad1444:
OK, I will add an issue to show how to reproduce the error.
TNX

stof:
I suspect that the issue is in your ElasticSearch index mapping. The error message says that the `context` field has the type `text`. I guess that everyone else not having the issue actually defines the right type for this (the `context` key is definitely not `text`)

mohammad1444:
Context is created by default with Filebeats in Elasticsearch and it's default type is text, I've tried to change the type, but the type of context is disabled for edit.
image

stof:
Well, to me the issue is using an index managed by Filebeat when you don't use Filebeat to ingest the data but the Monolog processor...

Anyway, the formatter is configurable in the handler, so you could still inject a custom formatter doing that json encoding in your project, without breaking things for everyone else.

mohammad1444:
OK, Is there any other tools suggested to use instead of Filebeats that is compatible with monolog?
Sorry for asking :)

stof:
Well, the ElasticsearchHandler uses the ElasticSearch API directly to put documents. So the tool to use is ElasticSearch itself.

PR #1606, opened 2021-10-27
Enhance SocketHandler::isConnected method

juan-morales:
I need someone else review oh this code.

I took it from [https://www.php.net/manual/en/function.stream-set-timeout.php#118619](url)

Thanks in advance

### DIFF
--- a/src/Monolog/Handler/SocketHandler.php
+++ b/src/Monolog/Handler/SocketHandler.php
@@ -250,7 +250,7 @@ class SocketHandler extends AbstractProcessingHandler
 
             $info = stream_get_meta_data($this->resource);
 
-            if ($info['timed_out']) {
+            if (isset($info['timed_out']) && $info['timed_out']) {
                 return false;
             }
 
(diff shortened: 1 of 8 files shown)


tests/Monolog/Handler/FlowdockHandlerTest.php
```diff
@@ -62,7 +62,7 @@ public function testWriteContent($content)
     private function createHandler($token = 'myToken')
     {
         $constructorArgs = [$token, Logger::DEBUG];
-        $this->res = fopen('php://memory', 'a');
+        $this->res = fopen('php://temp', 'a');
```
Seldaek, line 65: Why temp instead of memory? That writes to disk which does not seem necessary?

src/Monolog/Handler/SocketHandler.php
```diff
@@ -236,8 +236,31 @@ public function getChunkSize(): ?int
...
+
+        if (feof($this->resource)) {
+            return false;
+        }
+
+        $info = stream_get_meta_data($this->resource);
+
+        if (isset($info['timed_out']) && $info['timed_out']) {
+            return false;
+        }
```
Seldaek, line 254: This check is already covered below when writing IMO so not sure this is needed.

src/Monolog/Handler/SocketHandler.php
```diff
@@ -236,8 +236,31 @@ public function getChunkSize(): ?int
...
+
+        $rStreams = [$this->resource];
+        $wStreams = null;
+
+        if (!stream_select($rStreams, $wStreams, $wStreams, (int)$this->getTimeout())) {
+            return false;
+        }
```
Seldaek, line 261: I don't think this adds any value, it'll block for $timeout if there is no event to be read from the streams, but it doesn't necessarily mean it's timed out IMO if nothing happened on the stream.

PR #1607, opened 2021-11-02
Use DateTimeImmutable by proper namespace

curry684:
With PHP 8.1 this results in:
```
Attempted to load class "DateTimeImmutable" from namespace "Monolog".
Did you forget a "use" statement for another namespace?
```

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -307,7 +307,7 @@ class Logger implements LoggerInterface, ResettableInterface
                     'level' => $level,
                     'level_name' => $levelName,
                     'channel' => $this->name,
-                    'datetime' => new DateTimeImmutable($this->microsecondTimestamps, $this->timezone),
+                    'datetime' => new \DateTimeImmutable($this->microsecondTimestamps, $this->timezone),
                     'extra' => [],
                 ];
 


src/Monolog/Logger.php
```diff
@@ -307,7 +307,7 @@ public function addRecord(int $level, string $message, array $context = []): boo
                     'level' => $level,
                     'level_name' => $levelName,
                     'channel' => $this->name,
-                    'datetime' => new DateTimeImmutable($this->microsecondTimestamps, $this->timezone),
+                    'datetime' => new \DateTimeImmutable($this->microsecondTimestamps, $this->timezone),
```
juan-morales, line 310: according to :

src/Monolog/Test/TestCase.php

it should be **Monolog\DateTimeImmutable**

juan-morales (changes requested):
maybe adding the dependency at the top with the **use** keyword is what we need

src/Monolog/Logger.php
juan-morales, line 310: we need the usage of **Monolog\DateTimeImmutable** instead

juan-morales: we need the usage of **Monolog\DateTimeImmutable** instead

juan-morales (changes requested):

juan-morales:
According to this file:

src/Monolog/Test/TestCase.php

The right one is: **Monolog\DateTimeImmutable**

juan-morales (changes requested):
Please use the right dependency

juan-morales:
sorry for all the comments, I got errors while doing the review and now I cannot delete duplicates.

curry684:
Oh my you're completely right it never occurred to me there would be a local copy extending the global one. 

I'll dive into why the hell I'm getting that error then.

juan-morales:
you can try adding at the top

**use Monolog\DateTimeImmutable** and leave DateTimeImmutable without the "\" symbol. Let me know if I can help you out on this. Welcome to help

temp:
Same problem here, but with docker. No caching occurs, still got the error in some projects. Error goes away if you require Monolog/DateTimeImmutable in index.php, but this is only a temporary quick fix.
We really have no clue what's the problem here, autoloading via composer looks fine.

Seldaek:
See https://github.com/Seldaek/monolog/issues/1626 and https://github.com/php/php-src/issues/8164  just in case you haven't yet.

PR #1609, opened 2021-11-16
return self on builder methods for more consistent behaviour

codemasher:
Just an example i just ran into:

```php
$formatter = (new LineFormatter)
    ->setJsonPrettyPrint(true) // -> LineFormatter instance
    ->addJsonEncodeOption(JSON_UNESCAPED_SLASHES) // -> void
    ->addJsonEncodeOption(JSON_UNESCAPED_UNICODE); // boo!
```

I didn't go through the handler classes to check for similar cases.

### DIFF
--- a/src/Monolog/Formatter/JsonFormatter.php
+++ b/src/Monolog/Formatter/JsonFormatter.php
@@ -109,11 +109,13 @@ class JsonFormatter extends NormalizerFormatter
     }
 
     /**
-     * @return void
+     * @return self
      */
-    public function includeStacktraces(bool $include = true)
+    public function includeStacktraces(bool $include = true): self
     {
         $this->includeStacktraces = $include;
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


juan-morales:
In my own opinion, applying this pattern is nice, and gives us the possibility to write more readable code. Despite that, I believe that you should apply this pattern everywhere, at least all default formatters, to have a uniform code structure.

For example:

`src/Monolog/Formatter/LineFormatter.php` , is not on this change, and imo it should.

Can you please check a bit more in the code in order to apply uniformly this pattern wherever it needs to b applied? I also like this approach/pattern/style by the way.

juan-morales (changes requested):
imo, there is/are missing place/s  as I pointed in my comment.

codemasher:
You're right, I've overlooked this (and probably more). Also, i was hesitant to change `Logger::reset()` and maybe even `Logger::close()`. I'll recheck when i got more time, also feel free to point out what I may have missed. :)

(Personally i despise public setter and builder methods that return `void` for no apparent reason and while this may [be opinionated](https://github.com/nikic/FastRoute/pull/181) (like almost everything in programming), maybe just let people have nice things?)

Seldaek:
Thanks

PR #1612, opened 2021-11-24
IntrospectionProcessor returns also information about call type

jszczypk:
$record['extra']['callType'] will contain current call type: "->" or "::" depending if it was normal or static method call.
Gives possibility to nicely format messages "{$record['extra']['class']}{$record['extra']['callType']}{$record['extra']['function']} ..."


Seldaek:
Scheduling for 3.0 because adding this here seems like it could potentially break stuff.. Some log backends like Elasticsearch kinda depend on the shape of logs and if things change like this it may cause issues, so if it's only about nicer formatting but doesn't bring any other value it doesn't seem really worth the risk in a minor release.

jszczypk:
@Seldaek It is impossible for this patch to break anything - it's just additional field in extra, so Elasticsearch and others will continue to work as they work now.
Other projects will have the possibility to start using this feature from new minor version.
You can safely move it to 2.5.0

Seldaek:
Ah right I guess the problem was with keys of variable types, not new keys

PR #1613, opened 2021-11-24
Added user_agent to the WebProcessor

alexjose:
Adding `user_agent` in the `WebProcessor` will be a good to have data point.


Seldaek:
Tweaked to avoid adding it by default, so you would have to select it specifically in the list of extraFields, as it's quite long and isn't necessarily useful to all.

PR #1614, opened 2021-11-25
Fix ErrorHandler::handleError signature: make $context nullable

koenkivits:
Even though the PHP docs don't mention it, the error context can be null. We just experienced this with a failing `getaddrinfo` call from an internal PHP function, which triggered a warning which in turn crashed the Monolog error handler.

A more in-depth explanation for this fix can be found in the Sentry repo, where the same problem was fixed a while back: https://github.com/getsentry/sentry-php/issues/912#issuecomment-548580147

Thanks in advance!


Seldaek:
Thanks

PR #1616, opened 2021-11-26
Handling the maximum length of a Telegram message

evgeek:
The maximum allowed length of a Telegram message - 4096 symbols (['text' field](https://core.telegram.org/bots/api#message)). 
Longer messages are not delivered.

Therefore, I implemented truncating too long message to the maximum length allowed as the default behavior. 
An additional option that you can enable by `splitLongMessages` is  splitting too long message into several. Also, using `delayBetweenMessages` you can enable a 1 second delay between sending a split message to avoid the _429 Too Many Requests_ error (according to the recommendations - [My bot is hitting limits, how do I avoid this?](https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this))

### DIFF
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -105,14 +105,15 @@ class TelegramBotHandler extends AbstractProcessingHandler
     public function __construct(
         string $apiKey,
         string $channel,
-        $level = Logger::DEBUG,
-        bool $bubble = true,
+               $level = Logger::DEBUG,
+        bool   $bubble = true,
         string $parseMode = null,
-        bool $disableWebPagePreview = null,
-        bool $disableNotification = null,
-        bool $splitLongMessages = false,
-        bool $delayBetweenMessages = false
-    ) {
+        bool   $disableWebPagePreview = null,
+        bool   $disableNotification = null,
+        bool   $splitLongMessages = false,
+        bool   $delayBetweenMessages = false
+    )
+    {
         if (!extension_loaded('curl')) {
             throw new MissingExtensionException('The curl extension is needed to use the TelegramBotHandler');
         }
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -167,6 +187,16 @@ protected function write(array $record): void
      * @param string $message
      */
     protected function send(string $message): void
+    {
+        $messages = $this->handleMessageLength($message);
+
+        foreach ($messages as $key => $message) {
+            $key > 0 && sleep(1);
```
stof, line 194: sleeping inside a logger looks like a bad default behavior to me.

src/Monolog/Handler/TelegramBotHandler.php
```diff
@@ -77,6 +82,12 @@ class TelegramBotHandler extends AbstractProcessingHandler
      */
     private $disableNotification;
 
+    /**
+     * Truncate message longer than MAX_MESSAGE_LENGTH
+     * @var ?bool
```
stof, line 87: why allowing `null` to do the same than `false` ?

src/Monolog/Handler/TelegramBotHandler.php
evgeek: Is it better to remove the delay in the hope that throttling will not have time to work (at least 5-10 messages will pass)? Or change the default behavior to truncate message?

src/Monolog/Handler/TelegramBotHandler.php
evgeek: Made by analogy with other parameters (`$disableWebPagePreview`, `$disableNotification`). I can fix if this is not the best decision

src/Monolog/Handler/TelegramBotHandler.php
evgeek: Fixed: changed the default behavior to truncate long message; for a split message delay can be enabled with a additional parameter

src/Monolog/Handler/TelegramBotHandler.php
evgeek: Fixed

Seldaek:
Thanks

PR #1620, opened 2021-12-03
Store native BSON types as-is in the MongoDB formatter

ellotheth:
I like to throw MongoDB document IDs in my log context so I can cross-reference them later:

```php
$log->info("{$user->email} logged in", [ 'actor' => $user->id ]);
```

 Right now, `\Monolog\Formatter\MongoDBFormatter` flattens ObjectIds into arrays (e.g. `[ 'oid' => 'object id string', 'class' => ObjectId::class ]`). This PR allows all `\MongoDB\BSON\Type` implementations to pass through the formatter without additional processing.

### DIFF
--- a/src/Monolog/Formatter/MongoDBFormatter.php
+++ b/src/Monolog/Formatter/MongoDBFormatter.php
@@ -86,7 +86,7 @@ class MongoDBFormatter implements FormatterInterface
                 $array[$name] = $this->formatException($value, $nestingLevel + 1);
             } elseif (is_array($value)) {
                 $array[$name] = $this->formatArray($value, $nestingLevel + 1);
-            } elseif (is_object($value) && !($value instanceof Type)) {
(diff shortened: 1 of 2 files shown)


src/Monolog/Formatter/MongoDBFormatter.php
```diff
@@ -85,7 +86,7 @@ protected function formatArray(array $array, int $nestingLevel = 0)
                 $array[$name] = $this->formatException($value, $nestingLevel + 1);
             } elseif (is_array($value)) {
                 $array[$name] = $this->formatArray($value, $nestingLevel + 1);
-            } elseif (is_object($value)) {
+            } elseif (is_object($value) && !($value instanceof Type)) {
```
stof, line 89: parentheses are useless here

Seldaek:
Thanks

PR #1624, opened 2021-12-28
Check whether graylog2/gelf-php is installed  in GelfMessageFormatter constructor

b1rdex:
To make it obvious, the package is required to use the formatter.


Seldaek:
Thanks

PR #1632, opened 2022-02-02
change property visibility to protected

Flightfreak:
Turning the visibility of the properties protected makes the class useable for extension.


mfn:
Or: there are already getters for both of them, but sadly internally they are _not_ used. Which would solve the problem too…

Flightfreak:
You make a valid point, I overlooked the getters. Thanks!

Issue #924, opened 2017-02-16
Prevent infinite recursion in NormalizerFormatter::normalize


alchimik:
something like
``` php
    protected function normalize($data, $depth = 5)
    {
        if ($depth === 0) {
            return 'too deep';
        }
        
        ...

        if (is_array($data)) {
                ...
                $normalized[$key] = $this->normalize($value, $depth - 1);
                ... 
            return $normalized;
        }

        ...
    }
```

alchimik:
``` php 
$parent = [];
$child = [];
$child['parent'] = &$parent;
$parent['child'] = &$child;

function t($parent) {
	// log
}

t($parent);
```

alchimik:
fixed in #808

alchimik:
#808 - Move these changes to the branch 1x, please.

Seldaek:
Just FYI, change of plans, this was backported in #1050

AlanCrevon:
Hi.

I recently upgraded a long overdue projetct using a monolog version < 1.23.0 and this change breaks normalization for some important json data.

Do you think you could make the nesting limit optionnal ?

Thanks/

Seldaek:
@AlanCrevon what kind of depth are you looking at? I can see that 9 is pretty arbitrary, but I do wonder what would be a reasonable number to you?

Anyway perhaps we could add a setter to change the max depth on the NormalizerFormatter.. Can you please open a new issue (or even a PR?)

Geoffrey-R:
Hello,

@Seldaek https://github.com/Seldaek/monolog/pull/1633

PR #1633, opened 2022-02-08
Max depth formatter

Geoffrey-R:
Hello, 

This PR adds a configuration for max serialization depth (https://github.com/Seldaek/monolog/issues/924)

Thanks

### DIFF
--- a/src/Monolog/Formatter/JsonFormatter.php
+++ b/src/Monolog/Formatter/JsonFormatter.php
@@ -29,7 +29,6 @@ class JsonFormatter extends NormalizerFormatter
 
     protected $batchMode;
     protected $appendNewline;
-    protected $maxDepth;
 
     /**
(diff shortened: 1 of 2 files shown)


src/Monolog/Formatter/JsonFormatter.php
```diff
@@ -100,6 +103,22 @@ public function includeStacktraces($include = true)
...
+        return $this->maxDepth;
+    }
+
+    /**
+     * @param int $maxDepth
+     */
+    public function setMaxDepth($maxDepth)
+    {
+        $this->maxDepth = $maxDepth;
+    }
```
Seldaek, line 120: This here and the property declaration above + constructor assignment should probably be reused from NormalizerFormatter as this class extends it.

Seldaek:
Thanks

PR #1635, opened 2022-02-20
[RFC] Monolog 3.x 

Seldaek:
I did some experimentation with https://3v4l.org/Dcj4I#v8.1.3 and it seems workable to add a LogRecord class for 3.0 to replace the `$record` array. This would even allow most handlers to function with both versions of Monolog I believe if they drop the array type hint from the argument (or set it to `array|LogRecord` if they support php8+).

I am not sure if the ArrayAccess compat layer is worth it or not. We could also make a hard break..

This would require PHP 8.1.

The most interesting changes are in the LogRecord and Logger classes.

I'm happy to get people's thoughts/feedback here.


derrabus:
> I am not sure if the ArrayAccess compat layer is worth it or not. We could also make a hard break..

I wouldn't do it. Since the new class won't pass an `array` type declaration anymore, code will have to be adjusted anyway for the new `LogRecord` class. Instead, a `public function toLegacyRecord(): array` (naming tbd) could be added to `LogRecord` to ease the migration path.

```php
class MyHandler extends AbstractHandler
{
    public function handle(LogRecord|array $record): bool
    {
        if ($record instanceof LogRecord) {
            $record = $record->toLegacyRecord();
        }

        // insert 2.x-compatible code here
    }
}
```

composer.json
```diff
@@ -25,12 +25,10 @@
         "php-amqplib/php-amqplib": "~2.4 || ^3",
         "php-console/php-console": "^3.1.3",
         "phpspec/prophecy": "^1.6.1",
-        "phpunit/phpunit": "^8.5",
+        "phpunit/phpunit": "^9",
```
derrabus, line 28:
```suggestion
        "phpunit/phpunit": "^9.5",
```
Running PHPUnit 9.4 on PHP 8.1 is a bad idea. 🙃

src/Monolog/LogRecord.php
```diff
@@ -0,0 +1,122 @@
...
+        'formatted' => true,
+    ];
+
+    /** @var 'DEBUG'|'INFO'|'NOTICE'|'WARNING'|'ERROR'|'CRITICAL'|'ALERT'|'EMERGENCY' */
+    public readonly string $levelName; // TODO enum?
+
+    public function __construct(
+        public readonly \DateTimeImmutable $datetime,
+        public readonly string $channel,
+        /** @var Logger::DEBUG|Logger::INFO|Logger::NOTICE|Logger::WARNING|Logger::ERROR|Logger::CRITICAL|Logger::ALERT|Logger::EMERGENCY */
```
derrabus, line 37:
```suggestion
        /** @var Level */
```

src/Monolog/LogRecord.php
```diff
@@ -0,0 +1,122 @@
...
+        public readonly int $level, // TODO enum?
```
derrabus, line 38: Yes, let's make this en enum while we're at it.

src/Monolog/Handler/TestHandler.php
```diff
@@ -127,7 +125,7 @@ public function hasRecords($level): bool
     }
 
     /**
-     * @param string|array $record Either a message string or an array containing message and optionally context keys that will be checked against all records
+     * @param string|LogRecord $record Either a message string or an array containing message and optionally context keys that will be checked against all records
```
derrabus, line 128:
```suggestion
     * @param string|LogRecord $record Either a message string or a log record object
```

src/Monolog/Test/TestCase.php
```diff
@@ -66,7 +62,7 @@ protected function getIdentityFormatter(): FormatterInterface
         $formatter->expects($this->any())
             ->method('format')
             ->will($this->returnCallback(function ($record) {
-                return $record['message'];
+                return $record->message;
             }));
```
derrabus, line 66:
```php
$this->returnCallback(fn (LogRecord $record) => $record->message)
```

src/Monolog/Handler/ProcessableHandlerTrait.php
```diff
@@ -13,19 +13,18 @@
...
  * @author Jordi Boggiano <j.boggiano@seld.be>
- *
- * @phpstan-import-type Record from \Monolog\Logger
  */
 trait ProcessableHandlerTrait
 {
     /**
      * @var callable[]
-     * @phpstan-var array<ProcessorInterface|callable(Record): Record>
+     * @phpstan-var array<array<(callable(LogRecord): LogRecord)|ProcessorInterface>>
```
nicolas-grekas, line 27: While bumping to PHP 8.1, it would be great to stop using callable and always use \Closure instead. The new $callable(...) syntax makes it a no brainer. Closures are more performant, more secure (a closure cannot be unserialized, saving ourselves from related issues), and easier to check by the engine (think typed properties).

src/Monolog/LogRecord.php
Seldaek: The catch with enum is it means one cannot extend the allowed log levels anymore. On the other hand, extending that is probably a bad idea, and I am not sure if anyone actually does it.

src/Monolog/Handler/ProcessableHandlerTrait.php
Seldaek: Ah interesting, wasn't aware, thanks :)

It still makes it a bit harder to pass in an object with __invoke I guess? But let's give it a shot.

Seldaek:
> I wouldn't do it. Since the new class won't pass an array type declaration anymore, code will have to be adjusted anyway for the new LogRecord class. Instead, a public function toLegacyRecord(): array (naming tbd) could be added to LogRecord to ease the migration path.

For handlers I agree, as they typically consume a record and so they can do as they please, but for processors the current BC layer allows them to transparently transform a record by adding extra data on it and then return it, while if you would return an array in v3 that'd break the rest of the flow.

src/Monolog/Handler/ProcessableHandlerTrait.php
derrabus: > It still makes it a bit harder to pass in an object with __invoke I guess? But let's give it a shot.

No, you can turn an invokable object into a closure like any other callable:

```php
$myClosure = $myInvokableObject(...);
```

Seldaek:
Tons of clean-ups done already..

Still left to do:

- use real PHP 8 property types 
- drop SwiftMailer / replace with symfony/mailer
- update docs
- maybe switch Level to enums
- switch json error handling to use JSON_THROW_ON_ERROR
- switch to Closure instead of callable

PR #1644, opened 2022-03-17
Calculate expected exception message according to PHP version

blueForman:
In PHP versions prior to 8.0 failed was lowercase, after the version 8.0 failed is uppercase


Seldaek:
Haha thanks, I totally missed that, was thinking it was some insane whitespace problem, but now I see it ;)

PR #1650, opened 2022-04-05
Update .gitattributes

angeljqv:
Make the vendor 40kb smaller,
is it possible or is there any reason not to do it?


mfn:
Are you going through random PHP projects?

Same feedback from me as in https://github.com/guzzle/guzzle/pull/3004#issuecomment-1089026008
> As a user, I welcome have those things right in my local code base to check without requiring going somewhere else on the internet.

PR #1654, opened 2022-04-18
Use never return typehint

hemberger:
The `never` return typehint was added in PHP 8.1 and can be used
to indicate that a function will never return.

RFC: https://wiki.php.net/rfc/noreturn_type


Seldaek:
Makes sense but please remove the phpdoc which duplicate it then.

Seldaek:
Thanks!

PR #1655, opened 2022-04-18
Replace switch with match

hemberger:
The `match` expression is available as of PHP 8.0 and provides a more
intuitive interface than `switch` when a single value is returned.

Note that it performs an identity check (===) between the value and
case instead of a weak equality check (==).

https://www.php.net/manual/en/control-structures.match.php


Seldaek:
Thanks

Issue #1648, opened 2022-04-03
Refactorings


Seldaek:
- [x] https://github.com/Seldaek/monolog/pull/1656
- [x] use native PHP 8.1 property types (done in 7952a83e0cd63394638aaeaeeab47d1ea049deb5 and 0dac879)
- [x] drop SwiftMailer / replace with symfony/mailer refs https://github.com/Seldaek/monolog/issues/1621 and https://github.com/Seldaek/monolog/issues/1638
- [x] update docs
- [x] update UPGRADE.md
- [x] ~~switch json error handling to use JSON_THROW_ON_ERROR~~
- [x] switch to Closure instead of callable (except maybe Processors for BC)
- [x] update set_error_handler usage https://www.php.net/manual/en/function.set-error-handler.php#refsect1-function.set-error-handler-changelog

Happy to get help with any of these if someone wants to :)

PR #1656, opened 2022-04-18
Convert level/levelName to enums

Seldaek:
Refs #1648

See Monolog\Level & Monolog\LevelName

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -17,10 +17,8 @@ $finder = PhpCsFixer\Finder::create()
     ->in(__DIR__.'/tests')
 ;
 
-return PhpCsFixer\Config::create()
-    ->setUsingCache(true)
-    ->setRiskyAllowed(true)
-    ->setRules(array(
+$config = new PhpCsFixer\Config();
+return $config->setRules(array(
         '@PSR2' => true,
         // some rules disabled as long as 1.x branch is maintained
         'binary_operator_spaces' => array(
@@ -30,11 +28,11 @@ return PhpCsFixer\Config::create()
         'cast_spaces' => ['space' => 'single'],
         'header_comment' => ['header' => $header],
         'include' => true,
-        'class_attributes_separation' => ['elements' => ['method']],
+        'class_attributes_separation' => array('elements' => array('method' => 'one', 'trait_import' => 'none')),
         'no_blank_lines_after_class_opening' => true,
         'no_blank_lines_after_phpdoc' => true,
         'no_empty_statement' => true,
-        'no_extra_consecutive_blank_lines' => true,
+        'no_extra_blank_lines' => true,
         'no_leading_import_slash' => true,
         'no_leading_namespace_whitespace' => true,
         'no_trailing_comma_in_singleline_array' => true,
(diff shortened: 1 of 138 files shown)


.php-cs-fixer.php
```diff
@@ -30,11 +28,11 @@
         'cast_spaces' => ['space' => 'single'],
         'header_comment' => ['header' => $header],
         'include' => true,
-        'class_attributes_separation' => ['elements' => ['method']],
+        'class_attributes_separation' => array('elements' => array('method' => 'one', 'trait_import' => 'none')),
```
stof, line 31: why not keeping short arrays ?

src/Monolog/Formatter/ChromePHPFormatter.php
```diff
@@ -24,18 +25,21 @@ class ChromePHPFormatter implements FormatterInterface
...
-        Logger::DEBUG     => 'log',
-        Logger::INFO      => 'info',
-        Logger::NOTICE    => 'info',
-        Logger::WARNING   => 'warn',
-        Logger::ERROR     => 'error',
-        Logger::CRITICAL  => 'error',
-        Logger::ALERT     => 'error',
-        Logger::EMERGENCY => 'error',
-    ];
+    protected function toWildfireLevel(Level $level): string
```
stof, line 30: why is it protected while it was private before ?

src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -203,16 +203,12 @@ public function getSlackData(LogRecord $record): array
      */
     public function getAttachmentColor(int $level): string
```
stof, line 204: shouldn't this get a Level as argument ?

src/Monolog/Level.php
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * @return string
+     * @phpstan-return \Psr\Log\LogLevel::*
+     */
+    public function toPsrLogLevel(): string
+    {
+        return match ($this) {
+            self::Debug => 'debug',
```
stof, line 104: why not using the LogLevel constants there ?

src/Monolog/LogRecord.php
```diff
@@ -81,8 +77,15 @@ public function offsetUnset(mixed $offset): void
...
-            $offset = 'levelName';
+        if ($offset === 'level_name' || $offset === 'level') {
+            if ($offset === 'level_name') {
+                $offset = 'levelName';
+            }
+
+            // avoid returning readonly props by ref as this is illegal
+            $copy = $this->{$offset};
+
+            return $copy->value;
```
stof, line 88: should you return the value of an enum by ref ?

src/Monolog/Logger.php
```diff
@@ -287,16 +282,16 @@ public function useMicrosecondTimestamps(bool $micro): self
...
-    public function addRecord(int $level, string $message, array $context = []): bool
+    public function addRecord(int|Level $level, string $message, array $context = []): bool
     {
         $recordInitialized = count($this->processors) === 0;
 
         $record = new LogRecord(
             message: $message,
             context: $context,
-            level: $level,
+            level: $level instanceof Level ? $level : self::toMonologLevel($level),
```
stof, line 294: `toMonologLevel` already handles the case of `instanceof Level`

src/Monolog/Logger.php
```diff
@@ -494,7 +474,7 @@ public function getExceptionHandler(): ?callable
      */
     public function log($level, string|\Stringable $message, array $context = []): void
     {
-        if (!is_int($level) && !is_string($level)) {
+        if (!is_string($level) && !is_int($level) && !$level instanceof Level) {
             throw new \InvalidArgumentException('$level is expected to be a string or int');
```
stof, line 478: or a `Level`

Seldaek:
Thanks @stof for the review 👍🏻 Will merge now then.

PR #1657, opened 2022-04-22
Update README.md

kristos80:
Just an update on README.md, so that copying/pasting code from Basic Usage example will work


Seldaek:
Yeah the readme is currently that of main branch for 3.0, maybe I should set the default branch to 2.x for now..

PR #1658, opened 2022-04-25
replace deprecated method calls in GelfMessageFormatter

jortgies:
the getter/setter methods for file, level and facility are deprecated in gelf v1.1
add those fields as additional instead, as suggested in the gelf spec (https://docs.graylog.org/v1/docs/gelf#gelf-payload-specification)

### DIFF
--- a/src/Monolog/Formatter/GelfMessageFormatter.php
+++ b/src/Monolog/Formatter/GelfMessageFormatter.php
@@ -114,14 +114,14 @@ class GelfMessageFormatter extends NormalizerFormatter
         }
 
         if (isset($record['channel'])) {
-            $message->setFacility($record['channel']);
+            $message->setAdditional('facility', $record['channel']);
         }
         if (isset($extra['line'])) {
-            $message->setLine($extra['line']);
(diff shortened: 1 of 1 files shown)


Seldaek:
Thanks - this breaks the tests tho, and I do wonder if it's going to be a BC break as well for people using existing code? I don't know what the impact is if suddenly the gelf messages change? I am not familiar enough with the gelf backends to tell, so I'd appreciate feedback here.

If it's a BC issue we'd need to either only do this for graylog2/gelf-php v2+, or move it to Monolog v3.

jortgies:
You're right. This probably breaks BC as the field then shows up as `_facility` instead of `facility`. Which _could_ break stuff for people still using older graylog/gelf versions.

Seldaek:
OK let's target for 3.0 then, we can always backport to 2 if it makes sense.

Seldaek:
If you can rebase on main and fix tests that'd be great.

jortgies:
Not sure how to change the PR, so I created a new one at #1664

Issue #1660, opened 2022-04-26
Error Elasticsearch client reference and not supported _type tag


davidfabbretti:
https://github.com/Seldaek/monolog/blob/4192345e260f1d51b365536199744b987e160edc/src/Monolog/Handler/ElasticsearchHandler.php#L21

Must be:

use Elastic\Elasticsearch\Client;

AND

https://github.com/Seldaek/monolog/blob/4192345e260f1d51b365536199744b987e160edc/src/Monolog/Handler/ElasticsearchHandler.php#L138

Unsupported _type tag

davidfabbretti:
400 Bad Request: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Action/metadata line [1] contains an unknown parameter [_type]"}],"type":"illegal_argument_exception","reason":"Action/metadata line [1] contains an unknown parameter [_type]"},"status":400}

Seldaek:
The problem is you're using elasticsearch/elasticsearch v8 which changed some things.

See #1662 for ES8 support, but in the meantime I suggest you use v7.

ezimuel:
@davidfabbretti the `_type` is not supported anymore in Elasticsearch 8. This parameter has been deprecated in Elasticsearch 7, read [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html) for more details.

Seldaek:
OK should be fixed by #1662

PR #1662, opened 2022-05-03
ElasticSearch v8 support

Seldaek:
I'd be happy to have help/guidance here cc @ezimuel 

Any other changes required? 

#1660 mentioned something about `_type` failing, and I'm not sure what it's about as I have close to no experience with elastic search myself.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,19 +17,22 @@
         "psr/log": "^1.0.1 || ^2.0 || ^3.0"
     },
     "require-dev": {
+        "ext-json": "*",
         "aws/aws-sdk-php": "^2.4.9 || ^3.0",
         "doctrine/couchdb": "~1.0@dev",
         "elasticsearch/elasticsearch": "^7 || ^8",
         "graylog2/gelf-php": "^1.4.2",
+        "guzzlehttp/guzzle": "^7.4",
+        "guzzlehttp/psr7": "^2.2",
         "mongodb/mongodb": "^1.8",
         "php-amqplib/php-amqplib": "~2.4 || ^3",
         "php-console/php-console": "^3.1.3",
         "phpspec/prophecy": "^1.15",
         "phpstan/phpstan": "^0.12.91",
-        "phpunit/phpunit": "^8.5",
+        "phpunit/phpunit": "^8.5.14",
         "predis/predis": "^1.1",
         "rollbar/rollbar": "^1.3 || ^2 || ^3",
-        "ruflin/elastica": ">=0.90@dev",
+        "ruflin/elastica": "^7",
         "swiftmailer/swiftmailer": "^5.3|^6.0",
         "symfony/mailer": "^5.4 || ^6",
         "symfony/mime": "^5.4 || ^6"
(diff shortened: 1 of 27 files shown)


composer.json
```diff
@@ -19,9 +19,9 @@
     "require-dev": {
         "aws/aws-sdk-php": "^2.4.9 || ^3.0",
         "doctrine/couchdb": "~1.0@dev",
-        "elasticsearch/elasticsearch": "^7",
```
mimmi20, line 22: `ruflin/elastica` also has a dependency to `elasticsearch/elasticsearch`, and they don't have a 8.x release yet. How do you test the es8 branch in your pipeline?

mimmi20:
Could you update the dependency to `ruflin/elastica` too, as they are using the same major version numbers?

src/Monolog/Handler/ElasticsearchHandler.php
```diff
@@ -54,11 +56,15 @@ class ElasticsearchHandler extends AbstractProcessingHandler
...
-     * @param Client  $client  Elasticsearch Client object
-     * @param mixed[] $options Handler configuration
+     * @param Client|Client8 $client  Elasticsearch Client object
+     * @param mixed[]        $options Handler configuration
      */
-    public function __construct(Client $client, array $options = [], $level = Logger::DEBUG, bool $bubble = true)
+    public function __construct($client, array $options = [], $level = Logger::DEBUG, bool $bubble = true)
     {
+        if (!$client instanceof Client && !$client instanceof Client8) {
+            throw new \InvalidArgumentException('Elasticsearch\Client or Elastic\Elasticsearch\Client instance required');
```
stof, line 65: I would throw a TypeError here, so that it is the same type than when using an argument type

composer.json
Seldaek: Yeah ruflin/elastica is blocking this from being testable, we'd need to remove it in the pipeline to be able to run the tests against elsaticsearch/elasticsearch v8

Seldaek:
Well this is a friggin nightmare 😆 If someone has time to take over and fix things before I get to it feel free.

ezimuel:
@Seldaek I'll look into it soon. The `_type` has been deprecated in Elasticsearch 7 and removed in 8 (see [here](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html) for more details).

Seldaek:
Yup I got that far now that it's running the tests against a real instance but then the elastica handler tests started failing not sure why..

src/Monolog/Handler/ElasticsearchHandler.php
```diff
@@ -69,6 +81,14 @@ public function __construct(Client $client, array $options = [], $level = Logger
             ],
             $options
         );
+
+        if ($client instanceof Client8 || $client::VERSION[0] === '7') {
```
mimmi20, line 85: When is this comparison false? Clients before version 7 are not supported

Seldaek: Well only those versions are used in the tests but we don't enforce it when used within a project so I thought it safer to check and then used the same check in tests for consistency even tho there it's useless.

mimmi20:
Please have a look on PR #1666

Seldaek:
Hell of a scope creep here, but I think it's good to merge :) Thanks for the help.

PR #1663, opened 2022-05-03
Add SymfonyMailerHandler, deprecate SwiftMailerHandler

### LINKED ISSUE #1621: Implement a symfony/mailer handler
As swiftmailer is abandoned many of us are "forced" to search for an alternative. The recommended alternative is [symfony/mailer](https://github.com/symfony/mailer).

I'm not using the Symfony Framework but only the ```symfony/mailer```. I know there is an implementation inside the [monolog-bridge](https://github.com/symfony/monolog-bridge). But the ```monolog-bridge``` has some dependencies which I absolutely not need e.g. ```symfony/http-kernel```. So i don't want to require the ```monolog-bridge``` as I'm only missing the monolog handler. 

Is there a chance we could get an implementation directly inside the monolog package like with the swiftmailer handler?

### LINKED ISSUE #1638: Deprecate SwiftMailerHandler in 2.x
Perhaps add a new one based on symfony/mailer? Not sure if it is needed or if symfony's monolog-bridge covers that well enough already.

Also deprecate MandrillHandler's SwiftMessage arg

Seldaek:
Fixes #1638
Fixes #1621

It accepts MailerInterface or TransportInterface to be interoperable with Laravel too as per https://github.com/Seldaek/monolog/issues/1638#issuecomment-1065085268

### DIFF
--- a/src/Monolog/Handler/SwiftMailerHandler.php
+++ b/src/Monolog/Handler/SwiftMailerHandler.php
@@ -24,6 +24,7 @@ use Swift;
  * @author Gyula Sallai
  *
  * @phpstan-import-type Record from \Monolog\Logger
+ * @deprecated Since Monolog 2.6. Use SymfonyMailerHandler instead.
  */
 class SwiftMailerHandler extends MailHandler
 {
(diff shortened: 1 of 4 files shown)


Seldaek:
cc @stof if you have time to review this it'd be nice, I need it to finalize the 3.0 release

src/Monolog/Handler/SwiftMailerHandler.php
```diff
@@ -42,6 +43,8 @@ public function __construct(\Swift_Mailer $mailer, $message, $level = Logger::ER
     {
         parent::__construct($level, $bubble);
 
+        @trigger_error('The SwiftMailerHandler is deprecated. Use SymfonyMailerHandler instead.', E_USER_DEPRECATED);
```
nicolas-grekas, line 46: It's better to always mention the package+version that the deprecation is for since this makes it easier to understand aggregated deprecation reports that might be about various packages. (same in the annotations btw)

`since monolog/monolog v2.6`

PR #1664, opened 2022-05-04
replace deprecated method calls in GelfMessageFormatter

jortgies:
- the getter/setter methods for file, level and facility are deprecated in gelf v1.1
- add those fields as additional instead, as suggested in the gelf spec (https://docs.graylog.org/v1/docs/gelf#gelf-payload-specification)
- update tests to reflect changes

(continuation of #1658)


Seldaek:
Thanks!

PR #1666, opened 2022-05-07
update pipeline for Elasticsearch tests

mimmi20:


Seldaek:
Thanks a bunch, this is a better model I guess splitting the ES related tests in their own flow. I don't think it's justified to run so many of them though, but I'll remove some of the matrix variants myself in the other PR.

PR #1668, opened 2022-05-09
Remove LevelName enum in favor of a Level::getName method

### LINKED ISSUE #1667: v3: Consider merging Level and LevelName enums
Parsing v3 code I feel like splitting the level name in it's own Enum doesn't bring any benefit while on the contrary it adds complexity and probably confusion for users. 

They will have to constantly think about which one they should use in their context while in fact all they want is to use a Level. The name is just a property of the Level, the same way the `int` value is.

Could merging both enums into a single `Level` enum be considered an improvement ?

Something like this:

```php
// Not a backed enum anymore
enum Level
{
    /**
     * Detailed debug information
     */
    case Debug;
   
    private const DEBUG_VALUE = 100;
    private const DEBUG_NAME = 'DEBUG';

    /**
     * @phpstan-param Level::*_NAME $name
     */
    public static function fromName(string $name): self
    {
        return match ($name) {
            self::DEBUG_NAME => self::Debug,
        };
    }

    /**
     * @phpstan-param Level::*_VALUE $value
     */
    public static function fromValue(int $value): self
    {
        return match ($value) {
            self::DEBUG_VALUE => self::Debug,
        };
    }

    /**
     * @phpstan-return Level::*_NAME
     */
    public function getName(): string
    {
        return match ($this) {
            self::Debug => self::DEBUG_NAME,
        };
    }

    /**
     * @phpstan-return Level::*_VALUE
     */
    public function getValue(): int
    {
        return match ($this) {
            self::Debug => self::DEBUG_VALUE,
        };
    }

```

Difference between current and supposedly improved usage:

*Current version*

```php
// retrieved log level name as a string and test if it's debug
$logLevel = 'debug'
LevelName::from($logLevel) === LevelName::Debug

// should I compare the *name* or the level itself ? What's the best practice here ? 🤔 Maybe the level seems more appropriate
Level::fromLevelName(LevelName::from($logLevel)) === Level::Debug

// seems a bit verbose...

// retrieved log level as an int and test if it's debug
$logLevel = 100;
Level::from($logLevel) === Level::Debug

// now there's only one way to do the check, just because the initial value to represent a Level was an int instead of a string, seems not consistent.

//retrieving the name of a level
Level::Debug->toLevelName()->value;
```

*Improved version*

```php
// retrieved log level name as a string
$logLevel = 'debug'
Level::fromName($logLevel) === Level::Debug

// retrieved log level name as an int
$logLevel = 100
Level::fromValue($logLevel) === Level::Debug

// Only 1 level concept to manage the level, api is clear on whether you need to represent a level from its name or its value, both can even be compared

Level::fromValue(100) === Level::fromName('debug')

//retrieving the name of a level
Level::Debug->getName();
```

Seldaek:
Fixes #1667

The LevelName enum is gone, most relevant changes are in the Level enum https://github.com/Seldaek/monolog/pull/1668/files#diff-b1d4336a8d582db6a4cc0b40f17ab8adf1436ca288e41daa6eb0cdc931b4a480

cc @tigitz


stof (approved):

tigitz:
@Seldaek Awesome, feels much cleaner. Glad it led to a positive contribution. Thanks for your consideration !

PR #1671, opened 2022-05-13
RotatingFileHandler to maintain the correct maximum number files when using folder by date

kermage:
Setting filename format with slashes

```php
$handler = new RotatingFileHandler('path/to/save/sample.log', 2);

$handler->setFilenameFormat('{date}/{filename}', 'Y/m/d');
```

`path/to/save/[year]/[month]/[day]/sample.log`

### DIFF
--- a/src/Monolog/Handler/RotatingFileHandler.php
+++ b/src/Monolog/Handler/RotatingFileHandler.php
@@ -191,7 +191,11 @@ class RotatingFileHandler extends StreamHandler
         $fileInfo = pathinfo($this->filename);
         $glob = str_replace(
             ['{filename}', '{date}'],
-            [$fileInfo['filename'], '[0-9][0-9][0-9][0-9]*'],
+            [$fileInfo['filename'], str_replace(
+                ['Y', 'm', 'd'],
+                '[0-9]*',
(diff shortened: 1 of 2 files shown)


src/Monolog/Handler/RotatingFileHandler.php
```diff
@@ -191,7 +191,11 @@ protected function getGlobPattern(): string
         $fileInfo = pathinfo($this->filename);
         $glob = str_replace(
             ['{filename}', '{date}'],
-            [$fileInfo['filename'], '[0-9][0-9][0-9][0-9]*'],
+            [$fileInfo['filename'], str_replace(
+                ['Y', 'm', 'd'],
+                ['[0-9][0-9][0-9][0-9]', '[0-9][0-9]', '[0-9][0-9]'],
```
Seldaek, line 196: We should also support 2 digit years IMO as it is allowed by setFilenameFormat

```suggestion
                ['Y', 'y', 'm', 'd'],
                ['[0-9][0-9][0-9][0-9]', '[0-9][0-9]', '[0-9][0-9]', '[0-9][0-9]'],
```

PR #1672, opened 2022-05-16
Fixed the bug of log depth calculation error when using coroutine such as swoole.

limingxinleo:
https://github.com/hyperf/hyperf/issues/4740

https://github.com/hyperf/hyperf/pull/4754

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -148,9 +148,9 @@ class Logger implements LoggerInterface, ResettableInterface
     protected $exceptionHandler;
 
     /**
-     * @var int Keeps track of depth to prevent infinite logging loops
+     * @var LogDepthInterface Keeps track of depth to prevent infinite logging loops
      */
-    private $logDepth = 0;
+    protected $logDepth;
 
     /**
      * @psalm-param array<callable(array): array> $processors
@@ -166,6 +166,7 @@ class Logger implements LoggerInterface, ResettableInterface
         $this->setHandlers($handlers);
         $this->processors = $processors;
         $this->timezone = $timezone ?: new DateTimeZone(date_default_timezone_get() ?: 'UTC');
+        $this->logDepth = new LogDepth();
     }
 
     public function getName(): string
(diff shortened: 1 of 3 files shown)


limingxinleo:
Then I just need to rewrite `logDepth` to solve this problem easily

src/Monolog/Logger.php
```diff
@@ -148,9 +148,9 @@ class Logger implements LoggerInterface, ResettableInterface
     protected $exceptionHandler;
 
     /**
-     * @var int Keeps track of depth to prevent infinite logging loops
+     * @var LogDepthInterface Keeps track of depth to prevent infinite logging loops
      */
-    private $logDepth = 0;
+    protected $logDepth;
```
stof, line 153: why protected ?

limingxinleo: Because I will create a Logger which extends Monolog

And then rewrite logDepth

limingxinleo:
> I don't see how storing the depth in a separate class solves the bug.

I can rewrite logDepth like this

class LogDepth
{
public function increment()｛return 0}
public function decrement() ｛
return 0}
}

src/Monolog/Logger.php
stof: Well, if you want to replace the tracking of the log depth, what you need is an interface for that object (to allow providing another implementation) and a way to inject a custom implementation.

But inheritance-based extension points are the worse ones both in term of maintenance (it is a lot harder to write a BC layer for a protected hooking point than for a clean extension point) and in term of flexibility (if 2 packages extend the Logger to hook into different places, you have to choose between them, you cannot combine their extensions)

stof:
> I can rewrite logDepth like this

That would just disable the tracking of logging loops entirely, which could then be achieved by passing a boolean flag to control that tracking in a much simpler way.

limingxinleo:
> That would just disable the tracking of logging loops entirely, which could then be achieved by passing a boolean flag to control that tracking in a much simpler way.

You are right

PR #1675, opened 2022-05-24
Backport #1552 fix in 2.3.0 to 1.x branch

robocoder:


Seldaek:
This unfortunately caused more issues and had to be fixed in further commits (https://github.com/Seldaek/monolog/commit/0b22036ab67fa7752e53e8bb18891894e31137d1 https://github.com/Seldaek/monolog/commit/70fe0928675d1f1080a9d6496f549baf5ae482a9  & https://github.com/Seldaek/monolog/commit/9d1fed4affc44c39ef8e62f98c220151b76a5453)

For simplicity I think we could simply set the chunk size fixed at 512KB in 1.x which already increases a lot from the 8KB and most likely fixes it, logging more than 512KB in one record is not that common I'd hope, but it'd be unlikely to cause memory usage issues.

Seldaek:
The best fix though would be to upgrade to Monolog 2 (or 3) :)

robocoder:
ok, feel free to close.

I also didn't know stream_set_chunk_size() was php5.4+ until I saw the failed CI test.

src/Monolog/Handler/StreamHandler.php
```diff
@@ -23,6 +23,9 @@
  */
 class StreamHandler extends AbstractProcessingHandler
 {
+    const MAX_CHUNK_SIZE = 536870911;
```
Seldaek, line 26: 512KB is 524288, I think you have 512MB here :)

Seldaek:
Merged this with the correct 512KB chunk size as https://github.com/Seldaek/monolog/commit/07a0d1ed33ebd847164c986dbb4c255bef5d9df2 - thanks!

robocoder:
Thanks for catching that!

robocoder:
Thanks again. Two weeks have passed without any truncated lines.

PR #1681, opened 2022-05-31
Added `$logDepth` which you can used to close log depth check.

limingxinleo:
https://github.com/Seldaek/monolog/pull/1672

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -148,9 +148,9 @@ class Logger implements LoggerInterface, ResettableInterface
     protected $exceptionHandler;
 
     /**
-     * @var int Keeps track of depth to prevent infinite logging loops
+     * @var int|null Keeps track of depth to prevent infinite logging loops
      */
-    private $logDepth = 0;
+    private $logDepth;
 
     /**
      * @psalm-param array<callable(array): array> $processors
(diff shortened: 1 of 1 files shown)


limingxinleo:
https://github.com/hyperf/hyperf/issues/4740

https://github.com/hyperf/hyperf/pull/4754

src/Monolog/Logger.php
```diff
@@ -148,9 +148,9 @@ class Logger implements LoggerInterface, ResettableInterface
     protected $exceptionHandler;
 
     /**
-     * @var int Keeps track of depth to prevent infinite logging loops
+     * @var int|null Keeps track of depth to prevent infinite logging loops
      */
-    private $logDepth = 0;
+    private $logDepth;
```
Seldaek, line 153: Explicit null is better as it doesn't break with typed properties (3.x uses typed props and this will merge into it so rather be prepared)

```suggestion
    private $logDepth = null;
```

Seldaek:
Decided to do this differently, see https://github.com/Seldaek/monolog/commit/9c1566a971d42ff1acb4756708e8d9a5568bb194 - It'll be part of 2.7.0 when I get to it.

Seldaek:
So I think you can already update your conflict rule to conflict with 2.6.*, and add a method_exists() check to disable that feature if the method is present?

PR #1682, opened 2022-06-07
Possibility to overwrite the datetime of a record

Henning256:
Added an optional parameter to the loggers addRecord method to overwrite the datetime of a record.
This is useful to decouple the time a record is issued and when it is persisted.
For example when caching the records to log them asynchronous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 [![Total Downloads](https://img.shields.io/packagist/dt/monolog/monolog.svg)](https://packagist.org/packages/monolog/monolog)
 [![Latest Stable Version](https://img.shields.io/packagist/v/monolog/monolog.svg)](https://packagist.org/packages/monolog/monolog)
 
+> ⚠ This is the **documentation for Monolog 3.x**, if you are using older releases
+> see the documentation for [Monolog 2.x](https://github.com/Seldaek/monolog/blob/2.x/README.md) or [Monolog 1.x](https://github.com/Seldaek/monolog/blob/1.x/README.md) ⚠
 
 Monolog sends your logs to files, sockets, inboxes, databases and various
 web services. See the complete list of handlers below. Special handlers


Seldaek:
This sounds reasonable albeit kinda niche.. But adding it only to addRecord sounds ok to me as that's very low level.

Would you mind rebasing on the 2.x branch tho? I think this probably should go there as well as the 3.x adoption is very low still.

Seldaek:
Ah well, I messed up and pushed main into your main instead of a local branch I had to do the rebase, so that closed the PR..

Anyway the code is merged in 2.x branch as 24e414c - thanks :)

WoZ:
@Henning256 @Seldaek I've found this issue in the 2.7/3.1 changelog. @Henning256 I understand your approach and solved the same issue a few weeks ago, but I decided to store 2 times: when the log record was generated and when the log record was dispatched. It's just a tip, not to forget to log both of them, it will help to debug queues of log entries and delays in the dispatching. Cheers.

Henning256:
> @Henning256 @Seldaek I've found this issue in the 2.7/3.1 changelog. @Henning256 I understand your approach and solved the same issue a few weeks ago, but I decided to store 2 times: when the log record was generated and when the log record was dispatched. It's just a tip, not to forget to log both of them, it will help to debug queues of log entries and delays in the dispatching. Cheers.

Hi and thank you.
This sounds like something that belongs in your handler.
i.e. I wrote a Handler that logs to the database which logs the timestamp from the record as "issued_at" and the actual time as "created_at".

As I mentioned in my PR I implemented this feature for an async logging solution which works like this:
I have a handler that writes the log record along with a destination channel into the cache.
Then I implemented a Job that reads the cached records and logs them to their destination channels.
And this is exactly the point where the new feature comes in because now I can use the original timestamp from the record to log the record with the addRecord function to its destination channel and not loose the information when the record was issued.

Hope that helped someone.

Cheers

Issue #1678, opened 2022-05-26
Concurrency directory creation


Legion112:
Monolog version 2.4.0

We run test in parallel in sometimes we experience following behaviour:

```
 There is no existing directory at "/php-tests/report/logs/e2e/PathToTestDir" and it could not be cr  
  eated: File exists   
```
  
  This is from `\Monolog\Handler\StreamHandler::createDir`

Seldaek:
Oh we did check for concurrency but I think there is a missing `clearstatcache()` in there to properly fix this issue. I'll fix this thanks for reporting 👍🏻

Seldaek:
Hm no sorry.. PHP shouldn't cache the fact that a file is missing, so I don't quite see how this code can fail https://github.com/Seldaek/monolog/blob/60ad5183b5e5d6c9d4047e9f3072d36071dcc161/src/Monolog/Handler/StreamHandler.php#L193-L198

If mkdir failed, it checks for is_dir again and if the failure reason is `File exists` then it should definitely be there.

Is it possible that your concurrent test runs create the dir, log, and then delete the dir very quickly so you can run into a race condition where mkdir fails but is_dir still returns false as it was deleted already?

Legion112:
Retry seem like a realy good option here @Seldaek.

Seldaek:
@Legion112 I'm not sure this fixes anything in fact.. As per my comment on https://github.com/Seldaek/monolog/pull/1683#discussion_r893066866 - I believe you may then run into another race condition whereas the directory gets wiped by another process and then the file write fails. The only real solution if you do clear the directory after the test is to run tests in different directories IMO.

Legion112:
@Seldaek well let's check if `$this->errorMessage === 'File exists'` and do not throw exception in that case? 
What do you think?

Seldaek:
Maybe.. but if it exists why did is_dir() return false at line 196? Do you have a reliable way to reproduce the error so that we can try to inspect things further?

Legion112:
https://github.com/Seldaek/monolog/pull/1685

Legion112:
I don't know why `is_dir` returned false. 
I understand that is strange case. 
Maybe series of retry would prevent this issue. 
Like try to run `createDir ` body code 3 times only after 3 error throw exception.

Seldaek:
These are all just band-aids though, not fixes. I would really like to understand what is going on before doing anything that hides the problem.

Legion112:
[...]

PR #1683, opened 2022-06-09
maintain lock for directory creating in parallel/concurrent mode, try to retry if 'File exists occur, but directory is missing'

cheetaah:
Hello! Sometimes there is concurrency in the directory creation, like here https://github.com/Seldaek/monolog/issues/1678
For example, when you use paratest in multiple processes.
I think resource locking for directory creation may help here


src/Monolog/Handler/StreamHandler.php
```diff
@@ -188,15 +191,27 @@ private function createDir(string $url): void
...
-            restore_error_handler();
-            if (false === $status && !is_dir($dir)) {
-                throw new \UnexpectedValueException(sprintf('There is no existing directory at "%s" and it could not be created: '.$this->errorMessage, $dir));
+
+        $this->dirLock = tempnam(sys_get_temp_dir(), md5($dir . time()));
+
+        $resourceLock = fopen($this->dirLock, "r+");
+
+        //  Gets the exclusive lock for creating directory in parallel mode (eg: paratest)
+        if (true === flock($resourceLock, LOCK_EX)) {
```
Seldaek, line 200: I don't quite understand what this solves. The way I see it it is going to create a different dirLock file in every process (as tempnam creates a unique filename) so there is no real protection between processes. I also don't see why this is stored as a property vs a simple local variable here.

One could perhaps acquire a lock on `dirname($dir)` before creating it, that might help, but if you have tests running in parallel which create/destroy the dir repeatedly you will anyway run into issues, the only fix for this is to run every test in its own directory.

cheetaah: Yep, you are right. The resource lock must be the same for all concurrent processes, and should depend on only directory name. Directory lock variable can be local here. 

`but if you have tests running in parallel which create/destroy the dir repeatedly you will anyway run into issues`
yes, you are right, but it's not our case here.

we start paratest in 4-6 processes, it runs phpunit, phpunit uses monolog.
and no idea if phpunut drop monolog, which can be the problem of permanent create/destroy of test working directory..

cheetaah: Could you see then?

Seldaek: I don't quite understand what problem you are trying to fix then. The issue in https://github.com/Seldaek/monolog/issues/1678 is a different one IMO as it appears the dir gets wiped again.. 

Can you share what issue you are seeing? What error message? I'd really like to understand this better before adding layers and layers of locking that may not fix anything.

Seldaek:
Is that the issue you are seeing? Can you reproduce the issue reliably (which would allow us to confirm the fix works)?

Seldaek:
Closing as hopefully https://github.com/Seldaek/monolog/commit/af8efac0d58e524aead247ce64445a66e6d7c553 fixes this issue too - and frankly i am not willing to do more than that without really understanding what is going on.

PR #1689, opened 2022-06-13
finishing Syslog formatter

picass0:
This PR finishes work started by @dkarlovi here https://github.com/Seldaek/monolog/pull/1599

In theory metadata(context and extra) can be stored in native syslog structured data format https://datatracker.ietf.org/doc/html/rfc5424#section-6.3 (as suggested by @dkarlovi) but storing arbitrary structred data requires registering "private enterprise number" with IANA(it says so in here https://datatracker.ietf.org/doc/html/rfc5424#section-6.3.2) which looks like too much work, so i just added metadata as part of a message. Hope this is ok


dkarlovi:
@picass0 thanks so much for working on this! \o/

I'm not picky about structured data being in the message TBH. My only worry is if [lnav](https://lnav.org/) will be able to use it to filter the messages? Being able to have it running in development with my Symfony apps is basically the only reason why I'd want this formatter. Could you test your implementation with it to confirm it will work? If so, I'm all :+1: here.

picass0:
Thanks for response @dkarlovi )
i recorded video of  dev.log  with this formatter, looks really nice imho :)
as you can see in video searching and filtering works fine

https://user-images.githubusercontent.com/10927139/174167190-768d81d7-f0d2-45f9-82ed-3473f6ee02a0.mp4

Seldaek:
Thank you both!

Seldaek:
Just FYI https://github.com/Seldaek/monolog/commit/f7dfa00f084db62b1dec3b9ef683be686cfa31d7 reverts some changes done here to keep BC and make use of the new Level::toRFC5424Level method - hope that makes sense to you too.

picass0:
> Just FYI [f7dfa00](https://github.com/Seldaek/monolog/commit/f7dfa00f084db62b1dec3b9ef683be686cfa31d7) reverts some changes done here to keep BC and make use of the new Level::toRFC5424Level method - hope that makes sense to you too.

Sorry for breaking BC. I think u missed one use statement here  \Monolog\Formatter\SyslogFormatter:14

Seldaek:
Right, cleaned those up :) And no worries about the BC break, it was very minor so I thought I'd let it slide but then I realized I could as well just fix it easily.

PR #1690, opened 2022-06-15
Add basic support to Google Cloud Logging format

lcobucci:
Google Cloud Logging doesn't show the correct level log when using JsonFormatter, making observability a bit trickier. This applies minor tweaks to the default format, allowing log entries to be properly represented.

There are alternative packages to this but they add fields that aren't strictly required - also performing `debug_backtrace()` calls that are usually not desired when in production mode.

### DIFF
--- a/src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
+++ b/src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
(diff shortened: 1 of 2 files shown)


lcobucci:
@Seldaek does this make sense to you?

It would help us a lot to have it in Monolog since we're having to create a package (or duplicate) this teeny tiny piece of code.

lcobucci:
Thanks, Jordi!
The systems we're running are on PHP 8.1 and we were planning to go with v3.x. Google's lack of v3 support might change this decision but we'll see.

I believe the extension points are slightly different... so, I'd suggest merging this in and I can gladly send another patch for v2, what do you think?

Seldaek:
Thanks

PR #1708, opened 2022-06-16
explicit string variable

MathiasReker:
Converts implicit variables into explicit ones in double-quoted strings or heredoc syntax.


robocoder:
IIRC I read somewhere that using `${var}` in strings is going to be deprecated in php 8.2

stof:
it is indeed deprecated. See https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation

Issue #1720, opened 2022-06-26
New lines in stacktraces broken on Windows in some cases


abordage:
Monolog version 2

In the log:
```
#13 {...}\\laravel\\vendor\\laravel\
ova\\src\\Http\\Middleware\\Authorize.php(18): {...}
#14 {...}\\laravel\\vendor\\laravel\\framework\\src\\Illuminate\\Pipeline\\Pipeline.php(180): {...}
#15 {...}\\laravel\\vendor\\laravel\
ova\\src\\Http\\Middleware\\BootTools.php(20): {...}
#16 {...}\\laravel\\vendor\\laravel\\framework\\src\\Illuminate\\Pipeline\\Pipeline.php(180): {...}
#17 {...}\\laravel\\vendor\\laravel\
ova\\src\\Http\\Middleware\\DispatchServingNovaEvent.php(24): {...}
#18 {...}\\laravel\\vendor\\laravel\\framework\\src\\Illuminate\\Pipeline\\Pipeline.php(180): {...}
```

Must be:
```
#13 {...}\\laravel\\vendor\\laravel\\nova\\src\\Http\\Middleware\\Authorize.php(18): {...}
#14 {...}\\laravel\\vendor\\laravel\\framework\\src\\Illuminate\\Pipeline\\Pipeline.php(180): {...}
#15 {...}\\laravel\\vendor\\laravel\\nova\\src\\Http\\Middleware\\BootTools.php(20): {...}
#16 {...}\\laravel\\vendor\\laravel\\framework\\src\\Illuminate\\Pipeline\\Pipeline.php(180): {...}
#17 {...}\\laravel\\vendor\\laravel\\nova\\src\\Http\\Middleware\\DispatchServingNovaEvent.php(24): {...}
#18 {...}\\laravel\\vendor\\laravel\\framework\\src\\Illuminate\\Pipeline\\Pipeline.php(180): {...}
```

The bug is the same as https://github.com/Seldaek/monolog/issues/1629
It is recommended to disable `allowInlineLineBreaks`, but method `includeStacktraces` automatically turns on `allowInlineLineBreaks` (see [LineFormatter.php#L54-L63](https://github.com/Seldaek/monolog/blob/5579edf28aee1190a798bfa5be8bc16c563bd524/src/Monolog/Formatter/LineFormatter.php#L54-L63))

abordage:
[...]

PR #1721, opened 2022-06-26
Fix replaceNewlines method

abordage:
See https://github.com/Seldaek/monolog/issues/1720


Seldaek:
Thanks, merged slightly tweaked and added tests to verify in 320909a

PR #1722, opened 2022-06-28
Fix utf8_encode deprecation

### LINKED ISSUE #1718: PHP 8.2 - `utf8_encode` deprecated
Monolog version 1|2

**PHP 8.2** 
https://php.watch/versions/8.2/utf8_encode-utf8_decode-deprecated

**Suggestion:**
```php
function_exists('mb_convert_encoding') ? mb_convert_encoding($json, 'UTF-8', mb_list_encodings()) : utf8_encode($json);
```

----

https://github.com/Seldaek/monolog/blob/d381140ca134f7ece43cd2c139d2207bb26dc329/src/Monolog/Utils.php#L203
https://github.com/Seldaek/monolog/blob/5579edf28aee1190a798bfa5be8bc16c563bd524/src/Monolog/Handler/ChromePHPHandler.php#L152
https://github.com/Seldaek/monolog/blob/5579edf28aee1190a798bfa5be8bc16c563bd524/src/Monolog/Handler/ChromePHPHandler.php#L167

erikn69:
Closes #1718

### DIFF
--- a/src/Monolog/Handler/ChromePHPHandler.php
+++ b/src/Monolog/Handler/ChromePHPHandler.php
@@ -148,7 +148,8 @@ class ChromePHPHandler extends AbstractProcessingHandler
             self::$json['request_uri'] = $_SERVER['REQUEST_URI'] ?? '';
         }
 
-        $data = $this->encodeArray();
+        $json = Utils::jsonEncode(self::$json, Utils::DEFAULT_JSON_FLAGS & ~JSON_UNESCAPED_UNICODE, true);
+        $data = base64_encode($json);
         if (strlen($data) > 3 * 1024) {
             self::$overflowed = true;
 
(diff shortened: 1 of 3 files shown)


src/Monolog/Handler/ChromePHPHandler.php
```diff
@@ -163,15 +162,28 @@ protected function send(): void
...
+     */
+    private function encodeArray(): string
+    {
+        $json = Utils::jsonEncode(self::$json, Utils::DEFAULT_JSON_FLAGS & ~JSON_UNESCAPED_UNICODE, true);
+
+        return base64_encode(
+            function_exists('mb_convert_encoding')
+            ? mb_convert_encoding($json, 'UTF-8', mb_list_encodings())
+            : utf8_encode($json)
+        );
```
Seldaek, line 184: I think we can actually skip the encoding here, as json_encode will produce utf-8 or fail since json must be utf-8. And we also have JSON_INVALID_UTF8_SUBSTITUTE in the Utils::DEFAULT_JSON_FLAGS now so it is really useless I think.

Anyway I'll fix the patch.

Seldaek:
Thanks

Issue #1686, opened 2022-06-11
Adding support for RFC 5424 levels in Logger::toMonologLevel


Seldaek:
We could support this which would alleviate the need for hacks like https://git.drupalcode.org/project/monolog/-/blob/2.x/src/Logger/Logger.php#L29-50

The numbers do not conflict with the monolog level constants, so this should be fine.

In 3.x we could also have a method on Monolog\Level to turn a case back into a RFC 5424 level.

/cc @lussoluca any opinion from Drupal Monolog side? I think it'd be great to remove the need for overriding methods as it would prevent problems like #1684

lussoluca:
I agree with you, I can start working on this in the next days

PR #1723, opened 2022-07-05
#1686 - Add RFC 5424 support

lussoluca:


Seldaek:
I switched it to handle this in addRecord() and log() exclusively, because doing so in toMonologLevel carried some risk as this is used throughout handlers to allow configuration using level names, and if you pass level 0 for ex to say you want to react to all levels, this worked fine before but it would make things react only on EMERGENCY if toMonologLevel converts 0 to emergency.

Otherwise, look good to me now, thanks!

PR #1731, opened 2022-07-09
Refactor null coalescing operator

mehdihasanpour:


Seldaek:
Thanks but I'm closing as code style can be automated with php-cs-fixer and is the responsibility of maintainers, see https://github.com/composer/composer/pull/10871#issuecomment-1163039319 for a longer rationale.

PR #1739, opened 2022-07-26
Update BrowserConsoleHandler console output method

pafernandez-oesia:
Instead of using `console.log` for all log levels, it uses `debug`, `info`, `warn` and `error` methods depending on the log level.
This makes filtering logs easier in the browser console and highlights error level messages.


pafernandez-oesia:
I think there must be a cleaner way to do it. Maybe someone with better knowledge of the base code can make a suggestion.

Seldaek:
That looks alright to me but it'd be nice if you can rebase it on the 2.x branch so it can be added to 2.9 release as well. I'll then merge it into main and change it to a match block there which would look a little cleaner.

Seldaek:
Looks correct yes, thanks. I'm on vacation so merging and releasing will take a while

PR #1743, opened 2022-08-05
Add SapiProcessor

lyrixx:
I use this processor quite often, so I think it can be useful for others too

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### 3.3.0 (unreleased)
+
+ * Added `SapiProcessor`
+
 ### 3.2.0 (2022-07-24)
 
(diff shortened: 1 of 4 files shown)


Seldaek:
The SAPI itself seems kinda useless to me, but maybe I'm missing something? The process name thing seems more useful but still I kinda wonder what the use case is how do you typically use this?

lyrixx:
It's mainly to filter logs in kibana or similar tools. But indeed I copy this processor from project to project but I'm not sure it's very useful. I'm okay to close it your think it's useless

Seldaek:
OK, I don't super clearly see the use case I have to say, so yeah maybe rather close and you publish it as a third party repo if you wanna stop copy pasting it around :)

PR #1753, opened 2022-08-23
Fix cycle detection within fibers

### LINKED ISSUE #1752: Cycle detection doesn't work well with fibers
[Cycle detection](https://github.com/Seldaek/monolog/blob/e68c00670e310e276da9624d75965f5d2f28aa80/src/Monolog/Logger.php#L324) is broken when using fibers, because it assumes that if the function is entered multiple times, it's a cycle. With fibers, the function might be entered multiple times in different fibers if one of them suspends somewhere within the function.

What are your thoughts to moving cycle detection to an interface? There could be a `NullCycleDetector` that doesn't detect cycles at all, a `CountingCycleDetector` and then a [`FiberLocal`](https://github.com/revoltphp/event-loop/blob/4e7455506d9ff61f3705b76a20a8f12ad4dc1552/src/EventLoop/FiberLocal.php) based implementation that keeps track of the depth for each fiber separately. I'm of course also fine with fibers being supported natively, but I don't think it's a good idea to put this logic into `Logger` itself in that case.

kelunik:
We keep a separate depth count per fiber.

Fixes #1752.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -169,7 +169,7 @@ class Logger implements LoggerInterface, ResettableInterface
     private $logDepth = 0;
 
     /**
-     * @var \WeakMap|null Keeps track of depth inside fibers to prevent infinite logging loops
+     * @var \WeakMap<\Fiber, int>|null Keeps track of depth inside fibers to prevent infinite logging loops
      */
     private $fiberLogDepth;
 
@@ -196,7 +196,10 @@ class Logger implements LoggerInterface, ResettableInterface
         $this->timezone = $timezone ?: new DateTimeZone(date_default_timezone_get() ?: 'UTC');
 
         if (\PHP_VERSION_ID >= 80100) {
-            $this->fiberLogDepth = new \WeakMap();
+            // Local variable for phpstan, see https://github.com/phpstan/phpstan/issues/6732#issuecomment-1111118412
+            /** @var \WeakMap<\Fiber, int> $fiberLogDepth */
+            $fiberLogDepth = new \WeakMap();
+            $this->fiberLogDepth = $fiberLogDepth;
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/Monolog/Logger.php
```diff
@@ -387,7 +404,11 @@ public function addRecord(int $level, string $message, array $context = [], Date
             }
         } finally {
             if ($this->detectCycles) {
-                $this->logDepth--;
+                if (\PHP_VERSION_ID >= 80100 && $fiber = \Fiber::getCurrent()) {
```
Seldaek, line 407: Kinda unfortunate that we get a hit here of getting the current fiber for every log created, even if not using fibers at all.. I don't know maybe that's got very little perf impact and it's fine?

kelunik: We can reduce it to a single call instead of two by storing the current fiber in a local variable. I think a single call there is a fine price to pay for it working by default, but I'm obviously biased.

Seldaek: Yeah I mean this call could definitely be removed as $fiber is already set above, but my point was more avoiding both calls ;) But I do see the point about having it just work.

Seems like 100K calls to Fiber::getCurrent are done in 7/8ms in a quick test here, so I guess this is really a non-issue.

kelunik: Yes, apart from the typical function call overhead, it's very cheap, just a global lookup:

https://github.com/php/php-src/blob/396b2aab85ebab0a72206e74ce87eda75a4167fd/Zend/zend_fibers.c#L864-L875

What should I do about the PHPStan errors, seems like PHPStan doesn't know about fibers?

Seldaek: Yeah we can ignore those in the phpstan config I guess, and report it to PHPStan if there is no issue for it yet.

staabm: went ahead a reported a phpstan issue: https://github.com/phpstan/phpstan/issues/7843

kelunik: Thanks!

kelunik: I've added ignores and removed the duplicate call to `Fiber::getCurrent()`.

kelunik:
@Seldaek Anything left to do before this can be merged?

Seldaek:
Nope, looks good, thanks!

PR #1756, opened 2022-08-28
add phpdoc

moharami:


Seldaek:
Thanks but we do not duplicate real php types in phpdoc if they don't add any value.

Issue #1754, opened 2022-08-26
GoogleCloudLoggingFormatter timestamp does not work with Google Cloud Run


rpkamp:
Monolog version 2 and 3.

In our application we have a fingers crossed handler that passes to a stream handler that writes to `php://stderr` using the Google Cloud Logging Formatter. This mostly works fine, except that Google Cloud Run does not use the `timestamp` sent from monolog as the timestamp of the message. It seems to be taking the time it received the message instead. 
Most notably when we logged several messages, sleeping for 1 second between those messages, they all more or less the same timestamp in Google Cloud Logging, they certainly weren't a second apart. 

Diving into it, it turns out sending messages through `php://stderr` is different than sending them through the logging API. Instead of `timestamp` the field `time` must be used, and the time must be in UTC.

We found that the following works for us when writing to `php://stderr` on Google Cloud Run (in Monolog 2):

```
$record['time'] = $record['datetime']->setTimezone(new DateTimeZone('UTC'))->format('Y-m-d\TH:i:s.u\Z');
```

With this in place the messages were logged seconds apart, even though they were still logged all at once by the fingers crossed handler.

I do wonder whether this should be changed in Monolog, because I don't know if the current implementation has a use case that I'm unaware off that uses the format that is now in the code (something else than Cloud Run and/or writing to `php://stderr`). 

@lcobucci since you wrote the formatter in Monolog, do you have ideas on this?

lcobucci:
I'm running this via Google Kubernetes Engine (relying on `php://stderr` too) and it seems to be working fine. It's interesting that Cloud Run and GKE diverge... well, maybe it's just Google being Google idk 🤷‍♂️ 

I don't see a reason not to support both, though... maybe `$record['time'] = $record['timestamp'] = $record['datetime']->...`?

Seldaek:
Seems kinda silly to log the time twice there, but if it's gotta be done ..

Any chance that time would also work for GKE @lcobucci ?

lcobucci:
I'll test it and come back to you 👍

rpkamp:
@lcobucci Please also check if it works when logging in UTC time. Cloud Run doesn't like our default timezone (CEST).

(see my first post for how we're doing it now)

lcobucci:
Hey folks, sorry about the delay.

We've tested the format by writing straight to `STDERR`:

```jsonld
[...]

PR #1758, opened 2022-09-08
Fix date/time field for GCP logging

lcobucci:
Fixes https://github.com/Seldaek/monolog/issues/1754

We introduced the formatter using the incorrect field, which was making GCP not to extract the correct value from the payload.

This uses the field that actually maps to the expected format.

More info: https://cloud.google.com/logging/docs/structured-logging


stof (approved):

Seldaek:
Thanks

PR #1759, opened 2022-09-08
Fix date/time field for GCP logging

### LINKED ISSUE #1754: GoogleCloudLoggingFormatter timestamp does not work with Google Cloud Run

lcobucci:
Fixes #1754

This uses the field that actually maps to the expected format.

More info: https://cloud.google.com/logging/docs/structured-logging

### DIFF
--- a/src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
+++ b/src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
@@ -17,6 +17,7 @@ use Monolog\LogRecord;
 /**
  * Encodes message information into JSON in a format compatible with Cloud logging.
  *
+ * @see https://cloud.google.com/logging/docs/structured-logging
  * @see https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry
  *
  * @author Luís Cobucci <lcobucci@gmail.com>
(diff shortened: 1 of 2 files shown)


lcobucci:
@Seldaek it seems like the error comes from the main branch...

Seldaek:
That's very possible, given the latest PHPStan releases. I'll take a look when I have time :)

Seldaek:
Thanks

PR #1766, opened 2022-10-07
Add an option to set the op_type for the ElasticsearchHandler

identei:
When pushing logs to a [data stream](https://www.elastic.co/guide/en/elasticsearch/reference/current/data-streams.html) in Elasticsearch, the [op_type](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#docs-index-api-op_type) must be set to "create". By default, the ElasticsearchHandler uses the "index" op_type. I have added an option to the ElasticsearchHandler so that you can specify the op_type to use and therefore enable the ability to push logs into an Elasticsearch data stream.

The default selection still is set to "index" so no behavior will change unless you specifically include the "op_type" option when creating the client.

- Elasticsearch Documentation: [Add documents to a data stream](https://www.elastic.co/guide/en/elasticsearch/reference/current/use-a-data-stream.html#add-documents-to-a-data-stream)


Seldaek:
Thanks

PR #1772, opened 2022-11-03
Docs: Add mention about level_name

o5:


94noni (approved):

Seldaek:
Hey, thanks for the PR but this is strictly speaking not available on the LogRecord class, at least not as a property you're supposed to call `$record->level->getName()`. It is accessible as `$record['level_name']` for BC with monolog 2 handlers etc but this isn't really the recommended way so I don't think it should be in the docs. Does that make sense?

PR #1773, opened 2022-11-15
Add support for enum context values to psr formatter

bart-mollie:
This PR adds support for Enums to the PsrLogMessageProcessor.

Currently Enums passed in the context display as `[object enum-class-name]`. This PR changes this to display the value for `BackedEnums` and displaying the name for non backed enums.


Seldaek:
Thanks, merged in 2.x as https://github.com/Seldaek/monolog/commit/8255feb4446728566db3bf537574d18bdf850472

PR #1774, opened 2022-11-15
Update message-structure.md

94noni:
As channel is a string, better be explicit as there is not trace of $channel


Seldaek:
Thanks

PR #1777, opened 2022-11-21
change config for PHPUnit

mimmi20:
This PR wants to update the config for phpunit to the format of the 9.x version to avoid the deprecation message


Seldaek:
Thanks

PR #1778, opened 2022-11-21
update dev dependencies

mimmi20:
This PR wants to update some dev dependencies to newer versions, and fix some issues reported by phpstan


Seldaek:
Thanks

PR #1779, opened 2022-11-21
add some @throws in constructors

mimmi20:
In some constructors Exceptions are thown, but not mentioned in the Docblocks. This results in phpstan error messages when I want to catch these Exceptions.


Seldaek:
Thanks

PR #1781, opened 2022-12-06
Add the ability to set or redefine attributes for AMQPMessage.

pmikle:
One of content_type, content_encoding, message_id, user_id, app_id, delivery_mode, priority, timestamp, expiration, type or reply_to, headers.


Seldaek:
Thanks

PR #1793, opened 2023-02-18
Update README.md

94noni:
Hi
Small PR as the warning icon, to me, is too much for the phrase it details, an info icon seems better
also leverages github note feature in markdown


Seldaek:
Thanks

PR #1797, opened 2023-03-22
Add support for priority to Monolog processors

sudo-plz:
As suggested in the [Monolog-bundle](https://github.com/symfony/monolog-bundle/pull/455#discussion_r1143066893).

So that the order in which the processors are used by monolog can be determined.


Seldaek:
Thanks

PR #1802, opened 2023-04-09
TelegramBotHandler topics support

SamAsEnd:
The telegram API support sending message to a specific topic for supergroups.

This `message_thread_id` parameter wasn't added on #1451 because it didn't exist by that time.

https://core.telegram.org/bots/api#sendmessage


Seldaek:
Thanks!

PR #1803, opened 2023-04-18
Add cpu usage processor

JohJohan:
I added a processor to get avg cpu usage. Should this be added here or should I set it up as third party processor.

If it may be merged here should i target the 3.x branch or should I target 1.x?

### DIFF
--- a/src/Monolog/Processor/CPUUsageProcessor.php
+++ b/src/Monolog/Processor/CPUUsageProcessor.php
@@ -24,7 +24,7 @@ class CPUUsageProcessor implements ProcessorInterface
     public const AVG_SYSTEM_LOAD_5_MINUTE = 1;
     public const AVG_SYSTEM_LOAD_15_MINUTE = 2;
 
-    public const AVAILABLE_AVG_SYSTEM_LOAD = [
+    private const AVAILABLE_AVG_SYSTEM_LOAD = [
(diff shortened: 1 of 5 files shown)


JohJohan:
@Seldaek is this something for this repository or should i make it third party processor?

src/Monolog/Processor/CPUUsageProcessor.php
```diff
@@ -0,0 +1,63 @@
...
+        }
+        $this->avgSystemLoad = $avgSystemLoad;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function __invoke(LogRecord $record): LogRecord
+    {
+        $usage = sys_getloadavg();
```
stof, line 54: If the function is not defined (on Windows for instance, according to the documentation), this should probably return the record unchanged instead triggering of a fatal error

JohJohan: Yes your right, at first i thought it would have been a developer mistake to implement this processor on a windows server but it might be better to handle it and return unchanged $record

Seldaek:
I renamed it because IMO it's clearer that it's the load average that way and not cpu usage which I would expect as a 0-100 percentage.

But otherwise looking good, thanks!

PR #1805, opened 2023-04-27
Add support yearly or monthly rotation log file

liutaot:
The function of adding yearly or monthly rotation log file. Although there are relevant definitions in the code before, it has not really supported it.

### DIFF
--- a/src/Monolog/Handler/RotatingFileHandler.php
+++ b/src/Monolog/Handler/RotatingFileHandler.php
@@ -43,13 +43,13 @@ class RotatingFileHandler extends StreamHandler
      * @param int|null $filePermission Optional file permissions (default (0644) are only for owner read/write)
      * @param bool     $useLocking     Try to lock log file before doing any writes
      */
-    public function __construct(string $filename, int $maxFiles = 0, int|string|Level $level = Level::Debug, bool $bubble = true, ?int $filePermission = null, bool $useLocking = false)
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/RotatingFileHandler.php
```diff
@@ -198,4 +198,26 @@ protected function getGlobPattern(): string
...
+    protected function setDateFormat(string $dateFormat): void
+    {
+        if ($dateFormat !== self::FILE_PER_DAY && $dateFormat !== self::FILE_PER_MONTH && $dateFormat !== self::FILE_PER_YEAR) {
+            throw new InvalidArgumentException(
+                'Invalid date format - format must be one of ' .
+                'RotatingFileHandler::FILE_PER_DAY ("Y-m-d"), RotatingFileHandler::FILE_PER_MONTH ("Y-m") ' .
+                'or RotatingFileHandler::FILE_PER_YEAR ("Y"), or you can set one of the ' .
+                'date formats using slashes, underscores and/or dots instead of dashes.'
+            );
+        }
```
Seldaek, line 211: This validation isn't complete, it should match what is in setFilenameFormat - but I think setFilenameFormat should just call this setDateFormat and then move the validation in here so it is not duplicated.

Seldaek:
Thanks for the fixes!

PR #1806, opened 2023-05-11
Feature/phpunit 10

Chris53897:
fix ci deprecations for checkout-action   v2 => v3

withConsecutive replacement from https://stackoverflow.com/a/75686291

please have  a closer look if my test replacements are correct.

phpstan-error is unrelated


Seldaek:
Thanks!

PR #1807, opened 2023-05-12
chore: use stable version of graylog2/gelf-php

Chris53897:
Version 2 is stable and released now
https://github.com/bzikarsky/gelf-php/releases

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "aws/aws-sdk-php": "^3.0",
         "doctrine/couchdb": "~1.0@dev",
         "elasticsearch/elasticsearch": "^7 || ^8",
-        "graylog2/gelf-php": "^1.7.1 || ^2.0",
+        "graylog2/gelf-php": "^1.4.2 || ^2.0",
         "guzzlehttp/guzzle": "^7.4.5",
         "guzzlehttp/psr7": "^2.2",
         "mongodb/mongodb": "^1.8",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         "aws/aws-sdk-php": "^3.0",
         "doctrine/couchdb": "~1.0@dev",
         "elasticsearch/elasticsearch": "^7 || ^8",
-        "graylog2/gelf-php": "^1.4.2 || ^2@dev",
+        "graylog2/gelf-php": "^1.7.1 || ^2.0",
```
fredden, line 24: What here is *not* compatible with versions `1.4.2` through `1.7.0`?

Chris53897: I did not say this. I can revert that if you want

fredden: The `^2@dev` --> `^2.0` change makes sense and is in scope for this pull request, but dropping support for versions `1.4.2` through `1.7.0` seems accidental / out of scope here.

> I did not say this.

The code says it. I don't understand why that change (`^1.4.2` --> `^1.7.1`) has been made. If the change was intentional, please can you explain why it's been made?

> I can revert that if you want

Yes, I think that makes sense. It seems like the change was accidental / unintentional.

Chris53897: Done. I think the latest previous major version should be responsible for forward compatibility, that is why i do this in my personal projects. But your are right, let us set the scope for stability here.

fredden (approved):

Seldaek:
Thanks

PR #1815, opened 2023-06-21
Fixed bug that `restore_error_handler` will not be executed some times.

limingxinleo:


limingxinleo:
https://github.com/hyperf/hyperf/issues/2583

stof (approved):

Seldaek:
I don't quite understand why they get "Warning: Invalid callback Monolog\Handler\StreamHandler::customErrorHandler, cannot access private method Monolog\Handler\StreamHandler::customErrorHandler()"

As far as I can tell the current code is fine and if it fails it does not throw so the try/catch shouldn't be needed? https://3v4l.org/NOrFI

limingxinleo:
@Seldaek Could you please take a look at this.

https://3v4l.org/e3L7U

Thanks.

Seldaek:
Ok I get it now, fopen throws somehow here, I don't know why it does for your users tho as it shouldn't but ok, then the error handler isn't properly reset and from an outside scope it fails. Thanks for the example. I'll merge this then.

PR #1819, opened 2023-07-07
Fix LogRecord "extra" data leaking between handlers

cosminardeleanu:
Context:
- Let's say i have 3 handlers attached to monolog logger: file handler, slack handler, elastic search handler.
  - Note that each handler can push its own processor, which can alter the LogRecord entity.
- On the elastic search handler, we need extra info, so push a processor to add them,  (stupid example: `$logRecord->extra['new_entry_just_for_elastic_search'] = 'stuff';`
  - This "extra" data i also available for file handler and slack handler.
  - Basically i will get in slack handler and file handler, this extra that was supposed to be only for elastic search handler.

Why this is happening?
- LogRecord is passed by reference, to each handler.
- LogRecord accepts to alter data from `extra`, so if one handlers manages to alter this object, rest of remaining handlers, will use the altered object.

### DIFF
--- a/src/Monolog/Handler/FallbackGroupHandler.php
+++ b/src/Monolog/Handler/FallbackGroupHandler.php
@@ -33,7 +33,7 @@ class FallbackGroupHandler extends GroupHandler
         }
         foreach ($this->handlers as $handler) {
             try {
-                $handler->handle($record);
+                $handler->handle(clone $record);
                 break;
             } catch (Throwable $e) {
                 // What throwable?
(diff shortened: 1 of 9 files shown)


cosminardeleanu:
@Seldaek ^^

aistis-:
I confirm the issue persists with `3.4.0`. Waiting for the merge 🤞

stof:
@Seldaek with the migration to an object in Monolog 3, this is indeed an issue for handler-level processors rather than logger-level processors. Should the logger clone the record before calling the handler ?

Seldaek:
Yeah I think it'd be better to clone.. both in Logger and in GroupHandler and related handlers which fan out records to nested handlers. Otherwise the problem is just shifted one problem lower.

stof:
indeed, fan-out handlers also need to clone.

Seldaek:
Alright I think with my commit it should take care of it for all cases. I hope I didn't forget any fan-out handler.

PR #1825, opened 2023-08-02
Skip mongodb extension if extension is not installed

lyrixx:
Without this patch, `MongoDBHandlerTest::testHandleWithLibraryClient()` test fails

---

**⚠ I'm really not sure about this PR!**


Seldaek:
Yeah sorry the test checked for library presence, but not extension, and the library requires the extension to be present anyway.

PR #1828, opened 2023-08-03
Use `@return $this` where applicable

nicolas-grekas:
Spotted while preparing https://github.com/Seldaek/monolog/pull/1827


Seldaek:
Thanks

gmponos:
Hi, I am asking for learing purposes. What is the need for the docblock when the function is typehinted with self? What are the differences?

PR #1831, opened 2023-08-24
Fix basic example in README.md

simoheinonen:


stof:
-1 for that. My guess is that you are using an older version of Monolog while looking at the readme of version 3

simoheinonen:
hmm yeah you're right!

PR #1833, opened 2023-09-08
Fix fatal error in NormalizeFormatter.

koekaverna:
> PHP Fatal error:  method_exists(): The script tried to execute a method or access a property of an incomplete object. Please ensure that the class definition "Monolog\TestClass" of the object you
>  are trying to operate on was loaded _before_ unserialize() gets called or provide an autoloader to load the class definition in /var/www/html/vendor/monolog/monolog/src/Monolog/Formatter/NormalizerFormatter.php on line 177
> 

When NormalizeFormatter use method_exists on [incomplete object](https://www.php.net/manual/en/function.unserialize.php) it cause fatal error.


Seldaek:
Thanks! Merged in 2.x & main

koekaverna:
@Seldaek Hi, it's not fixed in 2.9.2 JsonFormatter. There is normalize method override NormalizerFormatter::normalize and have same problem...

Issue #1787, opened 2023-01-16
Telegram Handler


deulizealand:
Monolog version 2.8

How to send Telegram logs to private channel or private group instead of public channel ?

deulizealand:
i found that we can use chat id of the private group or channel

claudiu-cristea:
@deulizealand did you prefix with "@"? What about the dash in front of the ID?

chimit:
Using a chat ID like `-1654567890` doesn't work for me: `RuntimeException  Telegram API error. Description: Bad Request: chat not found.`

@alexandrmazur96 is there any way to support private chats/channels?

PR #1840, opened 2023-09-23
Clarified instructions about chat ID for Telegram

chimit:
These instructions clarify the details of setting Telegram chat ID for different types of chats.

https://github.com/Seldaek/monolog/issues/1787


Seldaek:
Thanks

PR #1841, opened 2023-09-30
Loggly handler loglevel comparison statement fixed  Issue: #1838

### LINKED ISSUE #1838: LogglyHandler - incorrect records comparison.
Monolog version 3.4.0

Loggly does not receive all logs. The reason is an incorrect records comparison in LogglyHandler. 

[Source reference](https://github.com/Seldaek/monolog/blob/9546d94bdc212214e23aa7ea5a3734417e33e5e1/src/Monolog/Handler/LogglyHandler.php#L129)

Now the code is: 

   `return ($record->level >= $level);`

    
It looks like should be:
    
`return ($record->level->value >= $level->value);`

We should compare values properties of records.

omerimzali:
"The issue was explained quite nicely, and I have created the PR."

Fixes #1838

### DIFF
--- a/src/Monolog/Handler/LogglyHandler.php
+++ b/src/Monolog/Handler/LogglyHandler.php
@@ -126,7 +126,7 @@ class LogglyHandler extends AbstractProcessingHandler
         $level = $this->level;
 
         $records = array_filter($records, function ($record) use ($level) {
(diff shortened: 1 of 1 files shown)


Seldaek:
Makes sense, sorry for missing that one. Thanks for the fix!

PR #1845, opened 2023-10-19
Pass maxNormalizeDepth and maxNormalizeItemCount in NormalizerFormatter

BafS:
The current way to change maxNormalizeDepth and maxNormalizeItemCount is to extends the NormalizerFormatter. 

We should instead be able to change them using the constructor (and in the future those fields should be private IMO).


stof:
The actual current way to customize them is to call the setters.

BafS:
But why is `$dateFormat` passed in the constructor and not the other "settable" arguments? It seems inconsistent

PR #1847, opened 2023-10-24
Add a WithMonologChannel attribute

stof:
This attribute is meant to be used by frameworks / integrations to choose which logger instance to inject in a class when they manage several channels, if they decide to use it.
This attribute will have no effect in Monolog itself as the wiring of the logger in other classes is not managed by Monolog.

### DIFF
--- a/src/Monolog/Attribute/WithMonologChannel.php
+++ b/src/Monolog/Attribute/WithMonologChannel.php
@@ -1,0 +1,29 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog\Attribute;
+
+/**
+ * A reusable attribute to help configure a class as expecting a given logger channel.
+ *
+ * Using it offers no guarantee: it needs to be leveraged by a Monolog third-party consumer.
+ *
+ * Using it with the Monolog library only has no effect at all: wiring the logger instance into
+ * other classes is not managed by Monolog.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
+final class WithMonologChannel
+{
+    public function __construct(
+        public readonly string $channel
+    ) {
+    }
+}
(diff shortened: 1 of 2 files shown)


stof:
the phpstan failure is unrelated to my change. It is fixed in #1848

src/Monolog/Attribute/WithMonologChannel.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * A reusable attribute to help configure a class as expecting a given logger channel.
+ *
+ * Using it offers no guarantee: it needs to be leveraged by a Monolog third-party consumer.
+ *
+ * Using it with the Monolog library only has no effect at all: wiring the logger instance into
+ * other classes is not managed by Monolog.
+ */
+#[\Attribute(\Attribute::TARGET_CLASS)]
```
nicolas-grekas, line 22: Should be allow it on parameters also, and possibly properties?

nicolas-grekas (commented):
Makes sense to me, next to AsMonologProcessor.

src/Monolog/Attribute/WithMonologChannel.php
stof: This won't work in MonologBundle. The tag would apply to the whole service. It does not support separate tags per parameter.
If you want to inject 2 different loggers in the same class with autowiring, you will need to rely on the named aliases or `#[Target]`

nicolas-grekas: Couldn't we turn the attribute into eg a binding in the bundle?

stof: btw, that's also why I did not make the attribute repeatable.

Seldaek:
Yeah I guess why not have this here as well

src/Monolog/Attribute/WithMonologChannel.php
stof: binding are not per-attribute but per service.

stof:
@Seldaek do you know when the 3.5 release will be done with this attribute included ? I'm adding support for this attribute in monolog-bundle.

Seldaek:
I'll try to do this soon but it might be a week or two.. Would like to go through all open stuff as it's been a while

Seldaek:
Ok 3.5.0 is out now https://github.com/Seldaek/monolog/releases/tag/3.5.0

PR #1848, opened 2023-10-24
Remove usage of soft-deprecated PHPUnit API

stof:
This method is not yet reported by PHPUnit as a warning. But as it has the `@deprecated` annotation, it gets reported by the phpstan setup as it detects usages of deprecated APIs.


Seldaek:
Thanks

Issue #1785, opened 2022-12-21
Level_name options


funtus00:
It would be great to have optional level_name name. For example NLog has feature to make level_name as 1, 3 letters or full name.
Current situation (log strings):

[2022-21-12 10:10:01.718][INFO][127.0.0.1] '6': user login
[2022-21-12 10:10:01.724][WARNING][127.0.0.1] '6': domain gmail.com is forbidden to use or does not exist  
[2022-21-12 10:10:01.724][ERROR][127.0.0.1] '6': user email is forbidden to use  

If set it level_name as 1 or 3 letters logs can be like this:

[2022-21-12 10:10:01.718][INF][127.0.0.1] '6': user login
[2022-21-12 10:10:01.724][WRN][127.0.0.1] '6': domain gmail.com is forbidden to use or does not exist  
[2022-21-12 10:10:01.724][ERR][127.0.0.1] '6': user email is forbidden to use  

[2022-21-12 10:10:01.718][I][127.0.0.1] '6': user login
[2022-21-12 10:10:01.724][W][127.0.0.1] '6': domain gmail.com is forbidden to use or does not exist  
[2022-21-12 10:10:01.724][E][127.0.0.1] '6': user email is forbidden to use  

It helps to read logs.
Idea do make option:
INFO = INF = I
ERROR = ERR = E
WARNING = WRN = W
DEBUG = DBG = D
and so on...

Seldaek:
So this would be for LineFormatter only right? Probably could easily be added as an additional option via a `setMaxLevelNameLength(1)`?

Seldaek:
Fixed by https://github.com/Seldaek/monolog/pull/1850

PR #1850, opened 2023-10-26
Added possibility to set max length for level name in LineFormatter

philipp-check24:
Fixes https://github.com/Seldaek/monolog/issues/1785


Seldaek:
Thanks, I did tweak the implementation a bit to use a setter instead of constructor as the constructor is already quite loaded and this is not really a major use case.

PR #1852, opened 2023-10-30
Include extra in context in PsrHandler

snapshotpl:
Currently PsrHandler pass only context, but missing all extra data. This PR adds option to merge this two arrays and pass as context into psr logger.


Seldaek:
Thanks

PR #1854, opened 2023-11-09
Main

theking2:


Seldaek:
Thanks but I do not accept new handlers like this which are not standalone and generic. Please submit it as a third party package and list it here if you like: https://github.com/Seldaek/monolog/wiki/Third-Party-Packages

PR #1855, opened 2023-11-13
AbstractRotatingFileHandler

theking2:
AbstractRotatingFileHanderl dealing with write, close, reset
Refactor RotatingFileHandler to implement the abstract


Seldaek:
Please publish this new handler as a standalone package instead of extending this, I have no interest in supporting cron expressions, logrotate is the correct way to do this IMO.

PR #1858, opened 2023-11-22
catching exceptions in exception handling

Henning256:
When an Exception is thrown while logging that has a non numberic code another exception is thrown while formating the exception. This is for example the case when logging to a database and a PDOException is thrown.

[object] (ErrorException(code: 0): Undefined property: Illuminate\\Database\\QueryException::$code

### DIFF
--- a/src/Monolog/Formatter/LineFormatter.php
+++ b/src/Monolog/Formatter/LineFormatter.php
@@ -240,7 +240,12 @@ class LineFormatter extends NormalizerFormatter
 
     private function formatException(\Throwable $e): string
     {
-        $str = '[object] (' . Utils::getClass($e) . '(code: ' . $e->getCode();
+        $str = '[object] (' . Utils::getClass($e) . '(code: ';
+        try {
+            $str .= $e->getCode();  // throws an exception when $code is a string like in PDOException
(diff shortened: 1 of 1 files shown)


stof:
`$e->getCode` is not supposed to throw an exception though. This should be reported to Laravel so that they fix their QueryException.

Henning256:
This is not a laravel problem. It is well known, that the PDOException which the QueryException extends causes this problem.
Look for example in many user comments of the official PHP documentation for the PDOException:
[The PDOException class @ php.net](https://www.php.net/manual/en/class.pdoexception.php)

I already tried to handle it in laravels QueryException but it extends PDOException and PDOException::getCode() is final.
There seems to be no known way to handle it in an extended PDOException. All solutions seem to base on the idea of catching the PDOException and raise a custom exception.

This all in mind it seems much more elegant to fix it here instead of everywhere a PDOException might occur.

stof:
@Henning256 `PDOException::getCode` does not throw an exception (and especially not one saying `Undefined property`).

Henning256:
Thank you stof!
I can not reproduce outside of LineFormater. Very strange.

Closing for further investigation.

PR #1859, opened 2023-12-02
Remove `final` from class declaration

felds:
Most (if not all) other classes in the project are not marked as final.

Although it's possible to extend the functionality of this class by using composition, it goes against how the rest of the library works for no apparent reason.

In my use case, I was trying to add a field linking the log entry to a request [as described in the Cloud Logging documentation](https://cloud.google.com/run/docs/logging#correlate-logs).


Seldaek:
Sure, thanks

PR #1868, opened 2024-01-22
[NormalizerFormatter] Avoid throwing an exception when the toString call fails

VincentLanglet:
HI @Seldaek

This is an improvement of https://github.com/Seldaek/monolog/issues/673

When a __toString method call throws an exception (which should occurs as few as possible), the Normalizer is throwing an exception too. I believe this behavior is wrong because:
- I cannot say monolog to prefer the default behavior over the __toString method
- If I wouldn't have a __toString method, it would have work correctly

Also, I have the following situation when using DoctrineFixtureBundle:
- I wrote a bug in my fixtures
- An exception is thrown
- During this process, monolog try to do some logs
- For the logs, he look for FooObject::__toString() method
- This method return the name of the FooObject which require a database request (because of doctrine lazy proxy)
- The entity is not found in database (of course, since the fixture loading failure)
- The __toString throws an exception
- Monolog crash
- I get the monolog failure instead of the doctrineFixture one to debug correctly my work.

As soon as I comment the toString method in my entity, I get the correct error.
Imho, if monolog cannot use the toString method, he shouldn't change the "error behavior" and fallback to jsonEncode like the tostring method doesn't exist.

WDYT ?

### DIFF
--- a/src/Monolog/Formatter/NormalizerFormatter.php
+++ b/src/Monolog/Formatter/NormalizerFormatter.php
@@ -210,8 +210,14 @@ class NormalizerFormatter implements FormatterInterface
                 $accessor = new \ArrayObject($data);
                 $value = (string) $accessor['__PHP_Incomplete_Class_Name'];
             } elseif (method_exists($data, '__toString')) {
-                /** @var string $value */
-                $value = $data->__toString();
+                try {
+                    /** @var string $value */
+                    $value = $data->__toString();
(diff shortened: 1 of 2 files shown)


VincentLanglet:
Friendly ping @Seldaek if you find some time to take a look :)

Seldaek:
Thanks yeah probably makes sense to avoid failures.

PR #1869, opened 2024-01-22
Convert GelfMessage into string required by createAmqpMessage

syssi:
This is just an idea how to make the GelfMessageFormatter and AmqpHandler compatible (again).

### DIFF
--- a/src/Monolog/Handler/AmqpHandler.php
+++ b/src/Monolog/Handler/AmqpHandler.php
@@ -11,6 +11,7 @@
 
 namespace Monolog\Handler;
 
+use Gelf\Message as GelfMessage;
 use Monolog\Level;
 use Monolog\Formatter\FormatterInterface;
 use Monolog\Formatter\JsonFormatter;
@@ -116,6 +117,10 @@ class AmqpHandler extends AbstractProcessingHandler
 
             $record = $this->processRecord($record);
             $data = $this->getFormatter()->format($record);
+
+            if($data instanceof GelfMessage) {
+                $data = json_encode($data->toArray());
+            }
 
             $this->exchange->batch_basic_publish(
                 $this->createAmqpMessage($data),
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/AmqpHandler.php
```diff
@@ -117,6 +118,10 @@ public function handleBatch(array $records): void
             $record = $this->processRecord($record);
             $data = $this->getFormatter()->format($record);
 
+            if($data instanceof GelfMessage) {
+                $data = json_encode($data->toArray());
+            }
```
Seldaek, line 123: This probably should be done in write() as well? Otherwise only handleBatch is supported, but not handling a single log record.

Seldaek: But I do kinda wonder what the point is of this PR. Shouldn't GelfMessageFormatter be used with GelfHandler rather? Why use AMQP? And if you need AMQP why not simply use a JsonFormatter or LineFormatter?

syssi: We prefer using the GelfMessageFormatter because of all the goodies of the implementation and tried to avoid implementing the specs a second time.

We have a local RabbitMQ broker per host and forward all log messages to a central server. So we don't talk to Graylog directly to keep the latency (`PHP->monolog->amqp->local RabbitMQ`) as low as possible.

Seldaek: Ok fair enough, but anyway `write()` should still be fixed IMO for completeness

syssi: Good point! I will have a look.

syssi: Done.

Seldaek:
Thanks!

PR #1872, opened 2024-03-15
fix: Implicitly marking parameter $datetime as nullable is deprecated…

thorsten:
… on PHP 8.4


Seldaek:
Thanks but it's been covered by another PR targetting 2.x and fixing all of the nullable types

PR #1877, opened 2024-04-02
Trace Context for GoogleCloudLoggingFormatter

iamacarpet:
Add Cloud Trace context to log entries formatted with GoogleCloudLoggingFormatter.

This allows log entries to be grouped with the HTTP request that generated them in GCP's Log Explorer:

Screenshot 2024-03-28 7 20 16 PM

@sl0wik @sonnysasaka - two big users of PHP on GCP, RFC please.

### DIFF
--- a/src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
+++ b/src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
@@ -24,6 +24,10 @@ use Monolog\LogRecord;
  */
 final class GoogleCloudLoggingFormatter extends JsonFormatter
 {
+    const CONTEXT_HEADER_FORMAT = '/([0-9a-fA-F]{32})(?:\/(\d+))?(?:;o=(\d+))?/';
+
+    private static ?string $traceID = null;
+
     protected function normalizeRecord(LogRecord $record): array
     {
         $normalized = parent::normalizeRecord($record);
@@ -32,9 +36,49 @@ final class GoogleCloudLoggingFormatter extends JsonFormatter
         $normalized['severity'] = $normalized['level_name'];
         $normalized['time'] = $record->datetime->format(DateTimeInterface::RFC3339_EXTENDED);
 
+        // Tag with Trace ID for request attribution
+        $normalized['logging.googleapis.com/trace'] = $this->getTraceID();
+
         // Remove keys that are not used by GCP
         unset($normalized['level'], $normalized['level_name'], $normalized['datetime']);
 
         return $normalized;
+    }
+
+    private function getTraceID(): ?string
+    {
+        if (empty($this->traceID) && !empty($_SERVER['HTTP_X_CLOUD_TRACE_CONTEXT'])) {
+            $matched = preg_match(
+                self::CONTEXT_HEADER_FORMAT,
+                $_SERVER['HTTP_X_CLOUD_TRACE_CONTEXT'] ?? '',
+                $matches,
+            );
+
+            if (!$matched) {
+                return null;
+            }
+
+            $projectID = $this->getProjectID();
+            if (empty($projectID)) {
+                return null;
+            }
+
+            $this->traceID = 'projects/'.$projectID.'/traces/'.strtolower($matches[1]);
+        }
+
+        return $this->traceID;
+    }
+
+    private function getProjectID(): ?string
+    {
+        if (isset($_SERVER['GOOGLE_CLOUD_PROJECT'])) {
+            return $_SERVER['GOOGLE_CLOUD_PROJECT'];
+        }
+
+        if (class_exists('\Google\Cloud\Core\Compute\Metadata')) {
+            return (new \Google\Cloud\Core\Compute\Metadata())->getProjectId();
+        }
+
+        return null;
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -45,6 +45,7 @@
         "ext-mongodb": "Allow sending log messages to a MongoDB server (via driver)",
         "mongodb/mongodb": "Allow sending log messages to a MongoDB server (via library)",
         "aws/aws-sdk-php": "Allow sending log messages to AWS services like DynamoDB",
+        "google/cloud": "Allow inclusion of additional contextual information on GCP",
         "rollbar/rollbar": "Allow sending log messages to Rollbar",
         "ext-mbstring": "Allow to work properly with unicode symbols",
         "ext-sockets": "Allow sending log messages to a Syslog server (via UDP driver)",


sonnysasaka:
Not sure if we need this support directly at monolog level, but I've been able to use https://github.com/googleapis/google-cloud-php-logging, which is a PSR logger which is supported as a "Handler" by monolog, e.g. in this laravel snippet:
```
<?php
namespace App\Logging;

use Google\Cloud\Logging\LoggingClient;
use Monolog\Logger;
use Monolog\Handler\PsrHandler;

class CreateGcpLogger
{
    /**
     * Create a custom Monolog instance.
     *
     * @param  array<string>  $config
     * @return \Monolog\Logger
     */
    public function __invoke(array $config)
    {
        $loggingClient = new LoggingClient([
            'projectId' => env('GOOGLE_PROJECT_ID'),
        ]);

        $logger = $loggingClient->psrLogger('app', [
            'batchEnabled' => true,
        ]);

        return new Logger('gcp', [new PsrHandler($logger)]);
    }
}
```

And this already includes the trace id to each log entry: https://github.com/googleapis/google-cloud-php-core/blob/686ffd0bb1328cd2ee963c4444e2ab4b111ba3ba/src/Report/GAEMetadataProvider.php#L39

This works out of the box for app engine. For cloud run you might want to look at `CloudRunMetadataProvider`: https://github.com/googleapis/google-cloud-php-core/blob/686ffd0bb1328cd2ee963c4444e2ab4b111ba3ba/src/Report/CloudRunMetadataProvider.php

So this could be an alternative, if monolog prefers to stay agnostic towards formatting and let other modules define their own format to be used by monolog.

iamacarpet:
Thanks @sonnysasaka ,

This came about from a conversation with @sl0wik who said his Firevel project was using this:

```
LOG_CHANNEL: stderr
LOG_STDERR_FORMATTER: Monolog\Formatter\GoogleCloudLoggingFormatter
```

in the environmental configuration for Laravel on App Engine Standard.

This method of logging more closely matches what is desirable in a containerised environment (logging via `stdout`/`stderr`) and works on PHP 7.4+ thanks to the PHP-FPM configuration option:

```
decorate_workers_output = no
```

Although I was concerned he & his users were missing out on logs being tagged together against the request that generated them, which our developers consistently say is their favourite developer friendly feature of GCP.

I believe that `Google\Cloud\Logging\LoggingClient` works by submitting to the Logs API endpoint directly, sped up by the Batch Daemon, which is only available on App Engine Flexible.

Using the `stderr` logging method along with this formatter allows for something that works across App Engine Standard, App Engine Flexible, Cloud Run, GKE and (potentially) GCE, without any latency associated with calling the Logs API directly.

Personally, we have our own log formatter as part of our GaeSupportLaravel package, however, adding the required stuff into here (since exists anyway) will hopefully bring this functionality to more people.

sonnysasaka:
Thanks for providing more context. With that, I think it makes sense for the google cloud formatter to live in monolog repo itself. I don't have a particular concern with the code except one small suggestion inline in the patch.

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
```diff
@@ -32,9 +36,49 @@ protected function normalizeRecord(LogRecord $record): array
...
+        return $this->traceID;
+    }
+
+    private function getProjectID(): ?string
+    {
+        if (isset($_SERVER['GOOGLE_CLOUD_PROJECT'])) {
+            return $_SERVER['GOOGLE_CLOUD_PROJECT'];
+        }
+
+        if (class_exists('\Google\Cloud\Core\Compute\Metadata')) {
```
sonnysasaka, line 78: maybe we should add this optional dependency in `composer.json`'s `suggest`  entry (https://getcomposer.org/doc/04-schema.md#suggest)? That way user can know which package to include to get this metadata magic,

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
```diff
@@ -32,9 +36,49 @@ protected function normalizeRecord(LogRecord $record): array
...
+            return (new \Google\Cloud\Core\Compute\Metadata())->getProjectId();
```
sonnysasaka, line 79: also, it looks like the implementation of `Google\Cloud\Core\Compute\Metadata::getProjectId()` [does not do caching](https://github.com/googleapis/google-cloud-php-core/blob/686ffd0bb1328cd2ee963c4444e2ab4b111ba3ba/src/Compute/Metadata/Readers/HttpHandlerReader.php#L51), so maybe it's better to retrieve the project id just once at the constructor of this `GoogleCloudLoggingFormatter` as not to incur HTTP round trip every time there is a new log entry to format.

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
iamacarpet: Good idea, I'll add it now.

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
iamacarpet: You'll hopefully see further up, we set the static property `$this->traceID` and once set, we always return that static value.

The project ID is only retrieved to construct that value the single time it is constructed, so hopefully this is already accounted for.

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
iamacarpet: Done, how is this?:

https://github.com/Seldaek/monolog/commit/05ae45ff4e0360963ab08c7d29372b2ad9f34a60#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R48

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
sonnysasaka: Caching `$this->traceID` may be fine if it's guaranteed that `GoogleCloudLoggingFormatter` is always created new for each request. I think this is the case with PHP app with nginx + php-fpm but may not be the case with other kind of deployment (like with Laravel Octane where objects are created just once when the app starts and reused across multiple requests).

I am not familiar enough with monolog + octane to confidently say whether Octane reuses the formatter object across different requests, but to be safe I would not cache the trace id but only the project id.

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
```diff
@@ -32,9 +36,49 @@ protected function normalizeRecord(LogRecord $record): array
         $normalized['severity'] = $normalized['level_name'];
         $normalized['time'] = $record->datetime->format(DateTimeInterface::RFC3339_EXTENDED);
 
+        // Tag with Trace ID for request attribution
+        $normalized['logging.googleapis.com/trace'] = $this->getTraceID();
```
Seldaek, line 40: Should this really be set at all if getTraceID returns null?

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
Seldaek, line 78:
```suggestion
        if (class_exists('Google\Cloud\Core\Compute\Metadata')) {
```

src/Monolog/Formatter/GoogleCloudLoggingFormatter.php
Seldaek: Yeah it's probably safer not to cache, as retrieving the trace id doesn't seem that expensive.

Seldaek: And I would definitely not cache this as a static property. Cache on an instance property then a new instance would at least refetch from a clean state. Static property will definitely cause issues in some envs.

Issue #1878, opened 2024-04-05
User definable duplication for DeduplicationHandler


cracksalad:
The DeduplicationHandler defines duplication as multiple records having the same message and level. In general this is probably the right way to do it, but in specific cases it might be required to include some other attributes like `context`, `extra` or `channel` in this condition as well. 
Therefore you would want to overwrite the `DeduplicationHandler::isDuplicate()` method to define duplication by yourself. This is currently impossible due to the method being `private`. That is why I suggest to change the visibility to `protected`.
Additionally it would probably be desirable to move the `duplicationStore` stuff to another method and pass the `$store` variable to the `isDuplicate()` method. Just to make it a bit simpler to overwrite the method.

Of course one could also (as a workaround) add the additional attributes from say `context` to the message itself, but that would change the outcome in your logfiles or what have you.

If you approve this suggestion, I am willing to help and perform those changes by myself and create a PR afterwards. Just let me know!

PR #1879, opened 2024-04-07
[#1878] User definable duplication for DeduplicationHandler

cracksalad:
This is a draft to solve #1878 

Feel free to comment on the changes I made!

Although the tests ran through without modification, this change is not backwards compatible when you look at it closely, because it changes the signature of several methods (changed visibility from private to protected, add parameters etc.).

### DIFF
--- a/src/Monolog/Handler/DeduplicationHandler.php
+++ b/src/Monolog/Handler/DeduplicationHandler.php
@@ -70,7 +70,7 @@ class DeduplicationHandler extends BufferHandler
 
         $gc = false;
         $store = null;
-        
+
         if (file_exists($this->deduplicationStore)) {
             $store = file($this->deduplicationStore, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
             if (is_array($store)) {
@@ -83,7 +83,7 @@ class DeduplicationHandler extends BufferHandler
                 }
             }
         }
-        
+
         $passthru = null;
 
         foreach ($this->buffer as $record) {
@@ -129,11 +129,12 @@ class DeduplicationHandler extends BufferHandler
 
         return false;
     }
-    
+
     /**
      * @return string The given record serialized as a single line of text
      */
-    protected function buildDeduplicationStoreEntry(LogRecord $record): string {
+    protected function buildDeduplicationStoreEntry(LogRecord $record): string
+    {
         return $record->datetime->getTimestamp() . ':' . $record->level->getName() . ':' . preg_replace('{[\r\n].*}', '', $record->message);
     }
 
--- a/src/Monolog/Handler/DeduplicationHandler.php
+++ b/src/Monolog/Handler/DeduplicationHandler.php
@@ -43,7 +43,6 @@ class DeduplicationHandler extends BufferHandler
     protected Level $deduplicationLevel;
 
     protected int $time;
-    
     protected bool $gc = false;
 
     /**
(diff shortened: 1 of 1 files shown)


src/Monolog/Handler/DeduplicationHandler.php
```diff
@@ -70,13 +68,33 @@ public function flush(): void
...
+        if (file_exists($this->deduplicationStore)) {
+            $store = file($this->deduplicationStore, FILE_IGNORE_NEW_LINES | FILE_SKIP_EMPTY_LINES);
+            if (is_array($store)) {
+                $yesterday = time() - 86400;
+                for ($i = 0; $i < count($store); $i++) {
+                    if (substr($store[$i], 0, 10) < $yesterday) {
+                        $gc = true;
+                        break;
+                    }
+                }
```
Seldaek, line 83: I would rather keep the gc property and have isDuplicate set it to true if an older entry is found, because the way this works here, if you override the file format in buildDeduplicationStoreEntry/isDuplicate, then you may break the GC functionality.

cracksalad: Fair point! I moved the GC trigger back to `isDuplicate()`, readded the property and added a comment to `isDuplicate()` which hopefully explains to overwriters that they should set `$this->gc`.

I hope the failing CI workflow is not a huge problem. I am not sure what to do about it, because my changes do not seem to be responsible.

Seldaek: Nope there's a mess on CI right now, I'm on it :)

src/Monolog/Handler/DeduplicationHandler.php
```diff
@@ -43,8 +43,8 @@ class DeduplicationHandler extends BufferHandler
     protected Level $deduplicationLevel;
 
     protected int $time;
-
-    private bool $gc = false;
+
```
Seldaek, line 46:
```suggestion

```

src/Monolog/Handler/DeduplicationHandler.php
```diff
@@ -70,13 +70,24 @@ public function flush(): void
             return;
         }
 
+        $gc = false;
```
Seldaek, line 73:
```suggestion
```

src/Monolog/Handler/DeduplicationHandler.php
```diff
@@ -70,13 +70,24 @@ public function flush(): void
...
             if ($record->level->value >= $this->deduplicationLevel->value) {
-                $passthru = $passthru === true || !$this->isDuplicate($record);
+                $passthru = $passthru === true || !is_array($store) || !$this->isDuplicate($store, $record);
                 if ($passthru) {
-                    $this->appendRecord($record);
+                    $line = $this->buildDeduplicationStoreEntry($record);
+                    file_put_contents($this->deduplicationStore, $line . "\n", FILE_APPEND);
+                    if (is_array($store)) {
+                        $store[] = $line;
+                    }
```
Seldaek, line 90: Let's make sure the store becomes active here to avoid sending duplicate lines into it if the buffer contains multiple of the same log records.

```suggestion
                    if (!is_array($store)) {
                        $store = [];
                    }
                    $store[] = $line;
```

Seldaek (changes requested):
Thanks for the updates, found a couple nitpicks still

Seldaek:
Thanks!

Issue #1862, opened 2023-12-18
The file does not close after writing


sadakov:
Monolog version 3

We use Laravel and queues. Queue is maintained by supervisor.

The queue writes a log.  But after deleting the log, the task in the queue no longer writes to the log. All because the file does not close after writing (fclose).

Here src/Monolog/Handler/StreamHandler.php, method write, but then it doesn't close stream.

I tested, if I added a fclose($stream) at the end of the write method, the problem disappears.

kirkbushell:
@Seldaek I think this may be a good path forward regarding the broken pipe errors people are seeing (including us, as well - as of 2 months ago). Do you foresee any issues with the stream being closed at the end of each write?

Rydgel:
We have this issue as well in production, seems like file descriptors aren't getting closed as the issue appears more and more as the server uptime increase. Restarting the kube pod fix the issue for a while. So yeah there must be some missing fclose somewhere.

stof:
Monolog handlers have a concept of closing them. The StreamHandler will close the stream only when closing the handler (assuming you configured it in a way where the handler is the one opening the stream, not when passing it a stream managed externally that it could not re-open later).
Opening and closing the stream for each log message being written would quickly turn your logger as a performance bottleneck.

Rydgel:
Yeah, as far as I understand it should closes on `destruct`. There is most likely a leak somewhere though, can the destruct call get skipped for some reasons? (crash, long running processes, etc)

Seldaek:
I think maybe what we could try here is to close the stream when `reset()` is called, as that should be called between jobs in long running processes. That should ensure we don't keep a resource open for too long, assuming Laravel resets monolog correctly when needed.

cjavad:
This might have caused regressions when used in a reentrant context (swoole async etc), tbd.

Issue #1860, opened 2023-12-08
Issue with Logging to stdout in Laravel 10.21 using Monolog v3 EXCEPTION#Cannot log request: fwrite(): Write of 378 bytes failed with errno=32 Broken pipe


xuandung38:
Monolog version 3

Hello,

I'm currently working with Laravel 10.21 and have encountered an issue when attempting to write logs to stdout using Monolog version 3. My configuration is as follows:

```php
'stdout' => [
    'driver' => 'monolog',
    'handler' => StreamHandler::class,
    'formatter' => '\Monolog\Formatter\JsonFormatter',
    'with' => [
        'stream' => 'php://stdout',
    ],
],

```

However, this setup results in an exception:

```ruby
Exception: /var/www/html/vendor/monolog/monolog/src/Monolog/Handler/StreamHandler.php::162 -> EXCEPTION#Cannot log request: fwrite(): Write of 378 bytes failed with errno=32 Broken pipe

```

I noticed a similar issue previously reported as [#1634](https://github.com/Seldaek/monolog/issues/1634), but it seems the problem persists. Any guidance or suggestions to resolve this would be greatly appreciated.

Thank you for your support.

Best regards,

msobkowiak-olx:
It is because of this:

https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/StreamHandler.php#L118

is_resource is not enough to handle writing to a closed socket, it needs to also check if the socket is writable, as an example:
```
 $meta = stream_get_meta_data($fp);

 if(is_writable($meta['uri'])) { 
 // stream is writable
 }
 ```
 
 Or re-create $this->stream on fwrite() error here:
 https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/StreamHandler.php#L149

ls-tyler-roach:
When can we anticipate a fix to this bug?

Seldaek:
See also comments on #1634 - I'm still waiting for a way to narrow this down.

msobkowiak-olx:
@Seldaek thanks.

This is really the same class of problem like programming over the network (although the pipe is a local object, it can be closed by the OS at any time really) - closing the pipe on the "remote" side (so the side external to the library execution context) is not detectable until you try to write to the pipe. The explanation of this can be read here - there is not much we can do about it really as it is all coded in the kernel:

https://stackoverflow.com/questions/8369506/why-does-sigpipe-exist

Seldaek:
[...]

Issue #1634, opened 2022-02-11
ErrorException: Caught ErrorException (HTTP/1.0 500 Internal Server Error): fwrite(): write of X bytes failed with errno=32 Broken pipe


ehginanjar:
Monolog version 2

Got these lot of errors when sudden burst traffic hit the service. Even though the request end up success with HTTP 200, unfortunately this make additional at least ~2s latency for every processed requests thus slowing down the service.

Stacktrace:
```
ErrorException: Caught ErrorException (HTTP/1.0 500 Internal Server Error): fwrite(): write of 111 bytes failed with errno=32 Broken pipe in /var/www/app/vendor/monolog/monolog/src/Monolog/Handler/StreamHandler.php:178
#0 [internal function]: Laravel\Lumen\Application->Laravel\Lumen\Concerns\{closure}()
#1 /var/www/app/vendor/monolog/monolog/src/Monolog/Handler/StreamHandler.php(178): fwrite()
#2 /var/www/app/vendor/monolog/monolog/src/Monolog/Handler/StreamHandler.php(162): Monolog\Handler\StreamHandler->streamWrite()
#3 /var/www/app/vendor/monolog/monolog/src/Monolog/Handler/AbstractProcessingHandler.php(48): Monolog\Handler\StreamHandler->write()
#4 /var/www/app/vendor/monolog/monolog/src/Monolog/Logger.php(327): Monolog\Handler\AbstractProcessingHandler->handle()
#5 /var/www/app/vendor/monolog/monolog/src/Monolog/Logger.php(565): Monolog\Logger->addRecord()
#6 /var/www/app/vendor/illuminate/log/Logger.php(174): Monolog\Logger->error()
#7 /var/www/app/vendor/illuminate/log/Logger.php(87): Illuminate\Log\Logger->writeLog()
#8 /var/www/app/vendor/illuminate/log/LogManager.php(543): Illuminate\Log\Logger->error()
#9 /var/www/app/vendor/laravel/lumen-framework/src/Exceptions/Handler.php(56): Illuminate\Log\LogManager->error()
#10 /var/www/app/app/Exceptions/Handler.php(51): Laravel\Lumen\Exceptions\Handler->report()
#11 /var/www/app/vendor/laravel/lumen-framework/src/Concerns/RegistersExceptionHandlers.php(104): App\Exceptions\Handler->report()
[...]

PR #1882, opened 2024-04-15
Close and reopen file handles if a write fails

Seldaek:
Fixes #1862, fixes #1860, fixes #1634

PR #1883, opened 2024-04-15
Close file handle after each write

Seldaek:
refs #1862, refs #1860, refs #1634

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -156,6 +156,36 @@ class StreamHandler extends AbstractProcessingHandler
     }
 
     /**
+     * {@inheritDoc}
+     */
+    public function handle(LogRecord $record): bool
+    {
+        $result = parent::handle($record);
+
+        // close the resource if possible to reopen it after we are done writing
+        if ($this->url !== null && $this->url !== 'php://memory') {
+            $this->close();
+        }
+
+        return $result;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function handleBatch(array $records): void
+    {
+        foreach ($records as $record) {
+            parent::handle($record);
+        }
+
+        // close the resource if possible to reopen it after we are done writing
+        if ($this->url !== null && $this->url !== 'php://memory') {
+            $this->close();
+        }
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


rossmckelvie:
Thank you for making this fix, happy to share it's working for us. We deployed this fix to production today and its resolved hangups we were seeing on `fwrite()` leading to requests taking 10+ seconds.

eerison:
Hello @Seldaek 

as it was found on version 2.x, couldn't it be released  on version 2.x? this way we could still have this issue solved on php 8.0 and 8.2, and make a smooth php upgrade :)

Seldaek:
There is https://github.com/Seldaek/monolog/pull/1882 for 2.x, but the problem is nobody has confirmed this fixes anything for them so far, so I'd rather not merge this for nothing as it does introduce overhead for everyone.

ChipZ91:
Unfortunately we have the same error despite the change.

ivanmartinvalle:
For reference this is how we applied the workaround directly without needing upstream changes. The project in question is on monolog 1.

```php
class ClosingStreamHandler extends StreamHandler
{
    protected function streamWrite($stream, array $record)
    {
        parent::streamWrite($stream, $record);

        // https://github.com/Seldaek/monolog/pull/1883
        parent::close();
    }
}
```

ChipZ91:
I have to correct myself. Unfortunately, we have the error twice in our system. Once through monolog and once via symfony/process.

I can confirm that the fix here fixes the monolog case.

Thank you for the work and @ivanmartinvalle for the information that helped me understand our case.

ChipZ91 (approved):
The fix is working for us.

Seldaek:
Looking at this again, I don't think I want to merge as is.. Maybe it would be ok if we can detect the fwrite failure and reopen the file then, instead of closing every time. That'd have less impact on users that are not affected by the problem.

Seldaek:
See https://github.com/Seldaek/monolog/pull/1882 for the new version of the patch with the retry, I think that should help still, and I'm a lot more comfortable merging this.

Seldaek:
Closing in favor of https://github.com/Seldaek/monolog/pull/1882 which is now merged in 2.x-dev and 3.x-dev

PR #1884, opened 2024-04-15
Optimistically recreate streams when write fails 

msobkowiak-olx:
To address https://github.com/Seldaek/monolog/issues/1860 - when fwrite() fails - only then recreate the stream resource.


Seldaek:
This will not work IMO as the failure to write might be turned into an exception by some frameworks/apps, and also the failure is detected, file reopened on next log, but you still lose one log entry, so not great.

PR #1886, opened 2024-04-19
add formatted key in ArrayAccess for LogRecord

lekoala:
since $record['formatted'] is a legal syntax, it should not be flagged by static analysis tools like phpstan


Seldaek:
Right, will merge for correctness tho I cannot remember why I did not include this field, I hope it was just a mistake and not intentional. On the other hand you really should use ->formatted now IMO.

PR #1887, opened 2024-04-30
Use fully-qualified name for native functions

dreis2211:
Hi,

as this is my first PR in this project and I have't found a contributing guideline, please let me know in case you need anything from me.

This PR makes use of fully-qualified names for native functions - aka prefixing them with `\` - to allow optimizations through the engine. This is definitely a micro-optimization and I haven't measured it here, but it's a relatively common optimization in the PHP world these days, so I hope this is considered. E.g see https://php.watch/articles/php-zend-engine-special-inlined-functions 

Cheers,
Christoph


Seldaek:
As much as I hate these `\`s, I guess it makes sense to do it in this project, thanks.

PR #1894, opened 2024-06-15
Update example in RedisHandler

javaDeveloperKid:
The third argument to RedisHandler is a log level, not `prod` string.


Seldaek:
Thanks wtf happened there :D

PR #1899, opened 2024-07-21
Make IntrospectionProcessor extendable

Sekiphp:
Currently is very hard to make IntrospectionProcessor extendable from custom implementation (we have custom features implemented for Magento logging).


Seldaek:
Thanks

PR #1903, opened 2024-08-02
Fix PHP 8.4 deprecation notices

SerafimArts:
Add `null` type hint for nullable method parameters.

See also https://wiki.php.net/rfc/deprecate-implicitly-nullable-types


Seldaek:
Thanks

PR #1904, opened 2024-08-06
Replace goo.gl URL

Seb33300:
Google URL Shortener links will no longer be available: https://developers.googleblog.com/en/google-url-shortener-links-will-no-longer-be-available/


stof (approved):

Seldaek:
Thanks

PR #1906, opened 2024-08-11
Normalize call normalizeRecord if necessary

johndodev:
Monolog version 3

Hello,

When you format a logRecord with JsonFormater::format(), it's calling : 
- NormalizeFormatter::format() 
- NormalizeFormatter::normalizeRecord() 

Then we have a nice formatted record, with exception datas, if any.

Now, if we try to format an array of records with JsonFormater::formatBatch() : 
- JsonFormater::formatBatchJson()
- JsonFormater::normalize()

Then, we lost the exception datas.

As I understand, formatBatch is a way to format array of records in an optimized way if possible, and the output should be the same as formatting each records in a foreach.

If this is the case, it may be a bug, and this PR fix it for me. 
I dont know if it's a problem for anything else than "exception datas" but this is this case wich make me found that.
For more context, I was writing a custom Handler, where I call `this->getFormatter()->formatBatch($records);` in `handleBatch()`.

### DIFF
--- a/src/Monolog/Formatter/JsonFormatter.php
+++ b/src/Monolog/Formatter/JsonFormatter.php
@@ -171,6 +171,10 @@ class JsonFormatter extends NormalizerFormatter
         }
 
         if (\is_object($data)) {
+            if ($data instanceof LogRecord) {
+                return $this->normalizeRecord($data);
+            }
+            
             if ($data instanceof \DateTimeInterface) {
                 return $this->formatDate($data);
             }


gnutix:
JFYI, the login of our application broke because of this change (the 2FA process writes some logs) : we have a custom JsonFormatter that had the following code : 

```php
if (isset($normalized['context']['exception'])) {
    $normalized = /* ... */;
}
```

Except now `$normalized['context']` can  be a `\stdClass`, so it threw an error "Cannot use object of type stdClass as array".
Thought you should know.

Seldaek:
I mean this is really a string of bad luck.. every change is a possible BC break given the right circumstances. I know it sucks when it hits you though.

gnutix:
True, though I don't think we're the only one to have a custom JsonNormalizer to add more contextual info to the logs. Seems to me like a very common use case, that's why I'm afraid others will be hit too.

Maybe just editing the release on github to add a warning / link to this conversation might do the trick ?

Seldaek:
Yeah this was always the case tho from format()'s perspective, only normalizeRecord now possibly returns objects too. I'll add a note but I am not sure how common this really is.. It is pretty nasty tho that isset would throw like that. I thought it was more forgiving.

PR #1911, opened 2024-09-04
Add AzureHandler to send logs to Microsoft Azure Monitor

damikael:


Seldaek:
Hey, sorry for the delay here but please submit this as a third party package on packagist.org and then add it here if you like https://github.com/Seldaek/monolog/wiki/Third-Party-Packages - I don't accept new handlers anymore which aren't very general purpose as it takes too much time to maintain.

src/Monolog/Handler/AzureHandler.php
```diff
@@ -0,0 +1,121 @@
...
+
+    /**
+     * @inheritDoc
+     */
+    public function write(LogRecord $record): void
+    {
+
+        // retrieve access_token
+        $url = "https://login.microsoftonline.com/" . $this->tenantId . "/oauth2/v2.0/token";
+        $postString = "
```
laurentmuller, line 71: You can replace string post field by an array as follow:

```php
 $postString = [
    'grant_type' => 'client_credentials',
    'scope' => 'https://monitor.azure.com//. default',
    'client_id' => $this->appId,
    'client_secret' => $this->appSecret 
];
```

You can then remove the  `CURLOPT_HTTPHEADER` option because PHP will automatically add the header `Content-Type` to `multipart/form-data`.

PR #1913, opened 2024-09-17
Allow the user to define a custom mode when `fopen`-ing a log file

phrfpeixoto:
I've included changes to both allow custom `fopen` modes, as well as change the visibility of the helper functions on this class.
I've made them final to keep the "do not touch" aspect, while still allowing subclasses to use them.

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -41,6 +41,8 @@ class StreamHandler extends AbstractProcessingHandler
     protected $filePermission;
     /** @var bool */
     protected $useLocking;
+	/** @var string */
+    protected $fileOpenMode = 'a';
     /** @var true|null */
     private $dirCreated = null;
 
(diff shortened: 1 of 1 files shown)


stof:
with the new argument, do you actually need to extend the class and call those protected functions ? If yes, why ? If no, I would not change their visibility as making them protected means they become part of the API that is covered by semver (and so they cannot be refactored to change their signature or be removed if the code does not need them anymore)

phrfpeixoto:
It doesn't. I've added that specifically to bring that discussion:
- If I change the method visibility I can override the class and solve my problem, so I don't need the extra parameter.
- If I add the extra parameter, I no longer need to override the class, so I don't need the method visibility.

Whichever change is less disruptive for the project is OK for me.

phrfpeixoto:
@stof Let me know what changes you would prefer.

phrfpeixoto:
@Seldaek The use case is having the file overwritten, instead of appended.

Seldaek:
Thanks

PR #1914, opened 2024-09-18
Remove Slim framework integration suggestion

N-M:
The suggested library (https://github.com/Flynsarmy/Slim-Monolog) only supports Slim 2, which has not been maintained for a long time. Monolog can be integrated with Slim v4 directly, without the need for any additional libraries.


Seldaek:
Thanks

PR #1919, opened 2024-10-23
Remove pointless json_encode availability checks

### LINKED ISSUE #1918: Pointless checks for availability of JSON extension (and @throws RuntimeException)
I just upgraded from Monolog version 2.9.3 to 3.7.0. Since I let PHPStan handle RuntimeExceptions as checked exceptions it now reports that I need to handle a RuntimeException whenever my code invokes `new JsonFormatter()`.

This is due to #1779 where the intention was to document all exceptions that are actually thrown. But I think in this case it's bogus because AFAICT this exception will never be thrown:

https://www.php.net/manual/en/json.installation.php states:

>  The JSON extension is bundled and compiled into PHP by default.

> As of PHP 8.0.0, the JSON extension is a core PHP extension, so it is always enabled.

I understand this as "it is enabled by default and there is no way to disable it since PHP 8.0".

Monolog requires PHP >= 8.1 since version 3.0.0, so in combination it means that no RuntimeException will ever be thrown due to a missing `json_encode` function. Right?

Please remove the unnecessary checks and @throws accordingly. I hate to add PHPStan ignores for stuff that never occurs.

InvisibleSmiley:
Fixes #1918.

### DIFF
--- a/src/Monolog/Formatter/ElasticaFormatter.php
+++ b/src/Monolog/Formatter/ElasticaFormatter.php
@@ -34,8 +34,6 @@ class ElasticaFormatter extends NormalizerFormatter
     /**
      * @param string  $index Elastic Search index name
      * @param ?string $type  Elastic Search document type, deprecated as of Elastica 7
-     *
-     * @throws \RuntimeException If the function json_encode does not exist
      */
     public function __construct(string $index, ?string $type)
     {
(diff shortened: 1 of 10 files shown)


Seldaek:
Why this? json ext can still be disabled AFAIK even tho it's enabled by default.

InvisibleSmiley:
> Why this? json ext can still be disabled AFAIK even tho it's enabled by default.

No. As I tried to explain it cannot be disabled anymore. If you don't trust my understanding of the official PHP documentation that I cited, maybe read https://php.watch/versions/8.0/ext-json.

Seldaek:
Ah sorry I missed the linked issue 🤦🏻‍♂️ Sounds good then, thanks. I missed that change in 8.0

PR #1924, opened 2024-11-13
Fix return type `normalizeException`

ro0NL:
I dont see how `normalizeException` can return a string.

### DIFF
--- a/src/Monolog/Formatter/NormalizerFormatter.php
+++ b/src/Monolog/Formatter/NormalizerFormatter.php
@@ -248,7 +248,7 @@ class NormalizerFormatter implements FormatterInterface
     }
 
     /**
-     * @return array<string, string|int|array<string|int|array<string>>>|string
+     * @return array<array-key, string|int|array<string|int|array<string>>>
      */
     protected function normalizeException(Throwable $e, int $depth = 0)
     {


src/Monolog/Formatter/JsonFormatter.php
```diff
@@ -216,7 +216,7 @@ protected function normalize(mixed $data, int $depth = 0): mixed
      * Normalizes given exception with or without its own stack trace based on
      * `includeStacktraces` property.
      *
-     * @return array<string, string|int|array<string|int|array<string>>>|string
+     * @return array<array-key, string|int|array<string|int|array<string>>>
      */
     protected function normalizeException(Throwable $e, int $depth = 0): array
```
ro0NL, line 221: hard `array` return type here even

Seldaek:
See the build failure. It is done like that to allow the child class (LineFormatter) to override this. It's very difficult to have strict types for NormalizerFormatter as it is extended for various different use cases.

src/Monolog/Formatter/JsonFormatter.php
```diff
@@ -216,7 +216,7 @@ protected function normalize(mixed $data, int $depth = 0): mixed
      * Normalizes given exception with or without its own stack trace based on
      * `includeStacktraces` property.
      *
-     * @return array<string, string|int|array<string|int|array<string>>>|string
+     * @return array<array-key, string|int|array<string|int|array<string>>>
```
ro0NL, line 219: keys might be non-string when max depth exceeds, and maybe in case of JsonSerializable

ro0NL:
@Seldaek got ya. updated.

ro0NL:
yeah ok. i dont feel like doing the phpstan bookkeeping. sorry.

we've  casted `JsonFormatter::normalize` on our side.

Seldaek:
Yeah that's fine by me.. I don't know which is best here tbh, we gotta ignore some errors somewhere, but maybe for lib consumers it would be better if we did it like you had and ignore the JsonFormatter extension error. But I guess your `array-key` change should be applied still right? Is `array-key` equivalent to `string|int`? I can look at doing the change and updating the baseline.

ro0NL:
psalm describes it as 
> array-key is the supertype (but not a union) of int and string.

so yes just a union in practice

Seldaek:
OK I think this works https://github.com/Seldaek/monolog/commit/e94000419394ff1bec801dad310432228f9fc19c

PR #1928, opened 2024-11-21
Deprecate DateTimeImmutable

### LINKED ISSUE #1926: Rename `Monolog\DateTimeImmutable` to something else, e.g. `JsonSerializableDateTimeImmutable`
First of all, thanks for this amazing package 💙 

So this is something that I'm bothered by for a long time. Every time I want to import `DateTimeImmutable`, PHPStorm asks me which one I want to import:

Screenshot 2024-11-21 at 12 14 13@2x

Even though the version of Monolog is namespaced, it's still annoying to think which one to pick.

I want to propose the following:

* Rename `Monolog\DateTimeImmutable` to something different, for example: `Monolog\JsonSerializableDateTimeImmutable`
* Keep the old `Monolog\DateTimeImmutable` as an alias for `Monolog\JsonSerializableDateTimeImmutable`, and mark it as deprecated.
* In the next major, remove `Monolog\DateTimeImmutable`

Or if there is a better way to do this, I'm all ears. I would be happy to create the PR.

/cc @Seldaek

ruudk:
Fixes #1926

### DIFF
--- a/src/Monolog/JsonSerializableDateTimeImmutable.php
+++ b/src/Monolog/JsonSerializableDateTimeImmutable.php
@@ -1,0 +1,48 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Monolog;
+
+use DateTimeZone;
+
+/**
+ * Overrides default json encoding of date time objects
+ *
+ * @author Menno Holtkamp
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ */
+class JsonSerializableDateTimeImmutable extends \DateTimeImmutable implements \JsonSerializable
+{
+    private bool $useMicroseconds;
+
+    public function __construct(bool $useMicroseconds, ?DateTimeZone $timezone = null)
+    {
+        $this->useMicroseconds = $useMicroseconds;
+
+        // if you like to use a custom time to pass to Logger::addRecord directly,
+        // call modify() or setTimestamp() on this instance to change the date after creating it
+        parent::__construct('now', $timezone);
+    }
+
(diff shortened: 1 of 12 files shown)


Seldaek:
Thanks

InvisibleSmiley:
FYI, citing https://semver.org/
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backward compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated.

Seldaek:
Not sure what you're trying to say?

InvisibleSmiley:
New version should have been 3.9.0, not 3.8.1

Seldaek:
Ah I see. Yeah that was a mistake rushing this release while sitting at SymfonyCon.

src/Monolog/JsonSerializableDateTimeImmutable.php
```diff
@@ -0,0 +1,48 @@
...
+class JsonSerializableDateTimeImmutable extends \DateTimeImmutable implements \JsonSerializable
+{
+    private bool $useMicroseconds;
+
+    public function __construct(bool $useMicroseconds, ?DateTimeZone $timezone = null)
+    {
+        $this->useMicroseconds = $useMicroseconds;
+
+        // if you like to use a custom time to pass to Logger::addRecord directly,
+        // call modify() or setTimestamp() on this instance to change the date after creating it
```
glensc, line 31: Doesn't immutable mean that you can't modify the instance?

I tested locally, and set

glensc: > This class behaves the same as [DateTime](https://www.php.net/manual/en/class.datetime.php) except new objects are returned when modification methods such as [DateTime::modify()](https://www.php.net/manual/en/datetime.modify.php) are called.

- https://www.php.net/manual/en/class.datetimeimmutable.php

Seldaek: You need to use the instance returned by setTimestamp and pass that to Logger::addRecord as custom datetime instance.

PR #1933, opened 2025-01-03
Fix ErrorLogHandler type hint: $messageType accepts only SAPI and OS

glensc:
Introduced via f43e3d563776da6e696648025437ea0576dd1e9a


Seldaek:
Thanks

Issue #1912, opened 2024-09-17
Overwriting a log file?


phrfpeixoto:
Hello there!

I'm using monolog 2.9 and I'm looking for an easy way to overwrite a log file.
My use case is that I have processes that run in frequent intervals, and I want to be able to capture both the full log for all operations, as well as the "last log" for the latest iteration.

I fussed into the StreamHandler, and it it always `fopen`s the file in append more.
I tried creating a custom handler, but it seems that visibility for some required methods is restricted to `private`.

PR #1875, opened 2024-03-31
StreamHandler::customErrorHandler as protected

corentin-cres:
Fixes https://github.com/Seldaek/monolog/issues/1866

Issue #1866, opened 2024-01-05
cannot access private method


buexplain:
Monolog version 3

```php
PHP Warning:  Invalid callback Monolog\Handler\RotatingFileHandler::customErrorHandler, cannot access private method Monolog\Handler\RotatingFileHandler::customErrorHandler()
```

lcharette:
I was able to replicate this UserFrosting by removing write permission to the log file, which resulted in a similar error message : 

```
Fatal error: Uncaught Error: Invalid callback Monolog\Handler\StreamHandler::customErrorHandler, cannot access private method Monolog\Handler\StreamHandler::customErrorHandler()
```

Using the example above (https://3v4l.org/ZHA87) on the same environment and on the same file (without write access) returned the proper output : 

```
string(40) "Failed to open stream: Permission denied" string(9) "Test test" 
```

The difference is our app use a custom StreamWrapper. So the core PHP `fopen` isn't actually called by Monolog's StreamHandler, but our own. And **both** will invoke `customErrorHandler`, which won't work on our side since the method is private. Therefore, `customErrorHandler` must be **public**.

Note the actual process when `customErrorHandler` is public is similar to this : 
1. Monolog StreamHandler call `fopen`
2. Our own StreamHandler calls `fopen`
3. `fopen` trows an exception
4. Our StreamHandler invoke `customErrorHandler` (even if we use `@fopen`), then do the rest of it's job and return `false`.
5. Monolog invoke `customErrorHandler` **again**

lcharette:
Here's a replication : https://3v4l.org/CjZbo

Funny enough, looks like it only apply to PHP 8.2 and above

Seldaek:
Great, thanks for the investigative work there! I guess maybe using a closure instead would be better, otherwise public method with an internal phpdoc.. Anyway I'll look at this next week

Seldaek:
@lcharette can you please check using dev-main if this problem is fixed for you?

lcharette:
I confirm problem is fixed with `dev-main`. Thanks !

Seldaek:
Great

interactiveRob:
just in case anyone else from Userfrosting winds up here and is wondering which log file @lcharette is talking about above. --

That would be the `/logs/userfrosting.log` file. In my case, the app did not have write permission on the log which triggered the same private method issue discussed here.

PR #1934, opened 2025-01-13
Change visibility of handler classes to protected

PillFall:
Some properties and methods of the handler classes are set to private and some others to protected.

I change the visibility of all of them to protected. I believe this is a better practice, as it allows us to extend the handler.

If there are any security concerns of why this is the way it is, please do let me know.

### Example

The `MongoDBHandler` stores the configuration private as shown:

https://github.com/Seldaek/monolog/blob/d97d5e9468aa7d10f95b2e0468f39afdbd08a4d1/src/Monolog/Handler/MongoDBHandler.php#L37-L41

But the `DynamoDBHandler` stores the configuration protected as shown:

https://github.com/Seldaek/monolog/blob/d97d5e9468aa7d10f95b2e0468f39afdbd08a4d1/src/Monolog/Handler/DynamoDbHandler.php#L32-L36

With how is the code right now, I can extend the `DynamoDbHandler` to add querying functionality, all without having to create a new configuration object (aka `Aws\DynamoDb\DynamoDbClient`).

But with `MongoDBHandler`, I have to redo the argument parsing in its construct function to include this functionality, creating also two copies of the same exact object in memory (aka `MongoDB\Driver\Manager` or `\MongoDB\Collection`).

### DIFF
--- a/src/Monolog/Handler/AmqpHandler.php
+++ b/src/Monolog/Handler/AmqpHandler.php
@@ -25,7 +25,7 @@ class AmqpHandler extends AbstractProcessingHandler
     protected AMQPExchange|AMQPChannel $exchange;
 
     /** @var array<string, mixed> */
-    private array $extraAttributes = [];
+    protected array $extraAttributes = [];
 
     protected string $exchangeName;
 
@@ -147,7 +147,7 @@ class AmqpHandler extends AbstractProcessingHandler
         return strtolower($routingKey);
     }
 
-    private function createAmqpMessage(string $data): AMQPMessage
+    protected function createAmqpMessage(string $data): AMQPMessage
     {
         $attributes = [
             'delivery_mode' => 2,
--- a/src/Monolog/Handler/BrowserConsoleHandler.php
+++ b/src/Monolog/Handler/BrowserConsoleHandler.php
@@ -165,7 +165,7 @@ class BrowserConsoleHandler extends AbstractProcessingHandler
         return self::FORMAT_UNKNOWN;
     }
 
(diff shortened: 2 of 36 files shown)


Seldaek:
Thanks for the PR, however I do not think this is a good change as it makes refactoring much harder as the API surface becomes much larger.

We do accept PRs to open up properties/methods to be overridden/extended whenever someone has a valid use case though.

PillFall:
> I do not think this is a good change as it makes refactoring much harder as the API surface becomes much larger.

@Seldaek can you explain this for me?

I don't get why this change will make refactoring harder, this PR will only change a private w/o extension to a private w/ extension.

The same properties and methods will be available in a similar fashion, but allowing users to further extend the API, there are zero noticeable changes for the user.

Also, this seems like a wanted feature and a main concern for users as this is a blocker for many projects.

See for example #1866, #1875, #1912, and so many others.

Seldaek:
Right now we get PRs to mark stuff protected when there's a use case for it. So we can discuss use cases, prevent some people from making mistakes, learn how people use it, etc.

If we mark it all protected, no method or property can be changed/refactored anymore without the chance of BC break. In terms of public API surface everything protected is part of the public API, and potentially used by someone somewhere.

I hope this helps clarify my view

stof:
@PillFall protected properties are covered by the backward compatibility requirements of semver (and properties are harder to handle than methods regarding compatibility layers when property hooks are not an option). Protected properties are closer to public ones than to private ones in term of impact on maintenance.

Issue #1902, opened 2024-07-31
Allow to specify time for log records


Chi-teck:
That would be nice to allow users mock time for log records.

`\Monolog\Logger::addRecord` has an optional parameter for log timestamp. However,  methods that implement `Psr\Log\LoggerInterface`  do not have such a parameter.

I think the best way to fix this is passing optional `$clock` parameter to the logger constructor that implements PSR 20 interface.
Then `addRecord` could fetch timestamp from the clock service

Seldaek:
Sounds reasonable to me if you want to send a PR :)

PR #1938, opened 2025-02-03
Feature/clock interface issue 1902

omerimzali:
This PR introduces a new feature to allow users to specify a custom time for log records in Monolog. The enhancement involves integrating a Clock service that implements the PSR-20 ClockInterface, enabling more flexible and testable timestamp management.

Clock Integration: Added a Clock class that provides timestamps using JsonSerializableDateTimeImmutable. This allows for consistent and customizable time handling.

Logger Modification: Updated the Logger class to accept an optional ClockInterface parameter. The addRecord method now uses this clock to fetch timestamps when specified.

@Seldaek https://github.com/Seldaek/monolog/issues/1902 I saw this issue and I'm sending this PR

### DIFF
--- a/src/Monolog/Clock.php
+++ b/src/Monolog/Clock.php
@@ -1,0 +1,39 @@
+<?php declare(strict_types=1);
+
+namespace Monolog;
+
+use Psr\Clock\ClockInterface;
+use Monolog\JsonSerializableDateTimeImmutable;
+
+class Clock implements ClockInterface
+{
+    private bool $useMicroseconds;
+    private ?\DateTimeZone $timezone;
+    private ?JsonSerializableDateTimeImmutable $fixedTime;
+
+    public function __construct(bool $useMicroseconds = true, ?\DateTimeZone $timezone = null)
+    {
+        $this->useMicroseconds = $useMicroseconds;
+        $this->timezone = $timezone;
+    }
+
+    public function now(): JsonSerializableDateTimeImmutable
+    {
+        return $this->fixedTime ?? new JsonSerializableDateTimeImmutable($this->useMicroseconds, $this->timezone);
+    }
+
+    public function setUseMicroseconds(bool $useMicroseconds): void
+    {
+        $this->useMicroseconds = $useMicroseconds;
+    }
+
+    public function setTimezone(?\DateTimeZone $timezone): void
+    {
+        $this->timezone = $timezone;
+    }
+
+    public function setFixedTime(JsonSerializableDateTimeImmutable $fixedTime): void
+    {
+        $this->fixedTime = $fixedTime;
+    }
+}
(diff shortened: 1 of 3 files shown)


src/Monolog/Logger.php
```diff
@@ -164,21 +166,28 @@ class Logger implements LoggerInterface, ResettableInterface
      */
     private bool $detectCycles = true;
 
+    /**
+     * @var ClockInterface|null
```
stof, line 170: this `@var` brings no value as it only repeat the native property type.

src/Monolog/Logger.php
```diff
@@ -19,10 +19,12 @@
 use Psr\Log\LoggerInterface;
 use Psr\Log\InvalidArgumentException;
 use Psr\Log\LogLevel;
+use Psr\Clock\ClockInterface;
 use Throwable;
 use Stringable;
 use WeakMap;
 
+
```
stof, line 27: should be removed. We don't need 2 empty lines.

src/Monolog/Clock.php
```diff
@@ -0,0 +1,39 @@
+<?php declare(strict_types=1);
+
+namespace Monolog;
+
+use Psr\Clock\ClockInterface;
+use Monolog\JsonSerializableDateTimeImmutable;
+
+class Clock implements ClockInterface
```
stof, line 8: I suggest using a less specific name than `Clock` to avoid having it appearing high in IDE completion in projects using Monolog (while Monolog is not meant to become a library providing a clock for generic purposes that would involve making it more customizable)

src/Monolog/Logger.php
```diff
@@ -329,7 +338,7 @@ public function useLoggingLoopDetection(bool $detectCycles): self
      *
      * @phpstan-param value-of<Level::VALUES>|Level $level
      */
-    public function addRecord(int|Level $level, string $message, array $context = [], JsonSerializableDateTimeImmutable|null $datetime = null): bool
+    public function addRecord(int|Level $level, string $message, array $context = [], bool $useClock = false): bool
```
stof, line 341: This signature change is a BC break, which is a no-go in a minor version (and a bad idea in a major version if we don't provide a migration path through deprecations first)

src/Monolog/Logger.php
```diff
@@ -356,8 +365,9 @@ public function addRecord(int|Level $level, string $message, array $context = []
         try {
             $recordInitialized = \count($this->processors) === 0;
 
+            $datetime = $useClock ? $this->clock->now() : new JsonSerializableDateTimeImmutable($this->microsecondTimestamps, $this->timezone);
```
stof, line 368: phpstan is right about the issue reported here: if the caller passes `true` for `$useClock`, this will break when the clock is `null`.

And this is also quite useless when using the PSR-3 interface as those would never pass `true` there.

omerimzali:
I'm closing the PR I'll create another one according to your review thank you @stof

PR #1941, opened 2025-02-05
Add `strtolower` for `fromName` match statement

Jonezzyboy:
Changes the `fromName` method to be case-insensitive for more flexible use.


Seldaek:
Thanks

PR #1942, opened 2025-02-06
Change datetime format from ISO 8601 to ATOM in ElasticsearchFormatter

rajmundtoth0:
ISO8601 is deprecated.


Seldaek:
Thanks

staabm:
@rajmundtoth0 is this change backwards compatible? at least it changed the underlying date format, see https://github.com/Seldaek/monolog/pull/1954

Seldaek:
I'm fairly sure elastic search can handle both formats so it seems fine to me.

Issue #1943, opened 2025-02-06
BufferHandler to allow change of wrapped handler in runtime


A-Shevchenko:
Hello,

it would be really good to allow change of wrapped handler by BufferHandler after it's instantiation.
Our use case - we want to buffer logs until execution end, however in some cases we want to change underlying handler based on some conditions in application logic and do not lose what was logged so far.

Seldaek:
Fixed by #1946

PR #1946, opened 2025-02-07
added possibility to set child handler after instantiation

A-Shevchenko:
Hello,

this is to address #1943 - pretty straightforward.
Also, can be easily backported to 2.x - please let me know if I should open another PR for 2.x branch


Seldaek:
Makes sense, thanks

Seldaek:
I'd rather keep these to 3.x+ though unless you personally need it in 2.x, as it's quite a special use case

PR #1948, opened 2025-02-21
Add a mail() hook to NativeMailHandler.

jdwx:
The existing call to the global mail() function in NativeMailHandler::send() is replaced with a call to a protected method NativeMailHandler::mail() that takes the same parameters. The provided implementation of NativeMailHandler::mail() passes through to the global mail() function, leaving existing behavior completely unchanged.

This allows subclassing NativeMailHandler so that the call to mail() can be replaced in cases where additional processing needs to happen.

Tests are mostly updated to use this method to test sent messages, but one test is added to make sure that everything gets passed through intact in the default case.

I have run into a number of cases specific to FreeBSD 14+ where its new native mailer (dma) often rejects messages generated by Monolog. The added functionality makes it much easier to shim in the needed adjustments without having to maintain an entire local copy of NativeMailHandler to alter one line.

This approach would also allow for other more general post-processing of outbound messages if needed.

Thanks!

### DIFF
--- a/src/Monolog/Handler/NativeMailerHandler.php
+++ b/src/Monolog/Handler/NativeMailerHandler.php
@@ -128,7 +128,7 @@ class NativeMailerHandler extends MailHandler
 
         $parameters = implode(' ', $this->parameters);
         foreach ($this->to as $to) {
-            mail($to, $subject, $content, $headers, $parameters);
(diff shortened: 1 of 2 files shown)


Seldaek:
Makes sense, thanks

Issue #1935, opened 2025-01-28
SendGrid v3 Support [help wanted!]


Seldaek:
Anyone relying on SendGridHandler, this is currently using their v2 API, which will be deprecated on Feb 10th 2025.

If someone can send a PR to update the code to use the v3 API that would be great, I think this should probably be fixed for Monolog 2.x and 3.x.

PR #1952, opened 2025-03-10
SendGrid V3 API

rajmundtoth0:
#1935

### DIFF
--- a/src/Monolog/Formatter/ElasticaFormatter.php
+++ b/src/Monolog/Formatter/ElasticaFormatter.php
@@ -77,9 +77,6 @@ class ElasticaFormatter extends NormalizerFormatter
     {
         $document = new Document();
         $document->setData($record);
-        if (method_exists($document, 'setType')) {
-            $document->setType($this->type);
-        }
         $document->setIndex($this->index);
 
         return $document;
--- a/src/Monolog/Formatter/ElasticaFormatter.php
+++ b/src/Monolog/Formatter/ElasticaFormatter.php
@@ -77,6 +77,9 @@ class ElasticaFormatter extends NormalizerFormatter
     {
         $document = new Document();
         $document->setData($record);
+        if (method_exists($document, 'setType')) {
+            $document->setType($this->type);
+        }
         $document->setIndex($this->index);
 
         return $document;
--- a/src/Monolog/Formatter/ElasticaFormatter.php
+++ b/src/Monolog/Formatter/ElasticaFormatter.php
@@ -77,6 +77,7 @@ class ElasticaFormatter extends NormalizerFormatter
     {
         $document = new Document();
         $document->setData($record);
+        /** @phpstan-ignore function.impossibleType */
         if (method_exists($document, 'setType')) {
             $document->setType($this->type);
         }
--- a/src/Monolog/Handler/SendGridHandler.php
+++ b/src/Monolog/Handler/SendGridHandler.php
@@ -15,7 +15,7 @@ use Monolog\Level;
 use Monolog\Utils;
 
 /**
- * SendGridrHandler uses the SendGrid API v2 function to send Log emails, more information in https://sendgrid.com/docs/API_Reference/Web_API/mail.html
+ * SendGridHandler uses the SendGrid API v3 function to send Log emails, more information in https://www.twilio.com/docs/sendgrid/for-developers/sending-email/api-getting-started
  *
  * @author Ricardo Fontanelli <ricardo.fontanelli@hotmail.com>
  */
@@ -25,6 +25,7 @@ class SendGridHandler extends MailHandler
 
     /**
      * The SendGrid API User
+     * @deprecated this is not used anymore as of SendGrid API v3
      */
     protected string $apiUser;
     /**
@@ -34,11 +35,12 @@ class SendGridHandler extends MailHandler
     protected array $to;
 
     /**
+     * @param string|null $apiUser Unused user as of SendGrid API v3, you can pass null or any string
      * @param list<string>|string $to
      * @throws MissingExtensionException If the curl extension is missing
      */
     public function __construct(
-        string $apiUser,
+        string|null $apiUser,
         protected readonly string $apiKey,
         protected string $from,
         array|string $to,
@@ -47,14 +49,13 @@ class SendGridHandler extends MailHandler
         bool $bubble = true,
         /** @var non-empty-string */
         private readonly string $sendGridApiUrl = 'https://api.sendgrid.com/v3/mail/send',
-        )
-    {
+    ) {
         if (!\extension_loaded('curl')) {
             throw new MissingExtensionException('The curl extension is needed to use the SendGridHandler');
         }
 
         $this->to = (array) $to;
-        $this->apiUser = $apiUser;
+        $this->apiUser = $apiUser ?? '';
         parent::__construct($level, $bubble);
     }
 
--- a/src/Monolog/Handler/SendGridHandler.php
+++ b/src/Monolog/Handler/SendGridHandler.php
@@ -20,7 +20,7 @@ use Monolog\Level;
  */
 class SendGridHandler extends MailHandler
 {
-    private const string CONTENT_TYPE = 'Content-Type: application/json';
+    private const CONTENT_TYPE = 'Content-Type: application/json';
     /**
      * @throws MissingExtensionException If the curl extension is missing
      */
--- a/src/Monolog/Handler/SendGridHandler.php
+++ b/src/Monolog/Handler/SendGridHandler.php
@@ -27,10 +27,12 @@ class SendGridHandler extends MailHandler
     public function __construct(
         private readonly string $apiKey,
         private readonly string $from,
+        /** @var list<string> */
         private readonly array $to,
         private readonly string $subject,
         int|string|Level $level = Level::Error,
         protected bool $bubble = true,
+        /** @var non-empty-string */
         private readonly string $sendGridApiUrl = 'https://api.sendgrid.com/v3/mail/send',
         )
     {
@@ -68,7 +70,8 @@ class SendGridHandler extends MailHandler
         curl_setopt($ch, CURLOPT_URL, $this->sendGridApiUrl);
         curl_setopt($ch, CURLOPT_POST, true);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
-        curl_setopt($ch, CURLOPT_POSTFIELDS, json_encode($body));
+        curl_setopt($ch, CURLOPT_POSTFIELDS, json_encode($body, JSON_THROW_ON_ERROR));
+
         Curl\Util::execute($ch, 2);
     }
 }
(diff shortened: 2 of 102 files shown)


src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +20,58 @@
...
-     * @param string          $apiKey  The SendGrid API Key
-     * @param string          $from    The sender of the email
-     * @param string|string[] $to      The recipients of the email
-     * @param string          $subject The subject of the mail
-     *
      * @throws MissingExtensionException If the curl extension is missing
      */
-    public function __construct(string $apiUser, string $apiKey, string $from, string|array $to, string $subject, int|string|Level $level = Level::Error, bool $bubble = true)
+    public function __construct(
+        private readonly string $apiKey,
```
stof, line 28: Turning a protected property into a private one is a BC break.

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +20,58 @@
...
-    /**
-     * @param string          $apiUser The SendGrid API User
-     * @param string          $apiKey  The SendGrid API Key
-     * @param string          $from    The sender of the email
-     * @param string|string[] $to      The recipients of the email
-     * @param string          $subject The subject of the mail
-     *
      * @throws MissingExtensionException If the curl extension is missing
      */
-    public function __construct(string $apiUser, string $apiKey, string $from, string|array $to, string $subject, int|string|Level $level = Level::Error, bool $bubble = true)
```
stof, line 58: Removing the first constructor argument is a BC break which cannot be accepted in a minor version of the library.

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +20,58 @@
...
+    public function __construct(
+        private readonly string $apiKey,
+        private readonly string $from,
+        /** @var list<string> */
+        private readonly array $to,
+        private readonly string $subject,
+        int|string|Level $level = Level::Error,
+        protected bool $bubble = true,
```
stof, line 34: Redefining a `$bubble` property here does not make sense, as we want to use the property of the parent class. This parameter should not be a promoted property.

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +20,58 @@
...
-     * @param string          $subject The subject of the mail
-     *
      * @throws MissingExtensionException If the curl extension is missing
      */
-    public function __construct(string $apiUser, string $apiKey, string $from, string|array $to, string $subject, int|string|Level $level = Level::Error, bool $bubble = true)
+    public function __construct(
+        private readonly string $apiKey,
+        private readonly string $from,
+        /** @var list<string> */
+        private readonly array $to,
```
stof, line 31: Changing the parameter type from `string|array` to `array` is a BC break, that is not worth it IMO

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +20,58 @@
...
+        private readonly string $subject,
+        int|string|Level $level = Level::Error,
+        protected bool $bubble = true,
+        /** @var non-empty-string */
+        private readonly string $sendGridApiUrl = 'https://api.sendgrid.com/v3/mail/send',
```
stof, line 36: what is the reason to make this configurable ?

rajmundtoth0: There are two urls, this, and one dedicated for the EU.

src/Monolog/Handler/SendGridHandler.php
Seldaek: To me the question here is: It seems like the v2 api had a user+key for auth, and v3 only has a key.. Does this mean v2 credentials will not work for v3? If so, and the v2 API is shut down now.. then this old code is anyway broken and the BC does not need to be respected IMO.

If there is a way to make an existing configuration remain compatible (incl key and the username would just be ignored) tho, then it would of course be preferable.

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +20,58 @@
...
-
+        $authorization = "Authorization: Bearer {$this->apiKey}";
         $ch = curl_init();
-        curl_setopt($ch, CURLOPT_URL, 'https://api.sendgrid.com/api/mail.send.json');
+        curl_setopt($ch, CURLOPT_HTTPHEADER, [self::CONTENT_TYPE , $authorization]);
+        curl_setopt($ch, CURLOPT_URL, $this->sendGridApiUrl);
         curl_setopt($ch, CURLOPT_POST, true);
         curl_setopt($ch, CURLOPT_RETURNTRANSFER, true);
-        curl_setopt($ch, CURLOPT_POSTFIELDS, http_build_query($message));
+        curl_setopt($ch, CURLOPT_POSTFIELDS, json_encode($body, JSON_THROW_ON_ERROR));
```
Seldaek, line 73: Should use `Utils::jsonEncode` instead of json_encode

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0: V3 uses only api key. I changed the code to make it BC compatible, although removing it is cleaner IMO.

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0: Fixed.

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0: Right, I changed it. Maybe we can get rid of that in a bigger refactor later on.

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0: Fixed.

src/Monolog/Formatter/ElasticaFormatter.php
```diff
@@ -77,10 +77,6 @@ protected function getDocument(array $record): Document
     {
         $document = new Document();
         $document->setData($record);
```
rajmundtoth0, line 79: This was removed from Elastica v7. This package relies on v 7 or v8.

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0: Right, thanks for the comment. Should be better now :)

src/Monolog/Handler/SendGridHandler.php
stof: I would say passing a `$region` parameter is probably more usable as API than configuring directly the URL of the endpoint.

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +21,72 @@
...
-     * @param string|string[] $to      The recipients of the email
-     * @param string          $subject The subject of the mail
-     *
+     * @param list<string>|string $to
      * @throws MissingExtensionException If the curl extension is missing
      */
-    public function __construct(string $apiUser, string $apiKey, string $from, string|array $to, string $subject, int|string|Level $level = Level::Error, bool $bubble = true)
+    public function __construct(
+        string $apiUser,
+        protected readonly string $apiKey,
```
stof, line 42: turning a protected property into a protected readonly one is also a BC break. You need to keep it non-readonly.

stof: same for the subject

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0: Like
`if ($this->region === 'eu') {
    $this->sendGridApiUrl = 'https://api.eu.sendgrid.com/v3/mail/send';
else {
    $this->sendGridApiUrl = 'https://api.sendgrid.com/v3/mail/send';
}
`
?

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +21,72 @@
...
+        private readonly string $from,
+        array|string $to,
+        private readonly string $subject,
```
rajmundtoth0, line 45:
```suggestion
        protected readonly string $subject,
```

src/Monolog/Handler/SendGridHandler.php
stof: yes.

But you don't need to store both the region and the api url as property. One of them can be a local variable (which one is local depends if you do the selection of the API endpoint in the constructor or in the `send` method)

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +21,72 @@
...
-     * @param string          $subject The subject of the mail
-     *
+     * @param list<string>|string $to
      * @throws MissingExtensionException If the curl extension is missing
      */
-    public function __construct(string $apiUser, string $apiKey, string $from, string|array $to, string $subject, int|string|Level $level = Level::Error, bool $bubble = true)
+    public function __construct(
+        string $apiUser,
+        protected readonly string $apiKey,
+        private readonly string $from,
```
stof, line 43: this should also stay protected and non-readonly to avoid a BC break.

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0: Right. IMO I prefer it to be configurable, maybe there will be more URLS in the future. If you want it to be dependent on the region, I can do that too. Maybe we should ask @Seldaek ?

src/Monolog/Handler/SendGridHandler.php
rajmundtoth0, line 43:
```suggestion
        protected string $from,
```

src/Monolog/Formatter/ElasticaFormatter.php
Seldaek: Please refrain from modifying unrelated files in this PR tho. This derails the review and conversation. Open a new PR if needed then we can review and merge things independently

rajmundtoth0: No problem, I just wanted to make PHPStan happy. But I open a PR for this then

src/Monolog/Handler/SendGridHandler.php
Seldaek: Sorry just to be 100% sure v2 api keys still work with v3 or do you need a new key?

src/Monolog/Handler/SendGridHandler.php
```diff
@@ -20,81 +21,72 @@
  */
 class SendGridHandler extends MailHandler
 {
+    private const CONTENT_TYPE = 'Content-Type: application/json';
+
     /**
      * The SendGrid API User
```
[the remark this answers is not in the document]
rajmundtoth0: https://www.twilio.com/docs/sendgrid/for-developers/sending-email/api-getting-started
image

rajmundtoth0:
I'm unsure if v2 keys work with v3, but I assume no.

Seldaek:
Thanks

Issue #1951, opened 2025-03-06
Rename TestCase class to prevent import conflicts with PHPUnit's TestCase


staabm:
Monolog version 3

Write your question here.

today I was stumbling over a problem of wrongly used `TestCase` class because autocompletion liked monologs class more then the phpunit one.

this remindet me of a similar issue we had in the past with PHPStan: https://github.com/phpstan/phpstan-src/pull/634

do you think monolog could make a similar move and deprecate+rename the monologs' TestCase class so autocompletion in popular IDEs won't accidentally add monologs TestCase class instead of e.g. PHPunits class?

I think most arguments why this is a good idea and actually the root cause is a phpstorm bug was already discussed in the linked PHPStan issue. I think a lot of people would have less problems when this class would be renamed, even though its totally fine because its in a dedicated namespace.

staabm:
just found a duplicate which already got closed: https://github.com/Seldaek/monolog/issues/1677

Seldaek:
Is the `@internal` workaround not working anymore?

Seldaek:
I mean renaming to MonologTestCase or something would be fine I think, with the deprecated/alias one extending from it and having the `@internal` tag. Do you want to PR this?

staabm:
Sure

PR #1953, opened 2025-03-16
Rename TestCase to MonologTestCase

staabm:
closes https://github.com/Seldaek/monolog/issues/1951

### DIFF
--- a/src/Monolog/Test/MonologTestCase.php
+++ b/src/Monolog/Test/MonologTestCase.php
@@ -22,8 +22,6 @@ use Psr\Log\LogLevel;
  * Lets you easily generate log records and a dummy formatter for testing purposes
  *
  * @author Jordi Boggiano <j.boggiano@seld.be>
- *
- * @internal feel free to reuse this to test your own handlers, this is marked internal to avoid issues with PHPStorm https://github.com/Seldaek/monolog/issues/1677
  */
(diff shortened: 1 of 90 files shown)


staabm:
instead of a class_alias I went with a regular class and marked it as `@deprecated`.
that way consumers of this lib - or other projects which mistakenly used the old base-class - will see phpstorm/phpstan reports about the wrongly used class-name (when using class-alias the old class-name cannot be marked as deprecated)

src/Monolog/Test/MonologTestCase.php
```diff
@@ -0,0 +1,73 @@
...
+use Monolog\JsonSerializableDateTimeImmutable;
+use Monolog\Formatter\FormatterInterface;
+use Psr\Log\LogLevel;
+
+/**
+ * Lets you easily generate log records and a dummy formatter for testing purposes
+ *
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * @internal feel free to reuse this to test your own handlers, this is marked internal to avoid issues with PHPStorm https://github.com/Seldaek/monolog/issues/1677
```
Seldaek, line 26: Not needed anymore

```suggestion
```

Seldaek:
Thanks

PR #1954, opened 2025-03-16
Fix build

staabm:
~~not sure why the expectation no longer matches/changed~~

update: I think this changed with https://github.com/Seldaek/monolog/pull/1942

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -43,12 +43,6 @@ parameters:
 			path: src/Monolog/Formatter/WildfireFormatter.php
 
 		-
-			message: '#^Method Monolog\\Handler\\BufferHandler\:\:setHandler\(\) has no return type specified\.$#'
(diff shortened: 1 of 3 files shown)


phpstan-baseline.neon
```diff
@@ -42,6 +42,12 @@ parameters:
 			count: 1
 			path: src/Monolog/Formatter/WildfireFormatter.php
 
+		-
+			message: '#^Method Monolog\\Handler\\BufferHandler\:\:setHandler\(\) has no return type specified\.$#'
```
staabm, line 46: baselined this error, as adding a new return type is likely a BC break

staabm: this method [was recently added](https://github.com/Seldaek/monolog/pull/1946) and not yet part of a release -> I added the return type instead

Seldaek:
Thanks

PR #1956, opened 2025-03-18
Updating while loops to have boolean conditions.

rajmundtoth0:


src/Monolog/Handler/Curl/Util.php
```diff
@@ -37,9 +37,10 @@ final class Util
      * @param  CurlHandle  $ch curl handler
      * @return bool|string @see curl_exec
      */
-    public static function execute(CurlHandle $ch, int $retries = 5, bool $closeAfterDone = true)
+    public static function execute(CurlHandle $ch, int $retries = 5, bool $closeAfterDone = true): bool|string
     {
-        while ($retries--) {
+        while (true) {
+            $retries--;
```
stof, line 43: This is broken. It does not stop the loop anymore when `$retries` reaches 0

src/Monolog/Formatter/LineFormatter.php
```diff
@@ -204,16 +204,19 @@ protected function normalizeException(\Throwable $e, int $depth = 0): string
...
-                if ($depth > $this->maxNormalizeDepth) {
-                    $str .= "\n[previous exception] Over " . $this->maxNormalizeDepth . ' levels deep, aborting normalization';
-                    break;
-                }
+            while (true) {
+            $previous = $e->getPrevious();
+            if (!($previous instanceof \Throwable)) {
+                break;
+            }
+            if ($depth > $this->maxNormalizeDepth) {
```
stof, line 212: this changes the logic because you are no longer incrementing the depth before that condition, while it was the case before.

src/Monolog/Formatter/LineFormatter.php
```diff
@@ -204,16 +204,19 @@ protected function normalizeException(\Throwable $e, int $depth = 0): string
...
         $str = $this->formatException($e);
 
-        if (($previous = $e->getPrevious()) instanceof \Throwable) {
-            do {
-                $depth++;
-                if ($depth > $this->maxNormalizeDepth) {
-                    $str .= "\n[previous exception] Over " . $this->maxNormalizeDepth . ' levels deep, aborting normalization';
-                    break;
-                }
+            while (true) {
```
stof, line 207: coding standards are off here

src/Monolog/Formatter/LineFormatter.php
```diff
@@ -204,16 +204,19 @@ protected function normalizeException(\Throwable $e, int $depth = 0): string
...
+            $previous = $e->getPrevious();
```
stof, line 208: this is broken, because you restart with `$previous = $e->getPrevious();` at each iteration of the loop instead of using the modified `$previous`

PR #1958, opened 2025-03-20
Fix PHPStan lvl9 issues on LineFormatter

rajmundtoth0:
This PR resolves 22 / 115 PHPStan lvl 9 issues.


Seldaek:
Not sure what the point is as we do not run phpstan with level 9..

rajmundtoth0:
Right, but this could be the first step to lvl 9

Seldaek:
Frankly I have no interest in level 9 support, mixed is a necessary evil sometimes, and Monolog normalized need this kind of fuzziness to be able to function, so I don't see the point of fighting against phpstan there as it will not provide any value. So thanks but I don't think it's worth it.

PR #1967, opened 2025-05-17
Fix race condition in StreamHandler directory creation with locking

TheJoeSchr:
This PR introduces a locking mechanism for directory creation in `StreamHandler` when `useLocking` is enabled. This prevents race conditions when multiple processes (e.g., Laravel queue workers using a 'single' log stack) try to create the
same log directory simultaneously. The handler now creates a lock file in the parent directory, acquires an exclusive lock, creates the target directory, and then releases the lock.

**Testing this fix:**

To manually test this behavior, especially in a concurrent environment like Laravel queues:
1.  Ensure `useLocking` is `true` for your `StreamHandler` configuration.
2.  Set up a scenario where multiple queue jobs are dispatched rapidly.
3.  Ensure these jobs log to the same file path, and the log directory does not exist before the jobs run.
4.  Observe that the directory is created successfully without errors from concurrent `mkdir` attempts, and all jobs log as expected.

New unit tests in `StreamHandlerTest.php` also cover the `createDir` method's locking behavior.

### DIFF
--- a/src/Monolog/Handler/StreamHandler.php
+++ b/src/Monolog/Handler/StreamHandler.php
@@ -231,14 +231,91 @@ class StreamHandler extends AbstractProcessingHandler
 
         $dir = $this->getDirFromStream($url);
         if (null !== $dir && !is_dir($dir)) {
-            $this->errorMessage = null;
-            set_error_handler(function (...$args) {
-                return $this->customErrorHandler(...$args);
-            });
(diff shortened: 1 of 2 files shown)


TheJoeSchr:
If the lock can't be acquired, an exponential and randomized step of with retries is implemented. This now also gets used for acquiring locks in `write()`

PR #1973, opened 2025-07-01
[GelfFormatter] Convert bool to int

### LINKED ISSUE #1972: [GelfFormatter] Missing boolean values
Monolog version: 3

The GELF Payload Specification does not support `bool` values as a addtional field. See https://go2docs.graylog.org/current/getting_in_log_data/gelf.html#GELFPayloadSpecification  

If a `bool` value is provieded Graylog will simply drop it.

Given the following example code

```php

        $this->logger->info("debug logger", [
            'int' => 113,
            'float' => 23.23,
            'trueBool' => true,
            'falseBool' => false,
            'object' => (object) [],
            'array' => ['some' => 'array'],
        ]);
```
will result in this log in Graylog:
Image

Therefore bool values should be converted before adding them to the Gelf Message.

am4rth:
Convert all `bool` values in `extra` and `context` to `int` to prevent Graylog from loosing them.

Fixes #1972

### DIFF
--- a/src/Monolog/Formatter/GelfMessageFormatter.php
+++ b/src/Monolog/Formatter/GelfMessageFormatter.php
@@ -116,6 +116,7 @@ class GelfMessageFormatter extends NormalizerFormatter
 
         foreach ($extra as $key => $val) {
             $key = (string) preg_replace('#[^\w.-]#', '-', (string) $key);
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #1980, opened 2025-07-30
Normalize DIRECTORY_SEPARATOR on file traces(exceptions)

angeljqv:
When using `setBasePath($path)`, this PR makes Windows and Linux logs look the same.


Seldaek:
Thanks

PR #1983, opened 2025-08-22
Add first-party `RedactSensitiveProcessor` to mask secrets in logs (m…

rishi111-code:
…essage/context/extra)

Introduce a new RedactSensitiveProcessor for Monolog that automatically masks sensitive keys (like password, token, api_key) and user-defined patterns (e.g., bearer tokens, email addresses) in both log messages and record data (context and extra). Full unit tests and documentation included.

Motivation:
Logs frequently leak sensitive information (e.g., auth tokens or credentials). This processor provides a built-in, secure, zero-dependency way to redact common data, meeting a real community need and improving the developer experience ‌— no custom wiring needed. It complements existing processors like UidProcessor by addressing security concerns directly. Better Stack
getparthenon.com

Implementation Details:

New class RedactSensitiveProcessor (in src/Monolog/Processor) implements ProcessorInterface.

Configurable constructor accepts:

$sensitiveKeys (case-insensitive array of keys to mask)

$patterns (regex patterns for message redaction)

$mask (custom replacement string)

It safely ignores invalid regex patterns.

Applies redaction on both context, extra, and the message.

Fully covered by PHPUnit tests including edge cases (nested arrays, invalid regex) in tests/Monolog/Processor/RedactSensitiveProcessorTest.php. Benefits:

Addresses a frequent security requirement directly in the core library.

Safe default behavior; invalid patterns are ignored, preventing runtime errors.

Lightweight and zero-dependency, posing a minimal maintenance burden.

Fully documented and tested, ready for production use.

Backward Compatibility & Performance:
Non-breaking addition. High-performance — simple string and array operations. Regex evaluation is optional and safe. Also doesn’t alter existing logging behavior when not configured.

### DIFF
--- a/src/Monolog/Processor/RedactSensitiveProcessor.php
+++ b/src/Monolog/Processor/RedactSensitiveProcessor.php
@@ -1,0 +1,77 @@
+<?php
+declare(strict_types=1);
+
+namespace Monolog\Processor;
+
+use Monolog\LogRecord;
+
+final class RedactSensitiveProcessor implements ProcessorInterface
+{
+    /** @var string[] */
+    private array $sensitiveKeys;
+
+    /** @var string[] */
+    private array $patterns;
+
+    private string $mask;
+
+    /**
+     * @param string[] $sensitiveKeys exact keys to redact in context/extra (case-insensitive)
+     * @param string[] $patterns PCRE regex patterns to mask inside string values (e.g. '/Bearer\\s+[A-Za-z0-9\\._-]+/')
+     * @param string   $mask replacement token
+     */
+    public function __construct(array $sensitiveKeys = ['password','passwd','pwd','secret','token','api_key','apikey','authorization','auth','cookie'], array $patterns = [], string $mask = 'REDACTED')
+    {
+        $this->sensitiveKeys = array_map('strtolower', $sensitiveKeys);
+        $this->patterns = $patterns;
+        $this->mask = $mask;
+    }
+
+    public function __invoke(LogRecord $record): LogRecord
+    {
+        $context = $this->sanitize($record->context);
+        $extra   = $this->sanitize($record->extra);
+
+        $message = $record->message;
+        foreach ($this->patterns as $pattern) {
+            if (@preg_match($pattern, '') === false) {
+                continue; // ignore invalid pattern instead of throwing inside logging path
+            }
+            $message = preg_replace($pattern, $this->mask, $message) ?? $message;
+        }
+
+        return $record->with(
+            message: $message,
+            context: $context,
+            extra: $extra
+        );
+    }
+
+    /** @param mixed $value */
+    private function sanitize(mixed $value): mixed
+    {
+        if (is_array($value)) {
+            $sanitized = [];
+            foreach ($value as $k => $v) {
+                $key = is_string($k) ? strtolower($k) : $k;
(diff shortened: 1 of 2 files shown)


src/Monolog/Processor/RedactSensitiveProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function __invoke(LogRecord $record): LogRecord
+    {
+        $context = $this->sanitize($record->context);
+        $extra   = $this->sanitize($record->extra);
+
+        $message = $record->message;
+        foreach ($this->patterns as $pattern) {
+            if (@preg_match($pattern, '') === false) {
+                continue; // ignore invalid pattern instead of throwing inside logging path
```
stof, line 38: I'm not sure we want to silently ignore failures. If the pattern is invalid and ignored silently, nothing will tell the dev that his expected masking won't heppe.

src/Monolog/Processor/RedactSensitiveProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+        );
+    }
+
+    /** @param mixed $value */
+    private function sanitize(mixed $value): mixed
+    {
+        if (is_array($value)) {
+            $sanitized = [];
+            foreach ($value as $k => $v) {
+                $key = is_string($k) ? strtolower($k) : $k;
```
stof, line 56:
```suggestion
                $key = \is_string($k) ? strtolower($k) : $k;
```

`is_string` is a compiler-optimized function, so it should be used in a fully-qualified way to benefit from it.

src/Monolog/Processor/RedactSensitiveProcessor.php
```diff
@@ -0,0 +1,77 @@
...
+    private array $patterns;
+
+    private string $mask;
+
+    /**
+     * @param string[] $sensitiveKeys exact keys to redact in context/extra (case-insensitive)
+     * @param string[] $patterns PCRE regex patterns to mask inside string values (e.g. '/Bearer\\s+[A-Za-z0-9\\._-]+/')
+     * @param string   $mask replacement token
+     */
+    public function __construct(array $sensitiveKeys = ['password','passwd','pwd','secret','token','api_key','apikey','authorization','auth','cookie'], array $patterns = [], string $mask = 'REDACTED')
```
Seldaek: Just an idea.. not sure if good or bad, but this should catch any long word of 30+ [a-z0-9] characters (optionally prefixed by foo_bar_ or the like, like many API tokens are), which is very likely to be a secret/token.. But of course it risks some collateral damage.

```suggestion
    public function __construct(array $sensitiveKeys = ['password','passwd','pwd','secret','token','api_key','apikey','authorization','auth','cookie'], array $patterns = ['{\b(?:[a-z]+_)*[a-zA-Z0-9]{30,}\b}'], string $mask = 'REDACTED')
```

src/Monolog/Processor/RedactSensitiveProcessor.php
Seldaek: IMO the best would be to run this pattern check *once* in the constructor, and if anything returns false you simply throw an InvalidArgumentException. That is how we handle all other misconfigurations in Monolog.

Seldaek:
Alternative PR here as a formatter, happy to hear feedback from others https://github.com/Seldaek/monolog/pull/2041

PR #1997, opened 2025-10-01
Fix PHP 8.5 `__sleep` deprecation

### LINKED ISSUE #1995: PHP 8.5 deprecated warning
Monologversion :
"version": "3.8.1",

PHP 8.5 RC1

001+ Deprecated: The __sleep() serialization magic method has been deprecated. Implement __serialize() instead (or in addition, if support for old PHP versions is necessary) in vendor/monolog/monolog/src/Monolog/Handler/Handler.php on line 19

angeljqv:
This patch breaks any implementation of __sleep in child classes/user handlers. Should be warned in the release notes.

Closes #1995

### DIFF
--- a/src/Monolog/Handler/Handler.php
+++ b/src/Monolog/Handler/Handler.php
@@ -48,25 +48,6 @@ abstract class Handler implements HandlerInterface
     {
         $this->close();
 
-        $reflClass = new \ReflectionClass($this);
-
-        $data = [];
-        foreach ($reflClass->getProperties() as $reflProp) {
-            if (!$reflProp->isStatic()) {
-                $data[$reflProp->getName()] = $reflProp->getValue($this);
-            }
-        }
-
-        return $data;
-    }
-
-    /**
-     * @param array<mixed> $data
-     */
-    public function __unserialize(array $data): void
-    {
-        foreach ($data as $name => $value) {
-            (new \ReflectionProperty($this, $name))->setValue($this, $value);
-        }
+        return (array) $this;
     }
 }


src/Monolog/Handler/Handler.php
```diff
@@ -44,19 +44,27 @@ public function __destruct()
...
         $this->close();
 
         $reflClass = new \ReflectionClass($this);
 
-        $keys = [];
+        $data = [];
         foreach ($reflClass->getProperties() as $reflProp) {
             if (!$reflProp->isStatic()) {
-                $keys[] = $reflProp->getName();
+                $reflProp->setAccessible(true);
```
stof, line 56: this call is useless. Monolog requires PHP 8.1+, and this call is a no-op in PHP 8 (and btw, it is *also* deprecated in PHP 8.5)

src/Monolog/Handler/Handler.php
```diff
@@ -44,19 +44,29 @@ public function __destruct()
...
         }
 
-        return $keys;
+        return $data;
+    }
+
+    /**
+     * @param array<mixed> $data
+     */
+    public function __unserialize(array $data): void
```
nicolas-grekas: this method should be removed: the implementation is almost what PHP does natively (it does more, taking mangled names into account if any) - the implementation here actually prevents any implementation of `__wakeup` in a child class from working.

src/Monolog/Handler/Handler.php
```diff
@@ -44,19 +44,29 @@ public function __destruct()
         }
     }
 
-    public function __sleep()
+    public function __serialize(): array
     {
         $this->close();
 
         $reflClass = new \ReflectionClass($this);
```
nicolas-grekas: all the reflection stuff is needless:
```suggestion
        return (array) $data;
```

there *could* be some optimization by removing mangled prefixes, except if any deep private prop is to be serialized. But if the size of the serialized payload doesn't matter, the cast to array is all that's needed.

Note that if a child class implements `__sleep()`, this patch will break them. Accounting for that needs more involving code. Maybe not worth it.

Issue #1984, opened 2025-08-29
RotateFileHandler unlink error not suppressed


greggebura:
Version 3.9.0

While rotating old files there is a code that should not suppressed errors if more than one processes will reach this part of code

```
// suppress errors here as unlink() might fail if two processes
// are cleaning up/rotating at the same time
set_error_handler(function (int $errno, string $errstr, string $errfile, int $errline): bool {
    return false;
});
```
Error handler should return true if error is handled properly.

PR #1999, opened 2025-10-16
Fix issue https://github.com/Seldaek/monolog/issues/1984

greggebura:
Fix issue https://github.com/Seldaek/monolog/issues/1984


GromNaN:
You should include a non-regression test with your patch, in [RotatingFileHandlerTest](https://github.com/Seldaek/monolog/blob/main/tests/Monolog/Handler/RotatingFileHandlerTest.php).

greggebura:
> You should include a non-regression test with your patch, in [RotatingFileHandlerTest](https://github.com/Seldaek/monolog/blob/main/tests/Monolog/Handler/RotatingFileHandlerTest.php).

I won't prepare any tests because:
- error is obvious - error handler has to return true if error was handled properly
- there are many tests that cover file rotation
- it is very difficult to provide test that will cover this case, because this error will only occur if two processes will reach this part of code in the same time

We need this fix, because this error produce many warnings in our project.

PR #2001, opened 2025-10-26
Add file rotation by hour

baks-dev:
Hello,

I suggest rotating log files hourly. Large logs are often difficult to load into the IDE, view and analyze. At the user's discretion, it would be useful to separate by specifying a format like 'Y-m-d-H' in the settings. I will give an example of a php configuration:

```php
    $config
        ->handler('request')
        ->formatter('monolog.formatter.standard')
        ->type('rotating_file')
        ->maxFiles(168)
        ->dateFormat('Y-m-d-H')
        ->filenameFormat('{filename}-{date}')
        ->path('%kernel.logs_dir%/request-%kernel.environment%.log')
        ->level('debug')
        ->channels()
        ->elements(['request']);
```


Seldaek:
Fixed/replaced by #2040

PR #2002, opened 2025-11-20
Add MS Teams webhook handler

Seb33300:
We are migrating from Slack to Microsoft Teams and we would like to continue receiving log notifications there, as this feature has been extremely useful for our workflow.

I am also planning to submit a PR on Laravel to support this handler if this PR is approved (Laravel already supports the slack handler).

The implementation is inspired from the existing [`SlackWebhookHandler`](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SlackWebhookHandler.php).

This handler support an additional parameter `toggleFields` where we can specify the list of field we want to display using a toggle button (eg: useful for stacktrace)

Here is how it looks:

(Note the username and avatar has to be configured when creating the webhook URL)

image

When we click on the toggle button:

image

### DIFF
--- a/src/Monolog/Handler/TeamsWebhookHandler.php
+++ b/src/Monolog/Handler/TeamsWebhookHandler.php
@@ -11,10 +11,9 @@
 
 namespace Monolog\Handler;
 
-use Monolog\Formatter\FormatterInterface;
 use Monolog\Level;
 use Monolog\Utils;
-use Monolog\Handler\Teams\TeamsRecord;
+use Monolog\Handler\Teams\TeamsPayload;
 use Monolog\LogRecord;
 
 /**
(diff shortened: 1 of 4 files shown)


src/Monolog/Handler/Slack/SlackRecord.php
```diff
@@ -97,10 +97,6 @@ public function __construct(
             ->includeContextAndExtra($includeContextAndExtra)
             ->excludeFields($excludeFields)
             ->setFormatter($formatter);
-
-        if ($this->includeContextAndExtra) {
-            $this->normalizerFormatter = new NormalizerFormatter();
-        }
```
Seb33300: This is already done in `->includeContextAndExtra()`

src/Monolog/Handler/TeamsWebhookHandler.php
```diff
@@ -0,0 +1,119 @@
...
+            $toggleFields
+        );
+    }
+
+    public function getTeamsRecord(): TeamsRecord
+    {
+        return $this->teamsRecord;
+    }
+
+    public function getWebhookUrl(): string
```
stof: I don't think those getters should exist

Seldaek:
Please submit this as a third party package and add it there https://github.com/Seldaek/monolog/wiki/Third-Party-Packages - or use one of the two that already exists https://github.com/cmdisp/monolog-microsoft-teams & https://github.com/actived/microsoft-teams-notifier

PR #2009, opened 2025-12-02
Remove deprecated curl_close() usage for PHP 8.5 compatibility

lacatoire:
This PR removes calls to `curl_close()`, which is deprecated in PHP 8.5 and has no effect anymore on CurlHandle objects since PHP 8.0.
The behavior remains unchanged while preventing deprecation warnings on recent PHP versions.
Happy to adjust the patch if needed.


Seldaek:
Thanks

PR #1526, opened 2021-01-06
Adds Mattermost handler

christingruber:
Added a Mattermost webhook handler, for example:

```php
$mattermostHandler = new MattermostWebhookHandler(
     'https://your.mattermost.server/hooks/YOUR-HOOK-ID',
     'Monolog',
     'town-square',
     'https://mattermost.org/wp-content/uploads/2016/04/icon.png'
);

$logger->pushHandler($mattermostHandler);

$logger->debug('Ping **test** :tada:');
```

Issue #306, opened 2014-01-19
WebHook handler


sagikazarmark:
This is something that is maybe not so important for developers as users.

The idea is that Monolog sends the record to a web hook. This is similar to the websocket idea #281 as the implementation is still needed on the other side, but there are some online services (like [Zapier](https://zapier.com/)) where this is not a problem and this is why I said this is a handler rather for users than developers.

PR #2010, opened 2025-12-11
Add generic webhook handler using HTTP(S) POST requests with a JSON payload

raphaelportmann:
I tried to implement a generic webhook handler using HTTP(S) POST requests with a JSON payload.
I have seen an old [discussion regarding the implementation of generic handlers](https://github.com/Seldaek/monolog/issues/306) and [using third-party packages for more specific handlers](https://github.com/Seldaek/monolog/pull/1526#issuecomment-756611693), but believe this handler would be a good addition to the monolog core package.
- As mentioned [here](https://github.com/Seldaek/monolog/issues/306) there are many plattforms like zapier and n8n (or building something yourself) that provide webhook endpoints to accept data in pretty much any format. Instead of creating more and more handlers for each of these plattforms or requiring third-party packages, this handler could be used for most of these plattforms.
- The handler currently extends the SocketHandler (adapted from the [FleepHookHandler integration](https://github.com/Seldaek/monolog/pull/415)). Is this still the preferred way instead uf using curl?
- The only handler that would provide similar functionality would be the [PSR-18 HTTP client integration](https://github.com/monolog-http/monolog-http), but requires another package and looks like maintenance lags behind.
- Didn't have the time to look into more features, like supporting secrets or other forms of authentication, but in that case it would probably be better to use PSR-18 clients directly. Is there a reason why this is a separate package but very specific integrations are part of the core?

My specific use-case would be to easily send logs to external webhooks - in my case inside a Laravel application, which allows for simple monolog configuration: e.g. https://github.com/laravel/laravel/blob/12.x/config/logging.php#L85
With this handler it could look like this:
```
'webhook' => [
    'driver' => 'monolog',
    'level' => env('LOG_LEVEL', 'warning'),
    'handler' => env('LOG_WEBHOOK_HANDLER', WebhookHandler::class),
    'handler_with' => [
        'url' => env('LOG_WEBHOOK_URL'),
    ],
],
// LOG_WEBHOOK_URL="https://example.com/webhook/abc123"
```

Just wanted to make a proposal. If you recommend another approach, let me know.

### DIFF
--- a/doc/02-handlers-formatters-processors.md
+++ b/doc/02-handlers-formatters-processors.md
@@ -45,6 +45,7 @@
 
 - [_SocketHandler_](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/SocketHandler.php): Logs records to [sockets](http://php.net/fsockopen), use this
   for UNIX and TCP sockets. See an [example](sockets.md).
+- [_WebhookHandler_](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/WebhookHandler.php): Logs records to a generic webhook URL using HTTP(S) POST requests with a JSON payload.
 - [_AmqpHandler_](https://github.com/Seldaek/monolog/blob/main/src/Monolog/Handler/AmqpHandler.php): Logs records to an [AMQP](http://www.amqp.org/) compatible
   server. Requires the [php-amqp](http://pecl.php.net/package/amqp) extension (1.0+) or
   [php-amqplib](https://github.com/php-amqplib/php-amqplib) library.
--- a/src/Monolog/Handler/WebhookHandler.php
+++ b/src/Monolog/Handler/WebhookHandler.php
@@ -1,0 +1,112 @@
+<?php declare(strict_types=1);
+
+/*
+ * This file is part of the Monolog package.
+ *
+ * (c) Jordi Boggiano <j.boggiano@seld.be>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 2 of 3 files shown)


raphaelportmann:
Alright, but before investing more time - do you think this handler is something that could be merged into the core as there are no generic HTTP handlers?
Will definitely look into migrating to curl. If you have other feedback for improvement also let me know.
Do you think integrating [PSR-18](https://www.php-fig.org/psr/psr-18/) HTTP Clients would be out-of-scope? Could also look into getting [monolog-http](https://github.com/monolog-http/monolog-http) up to date, but looks like the repo is not maintained actively.

Seldaek:
> * Is there a reason why this is a separate package but very specific integrations are part of the core?

Mostly historical reasons :)

> Do you think integrating [PSR-18](https://www.php-fig.org/psr/psr-18/) HTTP Clients would be out-of-scope? Could also look into getting [monolog-http](https://github.com/monolog-http/monolog-http) up to date, but looks like the repo is not maintained actively.

I would rather keep this in a separate lib which can be versioned separately, it just makes things a lot easier.

I'd first ping @gmponos if you're interested in maintaining it / bringing it back to speed maybe he can grant you access if he isn't interested anymore. Otherwise you can also fork it and publish an up to date fork.

IMO a generic http handler might be useful but I also see it's tricky to be truly generic as there are so many ways to authenticate, and to send payloads. You can grep the project for `Curl\Util::execute` and you'll find various handlers using curl right now for http.. But all of them are specific to some platform.

Issue #1865, opened 2024-01-03
Support more flexible normalizer


torreytsui:
Many times, I encountered use cases around custom normalization.

```php
NormalizerFormatter::normalize / is_object($data)
```

To name a few:

- Normalize a custom exception to include extra fields into the context
- Normalize a custom object to include extra fields into the context
- Redact fields from a custom object and selectively normalize non PII information into the context

I understand that we can write our Processor or a Formatter or implement the JsonSerializable to achieve that.

But I can also see that the "nomalization" logic was the main customisation we really needed in those use cases.

The `NormalizerFormatter` class is great and I don't want to duplicate its logic.

**What do you think if we make it more flexible and customizable?**

Suggestion 1: introduce protected normalizeObject() method

```php
// Monolog/Formatter/NormalizerFormatter.php

protected function normalize(mixed $data, int $depth = 0): mixed
{
    if (is_object($data)) {
        return $this->normalizeObject($data, $depth);
    }
}
```

Suggestion 2: introduce normalizer components

```php
// Monolog/Formatter/NormalizerFormatter.php

public function __construct(..., $normalizers);

protected function normalize(mixed $data, int $depth = 0): mixed
{
    // Loop through each normalizer, check its support, and invoke $normalizer->normalise($data, $depth)
}
```

Or other suggestions? or not really worth it?

cesarreyes3:
Would this help me if I want to customize the `file:line` format?
https://github.com/Seldaek/monolog/blob/479c936d2c230d8c467bdb3882afab45a6e6b8ad/src/Monolog/Formatter/LineFormatter.php#L261

Seldaek:
IMO you can already extend and override `normalize()`, handling any special types you want to handle, then defer to the parent method for the rest. I don't really see what a more complicated system would bring there given this isn't a very common case I think.

cobbles:
This may be a separate request but I thought I would ask on this thread in case it helps give context. We have a use case where we would like to limit the length of of the stack trace to a max length so that large logs to stdout do not get truncated and resulting in invalid json. 

Before I extend `NormaliserFormatter` is this a feature that would be desired?

If so I am happy to raise the Issue/PR for review

Seldaek:
But otherwise closing this issue as I don't think there's anything to address here.

PR #2015, opened 2026-01-07
feat(formatter): add maxTraceLength config option to NormalizerFormatter

cobbles:
Add maxTraceLength property to NormalizerFormatter to limit the number of stack trace frames included when normalizing exceptions.

Allows users who need control over their log length to avoid truncating the log and causing invalid json syntax

See comment here from about adding this feature https://github.com/Seldaek/monolog/issues/1865#issuecomment-3699854456


Seldaek:
Looks good yes, thanks

PR #2022, opened 2026-03-02
Fix missing timezone support in rotation date calculations

omniError:
The fix for #1982 is incomplete because the `getNextRotation` method is not aware of the time zone which was set in the constructor. Consider the scenario of daily rotation where the server is set to UTC but the constructor is set to another zone. This results in log entries being recorded unexpectedly to a different day's file during the time period where midnight has been reached in one zone but not the other.

This fix simply adds the existing time zone variable to the existing calculations.


Seldaek:
Thanks, makes sense

PR #2028, opened 2026-04-15
Fix HTTP status code modification while exception handling (#2027)

### LINKED ISSUE #2027: E_WARNING from http_response_code() in handleException when a status line has been registered via header() but not yet sent
Monolog version 3.9.0

Related issue: #1096 and #1106 

### Summary

`handleException` guards its call to `http_response_code(500)` with `!headers_sent()`, but this guard is insufficient. PHP maintains **three distinct phases** for HTTP headers, and the guard only covers two of them. When an exception is thrown during phase 2 (described below), the call produces a spurious `E_WARNING`:

```
E_WARNING: http_response_code(): Calling http_response_code() after header('HTTP/...') has no effect
```

### The three phases

**Phase 1 — No status code set yet**  
Neither `header('HTTP/...')` nor `http_response_code()` has been called. `headers_sent()` returns `false`. Calling `http_response_code()` works correctly as a "default value" that will be used if no explicit status line is ever registered.

**Phase 2 — Status line registered, but not yet written to the output stream**  
An explicit `header('HTTP/1.x NNN ...')` call has added a status line to PHP's internal header list (`SG(sapi_headers).headers`). The headers have **not** been flushed to the client yet, so `headers_sent()` still returns `false` — it reflects only whether bytes have been written to the output stream, not whether a status line has been staged.

At this point, calling `http_response_code()` has no effect on the actual response (the `header()` value wins) and PHP emits an `E_WARNING`. There is no PHP API to detect this state from userland without inspecting `headers_list()` manually.

**Phase 3 — Headers flushed to the output stream**  
Once any body output is produced, PHP flushes the staged headers first. Only now does `headers_sent()` return `true`. Both `header()` and `http_response_code()` become meaningless at this point.

### Why the current guard is insufficient

```php
// ErrorHandler.php ~L177
if (!headers_sent() && \in_array(...)) {
    http_response_code(500); // ← still fires during phase 2, causing E_WARNING
}
```

`!headers_sent()` is `true` in both phase 1 and phase 2. The guard therefore cannot distinguish between "safe to call" and "will produce a warning and have no effect".

Because `handleException` is invoked from a registered exception handler — which can fire at any point during request processing — it is impossible for user code to guarantee that no `header('HTTP/...')` call has occurred before the exception is thrown. This is not a usage error; it is a gap in the guard logic.

### Proposed fix

Check `headers_list()` for an existing HTTP status line before calling `http_response_code()`:

```php
if (!headers_sent() && !self::hasHttpStatusHeader() && \in_array(strtolower((string) \ini_get('display_errors')), ['0', '', 'false', 'off', 'none', 'no'], true)) {
    http_response_code(500);
}

private static function hasHttpStatusHeader(): bool
{
    foreach (headers_list() as $header) {
        if (stripos($header, 'HTTP/') === 0) {
            return true;
        }
    }
    return false;
}
```

Alternatively, a simpler one-liner using `array_filter`:

```php
$hasStatusHeader = (bool) array_filter(headers_list(), static fn($h) => stripos($h, 'HTTP/') === 0);

if (!headers_sent() && !$hasStatusHeader && \in_array(...)) {
    http_response_code(500);
}
```

### References

- [`SG(headers_sent)` in `SAPI.h`](https://github.com/php/php-src/blob/master/main/SAPI.h) — the flag that `headers_sent()` reads; set only when headers are actually written to the output stream.
- [`http_response_code()` implementation in `head.c`](https://github.com/php/php-src/blob/master/ext/standard/head.c) — checks only `SG(headers_sent)`, not whether a status line is already staged in `SG(sapi_headers).headers`.
- [php/php-src#18582](https://github.com/php/php-src/issues/18582) — upstream PHP issue acknowledging the incompatibility between `header('HTTP/...')` and `http_response_code()`.

tanakahisateru:
Fixes #2027

The `headers_sent()` guard does not prevent the `E_WARNING` from `http_response_code()`. The warning is emitted when `header('HTTP/...')` has already been called, but `headers_sent()` only becomes `true` after output is flushed to the stream — these are independent states. 

`headers_list()` might seem like an alternative to detect a previously set status line, but it intentionally excludes the HTTP status line.

Dropping the `headers_sent()` check and suppressing the warning with `@` is the most correct approach available: it sets 500 when possible and silently does nothing when it cannot.

### DIFF
--- a/src/Monolog/ErrorHandler.php
+++ b/src/Monolog/ErrorHandler.php
@@ -194,8 +194,8 @@ class ErrorHandler
             ($this->previousExceptionHandler)($e);
         }
 
-        if (!headers_sent() && \in_array(strtolower((string) \ini_get('display_errors')), ['0', '', 'false', 'off', 'none', 'no'], true)) {
-            http_response_code(500);
+        if (\in_array(strtolower((string) \ini_get('display_errors')), ['0', '', 'false', 'off', 'none', 'no'], true)) {
+            @http_response_code(500);
         }
 
         exit(255);


Seldaek:
I see the value of adding the @, but why remove the check? Sounds to me like it's still better to avoid creating an E_WARNING if we can (i.e. if headers are sent we know it is going to error so might as well skip it).

PR #2030, opened 2026-04-26
Fix TelegramBotHandler swallowing errors on non-JSON responses (#2026)

### LINKED ISSUE #2026: TelegramBotHandler: silent failure when API returns non-JSON response
## Summary

`TelegramBotHandler::sendCurl()` silently swallows errors when the Telegram API (or an intermediary proxy) returns a non-JSON response. The handler returns as if the message was sent successfully, with no exception and no indication of failure.

**Version:** 3.10.0

## Vulnerable code

[`TelegramBotHandler.php` lines 277–285](https://github.com/Seldaek/monolog/blob/b321dd6749f0bf7189444158a3ce785cc16d69b0/src/Monolog/Handler/TelegramBotHandler.php#L277-L285):

```php
$result = Curl\Util::execute($ch);
if (!\is_string($result)) {
    throw new RuntimeException('Telegram API error. Description: No response');
}
$result = json_decode($result, true);

if ($result['ok'] === false) {
    throw new RuntimeException('Telegram API error. Description: ' . $result['description']);
}
```

Three layers of failure compound:

1. `json_decode()` returns `null` for non-JSON input — no null check follows
2. `null['ok']` emits a warning and returns `null` in PHP 8+ — no type check
3. `null === false` evaluates to `false` — the error check is bypassed entirely

## Why it triggers

`curl_exec()` returns the response **body** for HTTP error status codes (502, 503, etc.), not `false`. The `is_string()` check on line 278 passes. The HTML/plaintext error page then proceeds to `json_decode`, which returns `null`.

Common real-world triggers:
- Cloudflare 502 Bad Gateway (returns HTML)
- Nginx error pages during Telegram API downtime
- Truncated responses from network timeouts
- Any proxy/CDN intercepting the request with a non-JSON error page

## Impact

The handler's purpose is critical alerting. When it fails silently during infrastructure issues — exactly when alerts matter most — log messages are permanently lost with zero indication. The Logger's try/catch (Logger.php:389) only works if an exception is thrown; since this bug suppresses the exception, the failure is completely invisible.

For comparison: SlackWebhookHandler, SendGridHandler, and IFTTTHandler don't attempt response validation at all. TelegramBotHandler is the only handler that validates, but gets it wrong for the non-JSON case.

## Suggested fix

```php
$result = json_decode($result, true);

if (!is_array($result)) {
    throw new RuntimeException('Telegram API error. Description: Unexpected non-JSON response');
}

if ($result['ok'] === false) {
    throw new RuntimeException('Telegram API error. Description: ' . ($result['description'] ?? 'Unknown error'));
}
```

Pablo1Gustavo:
Fixes #2026

## What it solves

`TelegramBotHandler::sendCurl()` was silently swallowing errors whenever the Telegram API or an upstream proxy returned a non-JSON body (Cloudflare 502 pages, Nginx error pages, truncated responses, etc.). `json_decode()` returned `null`, the guard `$result['ok'] === false` evaluated to `false` (since `null === false` is `false`), and no exception was raised — log records were lost during the exact infrastructure outages where alerting matters most.

## How it solves

- After `json_decode()`, reject anything that isn't a JSON object via `\is_array($result)`, throwing a `RuntimeException` with a clear `"Unexpected non-JSON response"` message.
- Replaced `$result['ok'] === false` with `($result['ok'] ?? null) !== true`. This also catches valid JSON responses that simply lack the `ok` key (e.g. proxy-injected error envelopes), which the old check would silently accept.
- Fall back to `"Unknown error"` when `description` is absent, preventing the previous undefined-key warning.
- Extracted the response validation into `validateApiResponse()` so each failure mode can be tested without a live HTTP call.

## Tests

Added a data-provider-driven test covering all six failure paths (curl failure, HTML body, empty body, `ok:false` with/without description, missing `ok` key) plus a happy-path assertion. The existing tests continue to pass.

### DIFF
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -256,7 +256,7 @@ class TelegramBotHandler extends AbstractProcessingHandler
         if ('' === trim($message)) {
             return;
         }
-        
+
         $ch = curl_init();
         $url = self::BOT_API . $this->apiKey . '/SendMessage';
         curl_setopt($ch, CURLOPT_URL, $url);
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

PR #2031, opened 2026-05-15
Fix AbstractProcessingHandler::handle() to call getBubble() instead of $this->bubble

philbates35:
## Summary

- `AbstractProcessingHandler::handle()` was reading the protected `$bubble` property directly (`false === $this->bubble`) rather than calling `getBubble()`
- This means subclasses that override `getBubble()` to dynamically control bubbling behaviour are silently bypassed
- Fix: replace with `!$this->getBubble()` so the getter is always honoured

## Context

This was discovered while working on a Symfony MonologBridge fix where a handler needed to override `getBubble()` to conditionally suppress propagation based on runtime state. The Symfony maintainers suggested the fix belongs upstream in Monolog: symfony/symfony#64207

### DIFF
--- a/src/Monolog/Handler/AbstractProcessingHandler.php
+++ b/src/Monolog/Handler/AbstractProcessingHandler.php
@@ -43,7 +43,7 @@ abstract class AbstractProcessingHandler extends AbstractHandler implements Proc
 
         $this->write($record);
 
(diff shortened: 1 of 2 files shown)


GromNaN (approved):

Seldaek:
Thanks

PR #2041, opened 2026-05-30
Add RedactingFormatter to automatically redact keys/sensitive data before logging records

### LINKED ISSUE #2018: Strip down data of a record context marked by SensitiveParameter
## Context

When logging, values passed as part of the record context may contain PHP objects that are normalized by a `Monolog\Formatter\FormatterInterface`:

```php
$logger->log(
  'My message',
  context: ['message' => $message, ...$additionalContext],
);
```

## Problem

Consider an object like this:

```php
final readonly class UpdateUser {
  function __construct(
    public string $id,``
    #[SensitiveParameter]
    public string $mySecretUrl,
  ) {}
}
```

Since PHP 8.2, `#[SensitiveParameter]` prevents `UpdateUser::$mySecretUrl` from appearing in stack traces. However, when such an object is passed as part of a log record context, this attribute is not respected and the sensitive value is still written to the logs.

## Proposal

Formatters could leverage `#[SensitiveParameter]` to automatically redact sensitive properties during normalization.

Based on https://github.com/symfony/symfony/pull/46183, this behaviour may not always be desired, so it could be implemented behind a configuration flag. That said, I would argue that redacting by default is the safer choice—frameworks like Symfony could then disable it in dev/debug mode if needed.

Additionally, since users often implement custom formatters, it would be helpful to expose this functionality in a reusable way.

Seldaek:
Closes #1983
Fixes https://github.com/Seldaek/monolog/issues/2018


src/Monolog/Formatter/RedactingFormatter.php
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * @param FormatterInterface $formatter     The formatter to wrap and whose output gets redacted
+     * @param list<string>       $sensitiveKeys Exact context/extra keys whose values to mask (case-insensitive)
+     * @param list<string>       $patterns      PCRE patterns to mask in the formatted output (e.g. self::TOKEN_PATTERN)
+     * @param string             $mask          Replacement token
+     *
+     * @throws \InvalidArgumentException If a pattern is not a valid PCRE regex
+     */
+    public function __construct(
+        private readonly FormatterInterface $formatter,
```
Seldaek, line 82: There maybe should be a way to access this from the outside so that Handlers that require a specific formatter can guarantee they get the right type or the right type wrapped in this?

src/Monolog/Formatter/RedactingFormatter.php
```diff
@@ -0,0 +1,304 @@
...
+        array $sensitiveKeys = ['password', 'passwd', 'pwd', 'secret', 'token', 'api_key', 'apikey', 'authorization', 'auth', 'cookie'],
+        array $patterns = [],
+        private readonly string $mask = '[REDACTED]',
+        private readonly bool $redactSensitiveParameters = true,
+        private readonly int $minSecretLength = 5,
```
theofidry, line 87: why does the secret require a minimum length?

src/Monolog/Formatter/RedactingFormatter.php
```diff
@@ -0,0 +1,304 @@
...
+        if ([] === $this->patterns && [] === $secrets) {
+            return $formatted;
+        }
+
+        if (\is_string($formatted)) {
+            if ([] !== $secrets) {
+                $formatted = str_replace($secrets, $this->mask, $formatted);
+            }
+            if ([] !== $this->patterns) {
+                $formatted = preg_replace($this->patterns, $this->mask, $formatted) ?? $formatted;
```
theofidry, line 290: doesn't that mean that if `preg_replace()` fails due to a PCRE limit we're falling back to the unredacted string, i.e. ending up with the secrets potentially leaking?

src/Monolog/Formatter/RedactingFormatter.php
```diff
@@ -0,0 +1,304 @@
...
+     */
+    private function collect(mixed $data, array &$secrets, array &$seen, int $depth): void
+    {
+        if ($depth > self::MAX_DEPTH) {
+            return;
+        }
+
+        if (\is_array($data)) {
+            foreach ($data as $key => $value) {
+                if (\is_string($key) && \in_array(strtolower($key), $this->sensitiveKeys, true)) {
```
theofidry, line 191: I think `Stringable`s should be handled here

src/Monolog/Formatter/RedactingFormatter.php
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    /**
+     * @param list<string> $secrets
+     */
+    private function addSecret(mixed $value, array &$secrets): void
+    {
+        // ints/floats are left out on purpose, sweeping a digit run out of the output
+        // would silently corrupt unrelated data
+        if (\is_string($value) && $value !== '' && \strlen($value) >= $this->minSecretLength) {
```
theofidry, line 268: same here (string -> Stringable)

src/Monolog/Formatter/RedactingFormatter.php
```diff
@@ -0,0 +1,304 @@
...
+     * @param  array<mixed> $data
+     * @return array<mixed>
+     */
+    private function redactKeys(array $data): array
+    {
+        foreach ($data as $key => $value) {
+            if (\is_string($key) && \in_array(strtolower($key), $this->sensitiveKeys, true)) {
+                $data[$key] = $this->mask;
+            } elseif (\is_array($value)) {
+                $data[$key] = $this->redactKeys($value);
```
theofidry, line 136: doesn't this mean that a self-referencing array can cause an infinite recursion here?

src/Monolog/Formatter/RedactingFormatter.php
```diff
@@ -0,0 +1,304 @@
...
+    {
+        $class = $data::class;
+        if (isset($this->sensitiveProperties[$class])) {
+            return $this->sensitiveProperties[$class];
+        }
+
+        $properties = [];
+        foreach (Utils::getSensitiveParameterNames($class) as $name => $_) {
+            try {
+                $properties[$name] = new \ReflectionProperty($class, $name);
```
theofidry, line 252: The method claims to handle a case such as:

```php
class ParentCredentials
{
    public function __construct(
        #[SensitiveParameter]
        private string $token,
    ) {}

    public function __toString(): string
    {
        return "token={$this->token}";
    }
}

 class Credentials extends ParentCredentials {}
```

But that reflection call right here wouldn't find `$token`, since it's a private property of the parent class, so is that scenario really handled?

src/Monolog/Formatter/RedactingFormatter.php
Seldaek: This is explained in the docblock above, to avoid replacing random strings that are very short if you pass something like "foo" in a sensitive parameter you maybe get collateral damage, and no secret worth their salt is that short anyway.

src/Monolog/Formatter/RedactingFormatter.php
Seldaek: https://github.com/Seldaek/monolog/pull/2041/changes/540a9c44d28b5c483edc98f7a27e19a270f42f4f

src/Monolog/Formatter/RedactingFormatter.php
Seldaek: https://github.com/Seldaek/monolog/pull/2041/changes/b117127f327fe04c722e9aa884153fb8f735ede1

src/Monolog/Formatter/RedactingFormatter.php
Seldaek: https://github.com/Seldaek/monolog/pull/2041/changes/3430bd9db3d65e61e4cde4d6b12790e248ed8826

src/Monolog/Formatter/RedactingFormatter.php
Seldaek: https://github.com/Seldaek/monolog/pull/2041/changes/dafa0021bcf60c830e909640cf16f7670ec608e1

src/Monolog/Formatter/RedactingFormatter.php
Seldaek: https://github.com/Seldaek/monolog/pull/2041/changes/78b574178d2976557cb9bd90faf1fc3144e7808a

PR #2042, opened 2026-06-26
Still dump the scalars, when the depth level is reached

profuel:
Hello, community!

We noticed that some of our objects are of exactly depth 9.
We though about changing the level to 10, but then I noticed that this change might be a better option.

What do you think?

Thank you, AndriyT


Seldaek:
Makes sense thanks

profuel:
Thank you!!!

PR #2049, opened 2026-07-19
fix(SyslogUdp): allow configuring the max UDP datagram length

### LINKED ISSUE #1826: SyslogUdp logs gets truncated
Monolog version 2.9.1 in a Laravel 9 environment on Ubuntu 22.
I use the SyslogUdpHandler to send logs via Syslog to third-party systems.

It works, but when the log message is longer, then it gets truncated. What can be the reason for this, and is there a way to not have it truncated?

If I log the message just before the `socket_sendto($this->getSocket(), $chunk, strlen($chunk), $flags = 0, $this->ip, $this->port);` call in `UdpSocket.php`, then I still see the correct length.

However, if I monitor it with tcdump, what I see is a single packet and that the message is truncated when it gets sent over:
`
<134>1 2023-08-02T12:45:41+00:00 myhost myapp 65630 - - Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas placerat justo ut imperdiet fringilla. Nam nec mauris ut ligula scelerisque lobortis laoreet quis nulla. Mauris malesuada augue lorem, a volutpat eros varius sit amet. Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Suspendisse tincidunt in dolor non consequat. Morbi congue ipsum porta arcu eleifend condimentum. Mauris tincidunt, mi ac suscipit consequat, magna diam accumsan augue, id tristique ex mauris vitae mauris. Fusce nec ipsum eros. Donec condimentum, sapien ut interdum pharetra, nisl est vehicula purus, quis sagittis magna purus eget leo. Duis dapibus nulla vel diam volutpat ullamcorper. Pellentesque scelerisque, orci eget dictum consequat, arcu magna fringilla eros, eget eleifend orci lectus et mi. Interdum et malesuada fames ac ante ipsum primis in faucibus. Proin nec lorem fringilla enim semper sollicitudin. Proin non convallis 
`

Since the actual log message is a JSON, it can't be parsed on the receiver if it's not complete.

nabsei:
## What

`UdpSocket::DATAGRAM_MAX_LENGTH` (65023 bytes) is the largest a UDP payload can theoretically be, but a datagram that size gets fragmented at the IP level, and many routers/firewalls drop fragmented UDP packets outright. That silently truncates or loses log messages that exceed the path MTU on the way to the syslog receiver, even though nothing on the sending side reports an error.

## Why this approach

An existing test (`testLongMessagesAreTruncated`) explicitly locks in the current 65023-byte behavior, and some setups over trusted/local networks may rely on it, so I didn't want to just lower the default and risk a behavior change / BC break. Instead, this adds an optional `$maxLength` constructor parameter to `UdpSocket`, threaded through `SyslogUdpHandler`'s constructor (the entry point most users actually use), so a value that fits within the path MTU (eg. 1024) can be opted into. The default is unchanged.

## Verification

- Added regression tests covering: a custom `$maxLength` truncates at the new length, the default is preserved when not specified, and `SyslogUdpHandler` correctly threads the value through to its internal `UdpSocket` (verified via reflection, since `$socket` isn't otherwise inspectable from outside).
- Full test suite passes (1182 tests); the only pre-existing failures are 3 unrelated ones in `JsonFormatterTest`/`NormalizerFormatterTest`/`ErrorHandlerTest` (confirmed identical on unmodified `main`), plus 30 skips for missing optional extensions (mongodb, redis, etc.) unrelated to this change.
- `phpstan analyse` is clean on the changed files.

**Honest caveat**: I could not verify the actual real-world symptom (fragmented UDP getting dropped by a router/firewall) with a live network reproduction — that's not something reproducible over loopback, which doesn't share the same MTU/fragmentation-handling constraints as a real network path. Verification here is limited to code review and the unit tests described above.

## Related issue

Fixes #1826

---
Disclosure: this PR was prepared with the assistance of Claude Code (Anthropic). The investigation, fix, and verification steps described above were done and checked by me.

### DIFF
--- a/src/Monolog/Handler/SyslogUdp/UdpSocket.php
+++ b/src/Monolog/Handler/SyslogUdp/UdpSocket.php
@@ -20,12 +20,22 @@ class UdpSocket
 
     protected string $ip;
     protected int $port;
+    protected int $maxLength;
     protected ?Socket $socket = null;
 
-    public function __construct(string $ip, int $port = 514)
+    /**
+     * @param ?int $maxLength Maximum size in bytes of a single UDP datagram sent to $ip:$port.
+     *                        Defaults to the largest a UDP payload can theoretically be, but
+     *                        messages that size get fragmented at the IP level, and many routers
+     *                        and firewalls drop fragmented UDP packets, which silently truncates
+     *                        or loses the log entry on the receiving end. Passing a value that
+     *                        fits within the path MTU (eg. 1024) avoids that.
+     */
+    public function __construct(string $ip, int $port = 514, ?int $maxLength = null)
     {
         $this->ip = $ip;
         $this->port = $port;
+        $this->maxLength = $maxLength ?? self::DATAGRAM_MAX_LENGTH;
     }
 
     public function write(string $line, string $header = ""): void
@@ -70,7 +80,7 @@ class UdpSocket
 
     protected function assembleMessage(string $line, string $header): string
     {
-        $chunkSize = static::DATAGRAM_MAX_LENGTH - \strlen($header);
+        $chunkSize = $this->maxLength - \strlen($header);
 
         return $header . Utils::substr($line, 0, $chunkSize);
     }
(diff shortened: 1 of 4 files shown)


Seldaek:
Thanks

PR #2050, opened 2026-07-23
Make IntrospectionProcessor SKIP_FUNCTIONS extendable

mhalavach:
SKIP_FUNCTIONS was made protected in https://github.com/Seldaek/monolog/pull/1899, so subclasses could override it, but IntrospectionProcessor still referenced it via `self::`, which resolves the constant at the declaring class and ignores any override.


Seldaek:
Thanks

PR #2056, opened 2026-08-04
Add FrankenPhpHandler

dunglas:
Adds a handler for [FrankenPHP's `frankenphp_log()`](https://frankenphp.dev/docs/logging/#frankenphp_log), which writes structured logs through FrankenPHP's Go `log/slog`-based logger instead of `error_log()`.

- Guards on `function_exists('frankenphp_log')` in the constructor, throwing `MissingExtensionException` outside FrankenPHP (same pattern as `ZendMonitorHandler`).
- Maps Monolog's 8 levels onto slog's level scale: the official `FRANKENPHP_LOG_LEVEL_*` constants for Debug/Info/Warning/Error, and custom ints for Notice/Critical/Alert/Emergency following slog's documented gap-of-4 convention for intermediate levels (see the docblock on `FrankenPhpHandler::toFrankenPhpLevel()`).
- `FrankenPhpHandlerE2ETest` (tagged `#[Group('e2e')]`) makes real HTTP calls to a FrankenPHP server and inspects its structured log output, skipping itself if no server is reachable. CI runs it against `dunglas/frankenphp:latest` via a `services:` container in `.github/workflows/frankenphp-e2e.yml`.

### DIFF
--- a/src/Monolog/Handler/FrankenPhpHandler.php
+++ b/src/Monolog/Handler/FrankenPhpHandler.php
@@ -76,9 +76,10 @@ class FrankenPhpHandler extends AbstractProcessingHandler
      */
     protected function writeFrankenPhpLog(string $message, int $level, array $context): void
     {
-        // frankenphp_log() merges $context into the same JSON object as its own "level" severity field;
-        // dropping Monolog's raw int here avoids a duplicate "level" key (level_name already carries it).
-        unset($context['level']);
+        // frankenphp_log() merges $context into the same JSON object it already fills with "msg",
+        // "level" and "ts", so drop Monolog's duplicates of those three. level_name still carries
+        // the Monolog level name, which slog's own severity cannot express.
+        unset($context['message'], $context['level'], $context['datetime']);
 
         \frankenphp_log($message, $level, $context);
     }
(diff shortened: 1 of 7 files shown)


src/Monolog/Handler/FrankenPhpHandler.php
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        parent::__construct($level, $bubble);
+    }
+
+    /**
+     * Translates Monolog log levels to FrankenPHP's slog-based levels.
+     *
+     * slog treats levels as arbitrary ints and recommends a gap of 4 between named levels to leave
+     * room for intermediate ones, e.g. Google Cloud Logging's Notice between Info and Warn.
```
Seldaek, line 48: I'm not sure I get the gap of 4 and then is uses gaps of two below between critical/alert/emergency. Probably a misunderstanding but anyway.

Seldaek:
Some of the CI failures are my bad and a rebase will fix them, but some are legit problems with this PR (zizmor job, FrankenPhpHandlerE2ETest failures in CI jobs)

src/Monolog/Handler/FrankenPhpHandler.php
Seldaek: Fixed in https://github.com/Seldaek/monolog/pull/2056/commits/1fc62e6479d6fb521a45acd5dfdf8d6ceb26b0d3 but happy to get your feedback on this still (and the other fixes).

Seldaek:
Thanks

PR #2057, opened 2026-08-16
Clarify handler stack traversal direction in docs (#1583)

### LINKED ISSUE #1583: DOCS: either example is wrong or explanation of handler stack is unclear
This is an issue in the docs.

From here: https://github.com/Seldaek/monolog/blob/main/doc/01-usage.md#core-concepts

> Every Logger instance has a channel (name) and a stack of handlers. Whenever you add a record to the logger, it **traverses the handler stack**. Each handler decides whether it fully handled the record, and if so, the propagation of the record ends there.

Question: traverses the stack in what direction? **Bottom to top or top to bottom?** Note that the concepts of "bottom" and "top" or "up" and "down" in a stack are well-defined: what you add first is at the bottom, what you add last is at the top, that is by definition. In other words, "pushing" to the stack means adding at the top, "popping" means taking from the top.
**Let's read on and see if the question is answered.** 

> This allows for flexible logging setups, for example having a **StreamHandler at the bottom** of the stack that will log anything to disk, and **on top of that add a MailHandler** that will send emails only when an error message is logged.

Ok so, in the example, the StreamHandler is explicitly said to be at the bottom of the stack, and the MailHandler is said to be on top of it. **My intuition** would tend to suggest that the **StreamHandler is called first**, given that in this example it logs _everything_, and then it passes the message on to the "next" handler which is the MailHandler, which only handles a subset of the messages. So, this would mean that the stack is traversed **from bottom to top**.

**However**, then it continues:

> Handlers also have a **$bubble property** which defines **whether they block the record** or not if they handled it.

`<ASIDE>`
Now first of all (and this is besides the point of this issue), when you say that there is a **boolean property _X_** that determines **whether or not _something_**, you should express it in such a way that when _X_ is true, _something_ is true. In this case, it is reversed. Obviously, **`$bubble=true`** means it bubbles, that is, it **does _not_ block** the record. But I only know that because I am already familiar with the concept of bubbling (in other contexts) and I know that bubbling is the opposite of blocking (when you block something, you stop it from bubbling). If I only read `$bubble` as a variable name with no intrinsic meaning and were to deduce the behavior from that sentence, I would either get it wrong (i.e. `$bubble=true` means to block the message), or I would not know (as in "`$block` defines _whether or not_ to block: which one is true and which one is false?").
`</ASIDE>`

Ok let's ignore that and get back to the issue.

I was saying, it continues:

> Handlers also have a **$bubble property** which defines **whether they block the record** or not if they handled it.
> In this example, setting the MailHandler's $bubble argument to false means that records handled by the MailHandler will not propagate to the StreamHandler anymore.

Now I am confused, because I was assuming that the stack was being traversed bottom-up, so, changing `$bubble` to false for the MailHandler wouldn't make a difference in this example (not as far as the two handlers mentioned are concerned).

So, assuming it's not an error in the docs, it must mean my previous assumptions were wrong.

Then the stack is traversed from the top down. Then it makes sense: the MailHandler is met first, and if its `$bubble` property is false, the message does not propagate to the StreamHandler below it.

But then it would mean that **bubbles travel from the top down**, which is counterintuitive as hell. In all the contexts where I've ever seen the concept of bubbling (e.g. event frameworks), bubbling is always conventionally pictured as going up, not down. That's why it's called bubbling, because bubbles in real life usually go upwards (at least air bubbles in a liquid).

So now I'm led to seriously question if you actually made a mistake somewhere and switched "top" with "bottom" or "StreamHandler" with "MailHandler" somewhere.
If not, then it means that the stack is traversed from the top down (which is not clearly stated and one has to deduce through an awful lot of reasoning), and that that the verb "to bubble" is used, counterintuitively, to depict the passing of message in the downwards direction. 

_Or_ it is an upside-down stack, one that grows downwards.

Note that I'm not being pedantic. I am genuinely not sure, at this stage, in which order I need to push the handlers into the stack to get the desired result.

wakqasahmed:
Fixes #1583

## Summary

The "Core Concepts" section of `doc/01-usage.md` described the handler stack
without stating which direction it is traversed, and phrased the `$bubble`
example in a way that reads as reversed / self-contradictory (as laid out in
detail in #1583).

This PR:
- States explicitly that the stack is traversed top to bottom (last-added
  handler runs first, first-added handler runs last).
- Rewrites the `StreamHandler` / `MailHandler` example so "added first, sits at
  the bottom" and "added on top, called first" are consistent with that
  traversal order.
- Rephrases the `$bubble` explanation so it states the default (`true`) and
  what `false` changes, instead of the ambiguous "whether they block the
  record" wording.

No behavior change — docs only.

## Why this wording

This matches the maintainer's own confirmation in the issue thread
(https://github.com/Seldaek/monolog/issues/1583#issuecomment-919428670):

> The stack is then traversed top to bottom, as per this code... The top
> level handlers all receive the records in order (top to bottom of stack)
> until bubbling is interrupted.
>
> ...I don't really see what can be done to fix it except documenting it
> better... If you feel like sending a PR with doc improvements now that you
> hopefully grasp things better, that would be great.

## Test plan

- [x] Docs-only change, no code/tests affected
- [x] Re-read the updated section for consistency with surrounding terminology

### DIFF
--- a/doc/01-usage.md
+++ b/doc/01-usage.md
@@ -21,17 +21,21 @@ composer require monolog/monolog
 ## Core Concepts
 
 Every `Logger` instance has a channel (name) and a stack of handlers. Whenever
-you add a [record](message-structure.md) to the logger, it traverses the handler stack. Each handler
-decides whether it fully handled the record, and if so, the propagation of the
-record ends there.
+you add a [record](message-structure.md) to the logger, it traverses the handler
+stack from top to bottom: the handler that was added *last* is called first, and
+the handler that was added *first* is called last. Each handler decides whether
+it fully handled the record, and if so, the propagation of the record stops
+there and handlers lower in the stack never see it.
 
-This allows for flexible logging setups, for example having a `StreamHandler` at
-the bottom of the stack that will log anything to disk, and on top of that add
-a `MailHandler` that will send emails only when an error message is logged.
-Handlers also have a `$bubble` property which defines whether they block the
-record or not if they handled it. In this example, setting the `MailHandler`'s
-`$bubble` argument to false means that records handled by the `MailHandler` will
-not propagate to the `StreamHandler` anymore.
+This allows for flexible logging setups, for example adding a `StreamHandler`
+first so that it sits at the bottom of the stack and logs everything to disk,
+then adding a `MailHandler` on top of it so that it is called first and only
+sends emails when an error message is logged. Handlers also have a `$bubble`
+property, which defaults to `true` and means the record keeps propagating down
+to the next handler even after this one handled it. In this example, setting
+the `MailHandler`'s `$bubble` argument to `false` means that once the
+`MailHandler` has handled a record, it will *not* propagate down to the
+`StreamHandler` anymore.
 
 You can create many `Logger`s, each defining a channel (e.g.: db, request,
 router, ..) and each of them combining various handlers, which can be shared
--- a/doc/01-usage.md
+++ b/doc/01-usage.md
@@ -22,20 +22,24 @@ composer require monolog/monolog
 
 Every `Logger` instance has a channel (name) and a stack of handlers. Whenever
 you add a [record](message-structure.md) to the logger, it traverses the handler
-stack from top to bottom: the handler that was added *last* is called first, and
-the handler that was added *first* is called last. Each handler decides whether
-it fully handled the record, and if so, the propagation of the record stops
-there and handlers lower in the stack never see it.
+stack from top to bottom. `pushHandler()` puts a handler on top of the stack, so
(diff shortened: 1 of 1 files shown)


wakqasahmed:
Hi @Seldaek — noticed this PR doesn't have a reviewer assigned yet — it's been about 4 days, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look when you get a chance, or point me to who should? Thanks!

Seldaek:
I'll get to it when I get to it thanks.

wakqasahmed:
Understood, no rush — thanks for taking a look when you can.

Seldaek:
Thanks

PR #2058, opened 2026-08-18
Replace glob() with RecursiveDirectoryIterator in RotatingFileHandler cleanup (#1784)

### LINKED ISSUE #1784: glob() not finding the existing log -> Log rotation doesn't delete files
Monolog version 1|2|3? it's version 2

Related to https://github.com/Seldaek/monolog/issues/204#issuecomment-19757534

I use the drupal monolog module which is using this monolog library.

In the following line I get an empty array:

https://github.com/Seldaek/monolog/blob/2.8.0/src/Monolog/Handler/RotatingFileHandler.php#L142

for the path "private://logs/debug.log" because glob can't compute the right paths. The files definitely exist. If I replace the path with an absolute one, it works, but then I'd have to adjust the paths for each and every environment which is tedious :-/

Maybe we can think about using something else than glob() here? Or is there something I can do?

See here as well: https://www.drupal.org/project/monolog/issues/3326496

Thanks in advance!

wakqasahmed:
Fixes #1784

## Problem

`RotatingFileHandler::rotate()` used `glob()` to find previously-rotated log files for cleanup. `glob()` does not support PHP stream wrappers (e.g. Drupal's `private://` scheme), so when the handler is given a stream-wrapper path, cleanup silently finds zero files and old rotated logs are never deleted, even though the current log file itself writes fine (since `StreamHandler` uses `fopen()`, which does support stream wrappers).

## Fix

This follows the approach @Seldaek sketched directly in the issue thread: replace the `glob()` call with `RecursiveDirectoryIterator` + `RecursiveIteratorIterator`, matching entries against a regex built from the same filename/date tokens `getGlobPattern()` used to build its glob pattern.

Implementation notes:
- Added `RotatingFileHandler::findRotatedFiles()`, called from `rotate()` instead of `glob($this->getGlobPattern())`. `getGlobPattern()` itself is left untouched for backwards compatibility with any subclasses that call/override it.
- Per Seldaek's caution about computing the base directory: `pathinfo()` collapses a stream-wrapper's `scheme://` down to a single trailing slash. The existing `getTimedFilename()`/`getGlobPattern()` code already relies on re-appending a `/` via string concatenation to restore the double slash — `findRotatedFiles()` mirrors that same technique for the directory passed to `RecursiveDirectoryIterator` and for the regex prefix, rather than deriving the base dir a different way.
- The regex is built by escaping the literal parts of `filenameFormat` with `preg_quote()` and substituting the date tokens (`Y`/`y`/`m`/`d`/`H`) with digit-count regex classes, so directory-nesting date formats like `Y/m/d/H` (already covered by `testRotationWithFolderByDate`) keep working via the recursive walk.
- Kept the existing `is_writable()`/`unlink()`/directory-cleanup logic in `rotate()` unchanged; only the file-discovery step changed.

## Test plan

- Added `RotatingFileHandlerTest::testRotationPrunesOldFilesThroughStreamWrapper()`, which registers a minimal stream wrapper fixture (`tests/Monolog/Handler/Fixtures/StreamWrapperPassthrough.php`) that proxies stream/directory operations to a real directory but is not resolvable via `glob()` — mirroring how Drupal's `private://` wrapper behaves. The test asserts old rotated files are actually pruned through the wrapper, which is the exact scenario from #1784.
- Existing filesystem-based tests (`testRotation`, `testRotationWithFolderByDate`, `testRotationWhenSimilarFileNamesExist`, etc.) exercise the regular non-wrapper path and were traced by hand against the new regex-matching logic to confirm unchanged behavior (same file selection/ordering/pruning as before).

**Disclosure:** I do not have a PHP runtime available in the environment I used to write this patch, so none of these tests (new or pre-existing) have actually been executed locally. I traced the logic carefully by hand instead. Please treat this as unverified until CI runs — happy to iterate on any failures.

### DIFF
--- a/src/Monolog/Handler/RotatingFileHandler.php
+++ b/src/Monolog/Handler/RotatingFileHandler.php
@@ -189,18 +189,13 @@ class RotatingFileHandler extends StreamHandler
     }
 
     /**
-     * Appends a trailing "/" to a pathinfo() dirname, restoring a stream
-     * wrapper URL's second slash when needed.
+     * Appends the trailing "/" to a pathinfo() dirname.
      *
-     * pathinfo() collapses "scheme://subdir" down to "scheme://subdir" (both
-     * slashes kept) but "scheme://file" with no subdirectory down to a bare
-     * "scheme:" (both slashes dropped) — there's no path segment left for the
-     * second slash to attach to. Appending a single "/" is correct for the
-     * former (produces "scheme://subdir/") but wrong for the latter (produces
-     * "scheme:/", a single-slash form PHP's stream wrapper dispatch does not
-     * recognise, so fopen() silently falls through to treating it as a local
-     * relative path instead of routing to the wrapper). Detecting the bare
-     * "scheme:" form and appending "//" instead restores "scheme://".
+     * pathinfo() reduces a wrapper URL with no subdirectory ("scheme://foo.log")
+     * to a bare "scheme:", and PHP does not route the single-slash "scheme:/"
+     * to the wrapper at all, so that case needs both slashes back. The pattern
+     * deliberately requires two characters before the colon so a Windows drive
+     * letter is left alone.
      */
     private static function appendDirectorySeparator(string $dirName): string
     {
@@ -315,7 +310,7 @@ class RotatingFileHandler extends StreamHandler
                 ['[0-9][0-9][0-9][0-9]', '[0-9][0-9]', '[0-9][0-9]', '[0-9][0-9]', '[0-9][0-9]'],
                 $this->dateFormat
             )],
-            ($fileInfo['dirname'] ?? '') . '/' . $this->filenameFormat
+            self::appendDirectorySeparator($fileInfo['dirname'] ?? '') . $this->filenameFormat
         );
         if (isset($fileInfo['extension'])) {
             $glob .= '.'.$fileInfo['extension'];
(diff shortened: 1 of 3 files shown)


Seldaek:
Please rebase to get the fixed CI - the main branch was broken sorry about that.

wakqasahmed:
Rebased onto latest main (picked up the CI fix). Should be running against fixed CI now.

Seldaek:
CI failures are in this PR now

wakqasahmed:
Found it — the rebase surfaced a real gap in the stream-wrapper handling when the dirname has no subpath. Pushed a fix; CI is green across all PHP/lint/PHPStan/ES matrix jobs now.

Seldaek:
Alrighty, had Claude do a review and fix a few more things, see individual commits. The most important change is converting glob to regex manually so that we don't break BC if getGlobPattern was overridden by someone.

Feel free to take a look make sure it looks good to you too, but to me it seems mergeable now.

wakqasahmed:
Went through all the follow-up commits — the unreadable-directory guard, the Windows separator normalization, deriving the scan from getGlobPattern() instead of duplicating it, and the flat-iterator perf fix all look right to me, and each is a real bug the original version had. CI is green across the whole matrix. Looks good to merge from my side.

Seldaek:
Thanks

PR #2061, opened 2026-09-01
Report stack trace frames that have no file/line instead of skipping them

### LINKED ISSUE #1736: NormalizerFormatter skips some stacktraces
Monolog version 2|3

Check the function.

https://github.com/Seldaek/monolog/blob/5579edf28aee1190a798bfa5be8bc16c563bd524/src/Monolog/Formatter/NormalizerFormatter.php#L230-L235

PHP Fatal Error caught by `register_shutdown_function()` has `trace` field. Inside this `trace` field there can be few elements without `file` and `line` fields.

Here is an example:

```php
# php test.php 
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/test.php on line 27
array(1) {
  [0]=>
  array(1) {
    ["function"]=>
    string(16) "shutdownFunction"
  }
}
```

Run this script and you will see the following output of the `$e->getTrace()` method:

You may change `register(false)` to `register(true)` and check the results:

```
# php test.php 
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 20480 bytes) in /var/www/html/test.php on line 28
array(2) {
  [0]=>
  array(3) {
    ["file"]=>
    string(22) "/var/www/html/test.php"
    ["line"]=>
    int(11)
    ["function"]=>
    string(16) "shutdownFunction"
  }
  [1]=>
  array(3) {
    ["function"]=>
    string(14) "shutdownMethod"
    ["class"]=>
    string(12) "ErrorHandler"
    ["type"]=>
    string(2) "::"
  }
}
```

All that records without `file` and `line` will be skipped and engineer can't understand the full picture.

If you agree with this I may provide PR.

Amoifr:
Fixes #1736

`normalizeException()` only kept trace frames that have both `file` and `line`. Frames called by the engine itself have neither: shutdown functions, callbacks run by internal functions like `array_map` or `usort`, destructors. They were dropped silently, so a trace could look shorter than it was, and for a fatal error caught in a shutdown function it disappeared entirely:

```php
register_shutdown_function(function () {
    // ... catch the fatal, log it
});
```

```
before: "(no trace key at all)"
after:  ["internal[{closure:/app/src/shutdown.php:5}]:0"]
```

Those frames are now reported by name, in the shape you and @WoZ settled on in the issue: `internal[Class.function]:0`, with `.` rather than `->` or `::` as @WoZ suggested for static methods. The data type does not change, the trace stays a list of strings.

Three decisions worth your opinion rather than mine:

- **The class prefix is unconditional, closures included.** On PHP 8.4+ that reads long, `Test.{closure:Test::run():12}`, because the closure name already repeats the scope. I kept it because on 8.1 to 8.3 the name is just `{closure}` and the class is then the only context there is. Checked on 8.2, 8.3, 8.4 and 8.5.
- **The base path is stripped from the label too.** Since PHP 8.4 a closure declared at file scope is named after its file, so reporting these frames would put an absolute path in the logs of anyone using `setBasePath()`. Three lines and one test, dropped easily if you would rather keep the label verbatim.
- **One existing test had to change.** `testExceptionTraceWithArgs` asserted on `trace[0]`, which is now the `usort` comparison closure. The assertion moved to `trace[1]` and a new one covers `trace[0]`.

Tests: the suite goes from 1268 to 1271, no failure, same skips and deprecations as on `main`. PHPStan clean. The PHP 8.5 deprecation that remains is the pre-existing `ArrayObject::__construct` one in the `__PHP_Incomplete_Class` branch.

### DIFF
--- a/src/Monolog/Formatter/NormalizerFormatter.php
+++ b/src/Monolog/Formatter/NormalizerFormatter.php
@@ -317,6 +317,20 @@ class NormalizerFormatter implements FormatterInterface
                     $file = preg_replace('{^'.preg_quote($this->basePath).'}', '', $file);
                 }
                 $data['trace'][] = $file.':'.$frame['line'];
+            } else {
+                // Frames called by the engine itself have no file/line: shutdown functions,
+                // callbacks run by internal functions, destructors. Skipping them made traces
+                // look shorter than they were, and entirely empty for fatal errors caught in a
+                // shutdown function, so they are reported by name instead.
+                $call = $frame['function'] ?? 'unknown';
+                if (isset($frame['class'])) {
+                    $call = $frame['class'].'.'.$call;
+                }
+                if ($this->basePath !== '') {
+                    // closure names embed the file they were declared in since PHP 8.4
+                    $call = preg_replace('{'.preg_quote($this->basePath).'}', '', $call);
+                }
+                $data['trace'][] = 'internal['.$call.']:0';
             }
         }
 
--- a/src/Monolog/Formatter/NormalizerFormatter.php
+++ b/src/Monolog/Formatter/NormalizerFormatter.php
@@ -324,8 +324,11 @@ class NormalizerFormatter implements FormatterInterface
                 // shutdown function, so they are reported by name instead.
                 $call = $frame['function'] ?? 'unknown';
                 if (isset($frame['class'])) {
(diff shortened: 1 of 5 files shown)


Seldaek:
Thanks

PR #2065, opened 2026-09-03
Allow reading the timestamp of log records from a PSR-20 clock

### LINKED ISSUE #1902: Allow to specify time for log records

Amoifr:
Fixes #1902

The PSR-3 methods have no `$datetime` parameter, so today the only way to decide what "now" means is to call `addRecord()` directly. This adds an optional PSR-20 clock to the logger, plus `setClock()` / `getClock()` mirroring the `setTimezone()` / `getTimezone()` pair that is already there:

```php
$logger = new Logger('app', [$handler], [], null, new MockClock('2024-03-05 11:22:33'));
```

The two points @stof raised on the previous attempt (#1938) drove the shape of this one. The clock parameter is added at the end of the constructor with a `null` default, so no existing signature changes and subclasses that redefine the constructor stay valid. And no class is added, in particular none named `Clock`: Monolog consumes a clock, it does not provide one.

**Reporting the instant onto the record was the part that needed care.** `LogRecord::$datetime` is a `JsonSerializableDateTimeImmutable`, whose private property is only initialised by the constructor, so `createFromFormat()` leaves it uninitialised (`Error: Typed property must not be accessed before initialization`). The instant is therefore applied to an instance that is already built, and two shortcuts had to be dropped after measuring them:

- `'@'.$now->format('U.u')` is wrong before 1970. `1950-01-02 03:04:05.123456` comes back as `03:04:04.876544`, the fraction being added to a negative timestamp.
- Going through a local wall clock time is ambiguous inside the repeated hour of a DST fall back.

So the code calls `setTimestamp()` to reach the right second, then adds the microseconds as a *relative* amount of time. Neither step ever goes through a local wall clock time, and both keep the named timezone configured on the logger (`Europe/Paris`, not `+01:00`). Covered by a data provider over: a modern instant, before the epoch, the epoch itself, microseconds at 0, 1 and 999999, the ambiguous hour of a Paris DST fall back, and a clock ticking in another timezone.

`add()` rather than `modify()`, because PHPStan also runs on PHP 8.1 where `modify()` returns `static|false`, which would not survive level 8.

**One call for you:** `psr/clock` is in `require-dev` and `suggest`, not in `require`. The type only resolves when a non-null value is passed, and having a clock instance to pass implies the package is already installed, so this costs nothing to the existing installs. Promoting it to `require` is a one line change if you would rather declare it properly, and `$clock` is `protected` for consistency with `$timezone`, happy to make it private if you prefer.

`isHandling()` builds a throwaway record too and now goes through the same helper, so a handler that looks at the datetime sees the same time in both paths.

Tests: the suite goes from 1276 to 1289, no failure, same skips and deprecations as on `main`. PHPStan clean. The `Timezone ID 'CET' is invalid` notice on PHP 8.5 is pre-existing in `testTimezoneIsRespectedInOtherTimezone`. PHP 8.1 is the one version of the matrix I could not replay locally.

Documented under "Configuring a logger" in `doc/01-usage.md`. `CHANGELOG.md` left alone since you write it at release time.

### DIFF
--- a/src/Monolog/Logger.php
+++ b/src/Monolog/Logger.php
@@ -16,6 +16,7 @@ use DateTimeZone;
 use Fiber;
 use Monolog\Handler\HandlerInterface;
 use Monolog\Processor\ProcessorInterface;
+use Psr\Clock\ClockInterface;
 use Psr\Log\LoggerInterface;
 use Psr\Log\InvalidArgumentException;
 use Psr\Log\LogLevel;
@@ -146,6 +147,11 @@ class Logger implements LoggerInterface, ResettableInterface
 
     protected DateTimeZone $timezone;
 
+    /**
+     * Clock used to timestamp new records, or null to read the current time from the engine
+     */
+    protected ClockInterface|null $clock = null;
+
     protected Closure|null $exceptionHandler = null;
 
     /**
@@ -169,15 +175,17 @@ class Logger implements LoggerInterface, ResettableInterface
      * @param list<HandlerInterface> $handlers   Optional stack of handlers, the first one in the array is called first, etc.
      * @param callable[]         $processors Optional array of processors
      * @param DateTimeZone|null  $timezone   Optional timezone, if not provided date_default_timezone_get() will be used
+     * @param ClockInterface|null $clock     Optional clock to read the current time from, if not provided the engine's current time is used
      *
      * @phpstan-param array<(callable(LogRecord): LogRecord)|ProcessorInterface> $processors
      */
-    public function __construct(string $name, array $handlers = [], array $processors = [], DateTimeZone|null $timezone = null)
+    public function __construct(string $name, array $handlers = [], array $processors = [], DateTimeZone|null $timezone = null, ClockInterface|null $clock = null)
     {
         $this->name = $name;
         $this->setHandlers($handlers);
         $this->processors = $processors;
         $this->timezone = $timezone ?? new DateTimeZone(date_default_timezone_get());
+        $this->clock = $clock;
         $this->fiberLogDepth = new \WeakMap();
     }
 
@@ -357,7 +365,7 @@ class Logger implements LoggerInterface, ResettableInterface
             $recordInitialized = \count($this->processors) === 0;
 
             $record = new LogRecord(
-                datetime: $datetime ?? new JsonSerializableDateTimeImmutable($this->microsecondTimestamps, $this->timezone),
+                datetime: $datetime ?? $this->createDateTime(),
                 channel: $this->name,
                 level: self::toMonologLevel($level),
                 message: $message,
@@ -519,7 +527,7 @@ class Logger implements LoggerInterface, ResettableInterface
     public function isHandling(int|string|Level $level): bool
     {
         $record = new LogRecord(
-            datetime: new JsonSerializableDateTimeImmutable($this->microsecondTimestamps, $this->timezone),
+            datetime: $this->createDateTime(),
             channel: $this->name,
             message: '',
             level: self::toMonologLevel($level),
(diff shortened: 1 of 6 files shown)


src/Monolog/Logger.php
```diff
@@ -705,6 +713,60 @@ public function getTimezone(): DateTimeZone
...
+     */
+    private function createDateTime(): JsonSerializableDateTimeImmutable
+    {
+        $datetime = new JsonSerializableDateTimeImmutable($this->microsecondTimestamps, $this->timezone);
+
+        if (null === $this->clock) {
+            return $datetime;
+        }
+
+        $now = $this->clock->now();
```
Seldaek: I think we should give the clock a chance to return the correct data type at least..

```suggestion
        if (null === $this->clock) {
            return new JsonSerializableDateTimeImmutable($this->microsecondTimestamps, $this->timezone);
        }
        
        $now = $this->clock->now();
        if ($now instanceof JsonSerializableDateTimeImmutable) {
            return $now;
        }
        
        $datetime = new JsonSerializableDateTimeImmutable($this->microsecondTimestamps, $this->timezone);
```

src/Monolog/Logger.php
```diff
@@ -705,6 +713,60 @@ public function getTimezone(): DateTimeZone
...
+        $datetime = $datetime->setTimestamp($now->getTimestamp());
+        $microseconds = (int) $now->format('u');
+
+        if (0 !== $microseconds) {
+            // DateInterval has no notation for microseconds, they can only be set on the property
+            $interval = new \DateInterval('PT0S');
+            $interval->f = $microseconds / 1000000;
+
+            $datetime = $datetime->add($interval);
+        }
```
Seldaek: This should be as accurate and faster on 8.4 and looks less hacky on 8.3 and below, not sure about performance tho but anyone who cares about this level of performance should be running 8.4 or higher anyway at this point IMO.

```suggestion
        if (\PHP_VERSION_ID >= 80400) {
            $datetime = $datetime->setTimestamp($now->getTimestamp());
            $datetime = $datetime->setMicrosecond($now->getMicrosecond());
        } else {
            $datetime = $datetime->modify('@'.$now->format('U.u');
        }
```

src/Monolog/Logger.php
Amoifr: Applied in 8c321db, with a test covering it: a clock returning a `JsonSerializableDateTimeImmutable` in its own timezone is handed straight through.

src/Monolog/Logger.php
Amoifr: Took the 8.4 branch in 8c321db, it is cleaner, thanks. I kept the interval for older versions though, because `modify('@'.$now->format('U.u'))` is what the current code deliberately avoids, and the data provider covers both reasons:

```
clock       : 1950-01-02 03:04:05.123456 UTC
modify(@U.u): 1950-01-02 03:04:04.876544 +00:00
```

The second is wrong before 1970, the fraction being added to a negative timestamp, and `@` forces UTC, so the logger timezone becomes an offset even for a modern instant, which `testClockInstantIsRenderedInTheLoggerTimezone` asserts against. I checked the two paths agree on every instant of the provider.

One side effect worth mentioning: `setMicrosecond()` fails the PHPStan (8.1) job with `method.notFound`, reproduced here on 8.2. I used the `phpstan-ignore-by-php-version.neon.php` mechanism already in the repo rather than an inline ignore, which would then be unmatched on the latest job.

Seldaek:
Thanks!

PR #2066, opened 2026-09-05
Fix TelegramBotHandler breaking HTML tags on split (#1990)

### LINKED ISSUE #1990: TelegramBotHandler: API returns error when truncating/splitting in HTML parse mode
Monolog 3.9

I set [`$parseMode`](https://github.com/Seldaek/monolog/blob/3.9.0/src/Monolog/Handler/TelegramBotHandler.php#L114) to `'HTML'` and keep the default [`$splitLongMessages`](https://github.com/Seldaek/monolog/blob/3.9.0/src/Monolog/Handler/TelegramBotHandler.php#L117) as `false`.

When the message is looooong, it [gets truncated](https://github.com/Seldaek/monolog/blob/3.9.0/src/Monolog/Handler/TelegramBotHandler.php#L292). If the truncation happens inside HTML tag,
the [Telegram API returns error 400](https://github.com/Seldaek/monolog/blob/3.9.0/src/Monolog/Handler/TelegramBotHandler.php#L280):

> Bad Request: can't parse entities: Can't find end tag corresponding to start tag "code"

In my example the loooong string has `<code>blah blah</code>`. Lets say it gets truncated at first `blah`, then the `</code>` is gone and API complains.

* Possible solution 1: would need to fix (actually parse?) HTML if truncating?
* Possible solution 2: strip HTML tags when truncating? (kinda ugly)
* Low hanging fruit: at least want about that in the (php) docs?

---

EDIT: actually, with `$splitLongMessages=true` same kind of problem – the message splits inside HTML causing Telegram API error(s).

wakqasahmed:
Setting parseMode to HTML and hitting the length limit corrupts the message: truncation just cuts the string at a byte offset, so a `<code>` or `<b>` tag opened earlier in the text loses its closing tag, and Telegram's API rejects the whole message with a 400 ("can't parse entities"). Splitting has the same problem since str_split() cuts at fixed byte boundaries regardless of any tag straddling the boundary.

This only changes behavior when parseMode is HTML. Truncation now backs off before any tag that got cut in half and appends closing tags for whatever is still open at the cutoff, before adding the truncated marker. Splitting tracks open tags across chunks, closes them at the end of a chunk and reopens the same tags (with their attributes, so `<a href="...">` survives too) at the start of the next one, shrinking the chunk further if needed to stay under the length limit. Plain text and the other parse modes still go through the original str_split()/substr() path, untouched.

Added two tests in TelegramBotHandlerTest with a long `<code>` block: one confirms truncation ends with a closed `</code>` tag and stays under the length limit, the other confirms splitting reopens `<code>` in every chunk and reassembles back to the original content. Ran both against the old code first to confirm they fail (unclosed/duplicated tags), then against the fix. Also ran the full TelegramBotHandlerTest suite (12 tests, 34 assertions) and phpstan on the changed file, both clean.

Closes #1990

### DIFF
--- a/src/Monolog/Handler/TelegramBotHandler.php
+++ b/src/Monolog/Handler/TelegramBotHandler.php
@@ -329,9 +329,139 @@ class TelegramBotHandler extends AbstractProcessingHandler
     {
         $truncatedMarker = ' (…truncated)';
         if (!$this->splitLongMessages && \strlen($message) > self::MAX_MESSAGE_LENGTH) {
-            return [Utils::substr($message, 0, self::MAX_MESSAGE_LENGTH - \strlen($truncatedMarker)) . $truncatedMarker];
+            $maxLength = self::MAX_MESSAGE_LENGTH - \strlen($truncatedMarker);
+            $truncated = Utils::substr($message, 0, $maxLength);
+
+            if ($this->parseMode === 'HTML') {
+                $truncated = $this->trimPartialHtmlTag($truncated);
+                $closing = $this->closingHtmlTagsFor($this->updateOpenHtmlTags($truncated, []));
+
+                while ($truncated !== '' && \strlen($truncated . $closing) > $maxLength) {
+                    $truncated = $this->trimPartialHtmlTag(Utils::substr($truncated, 0, -1));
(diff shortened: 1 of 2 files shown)


Seldaek:
Thanks

